PR #17, opened 2018-04-18
Connect the front-end with the back-end

sbegaudeau:
This commit will add new stateful components which will be used to hold
the state for the user interface and communicate with the server. The
static data have been removed and the user interface is now computed
from the responses of the server.

The stateful containers, which can not be easily reused have been placed
in a dedicated folder named containers. This way, reusable and non
reusable components can easily be identified.

Bug: https://github.com/eclipse/sirius-components/issues/7

## Types of changes

* [ ] Bug fix (non*breaking change which fixes an issue)
* [x] New feature (non*breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [x] I have read the **CONTRIBUTING** document.
* [x] My code follows the code style of this project.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.
* [x] All new and existing tests passed.
* [ ] Continuous integration improvement

### DIFF
--- a/src/containers/projects/ProjectViewStateContainer.js
+++ b/src/containers/projects/ProjectViewStateContainer.js
@@ -1,0 +1,51 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Obeo and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2
+ * which accompanies this distribution and is available at
+ * https://www.eclipse.org/legal/epl-2.0.
+ *******************************************************************************/
+
+import React, { Component } from 'react';
+import { withRouter } from 'react-router-dom';
+
+import { ProjectView } from '../../components/projects/project/ProjectView';
+
+/**
+ * The ProjectViewStateContainerWithoutRouter is the stateful component used to
+ * manipulate the state of the ProjectView.
+ *
+ * It leverages the router API in order to retrieve the name of the project to
+ * display from the URL.
+ */
+class ProjectViewStateContainerWithoutRouter extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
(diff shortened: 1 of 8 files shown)


src/containers/projects/ProjectViewStateContainer.js
```diff
@@ -0,0 +1,51 @@
...
+  render() {
+    const { project } = this.state;
+
+    if (project === undefined) {
+      return <p>Loading</p>;
+    }
+    return <ProjectView project={project} {...this.props} />;
+  }
+}
+export const ProjectViewStateContainer = withRouter(ProjectViewStateContainerWithoutRouter);
```
florianbarbin: That means ProjectViewStateContainerWithoutRouter will have access to the route ?

sbegaudeau: ProjectViewStateContainerWithoutRouter is not exported, we only export the version decorated by "withRouter()". This usage of the high order component pattern (a pattern that I have presented during the training) will add several things in the props of the component.

The component will thus have access to some properties computed by the router like "match" (line 30) in order to retrieve the name of the project from the current URL. We can retrieve the value "projectName" from the match because we have specified this parameter in our URL in Main.js (you can also see it below).

```javascript
<main className={mainClassNames} {...props}>
  <Switch>
    <Route exact path="/" component={DashboardViewStateContainer} />
    <Route exact path="/projects" component={ListProjectsViewStateContainer} />
    <Route exact path="/projects/:projectName" component={ProjectViewStateContainer} />
  </Switch>
</main>
```

sbegaudeau:
In order to properly evaluate this review, you can use [this other review](https://git.eclipse.org/r/119016) from the Eclipse Sirius project in order to have a server with the proper API.

PR #106, opened 2020-11-13
[105] Take Domain class into account for DomainBasedEdges conversion

AxelRICHARD:
Signed-off-by: Axel RICHARD <axel.richard@obeo.fr>

### DIFF
--- a/backend/sirius-web-compatibility/src/main/java/org/eclipse/sirius/web/compat/diagrams/EdgeMappingConverter.java
+++ b/backend/sirius-web-compatibility/src/main/java/org/eclipse/sirius/web/compat/diagrams/EdgeMappingConverter.java
@@ -210,7 +210,8 @@ public class EdgeMappingConverter {
         if (edgeMapping.isUseDomainElement()) {
             String semanticCandidatesExpression = edgeMapping.getSemanticCandidatesExpression();
             String preconditionExpression = edgeMapping.getPreconditionExpression();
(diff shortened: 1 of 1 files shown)


backend/sirius-web-compatibility/src/main/java/org/eclipse/sirius/web/compat/diagrams/EdgeMappingConverter.java
```diff
@@ -210,7 +210,11 @@ public EdgeDescription convert(EdgeMapping edgeMapping) {
         if (edgeMapping.isUseDomainElement()) {
             String semanticCandidatesExpression = edgeMapping.getSemanticCandidatesExpression();
             String preconditionExpression = edgeMapping.getPreconditionExpression();
-            semanticElementsProvider = new SemanticCandidatesProvider(this.interpreter, EOBJECT, semanticCandidatesExpression, preconditionExpression);
+            String domainClass = edgeMapping.getDomainClass();
```
pcdavid, line 213: `DomainClassPredicate` (which is used underneath) already handles the case where the domain class is blank as equivalent to `EObject`. It just does not accept `null`. Proposal
```java
String domainClass = Optional.ofNullable(edgeMapping.getDomainClass()).orElse("");
```
and let `DomainClassPredicate` treat "" as "anything".

pcdavid (changes requested):

sbegaudeau:
It looks good on my side, I'll ping @pcdavid to see if it's ok for him too

sbegaudeau (approved):

pcdavid (approved):

PR #111, opened 2020-11-24
Update versions of GraphQL Java, Guava and JS/ES compilation target

pcdavid:
Move to the most recent versions of GraphQL Java and Guava, and target an earlier version of JS/ES (ES2019 instead of ES2020) to improve compatibility.


sbegaudeau (changes requested):
Everything seems fine but an important issue has been discovered with GraphQL Java 16, could you update to 16.1 please?

sbegaudeau (changes requested):
sirius-web-graphql-utils is still using GraphQL Java 14

backend/sirius-web-graphql-utils/pom.xml
```diff
@@ -47,7 +48,7 @@
 		<dependency>
         	<groupId>com.graphql-java</groupId>
         	<artifactId>graphql-java</artifactId>
-        	<version>14.0</version>
+        	<version>${graphql-java.version}</version>
```
sbegaudeau, line 51: Could you fix the problem in the appropriate commit please? Everything is fine apart from that.

sbegaudeau (changes requested):

backend/sirius-web-graphql-utils/pom.xml
pcdavid: I've squashed all the GraphQL updated into a single commit (14.0 => 16.1).

sbegaudeau (approved):

PR #123, opened 2020-11-26
Create feature_template.md

hmarchadour:


sbegaudeau (changes requested):
Could you use the same Git commit message pattern as used since the beginning of the project?

sbegaudeau (approved):

PR #128, opened 2020-11-27
Use Apollo instead of our custom GraphQL hooks for the diagram part

pcdavid:
This is not completely finished, in particular I have not tested the error handling. In particular the `connection_error` message type which was handled before is ignored at the moment. I'm publishing it anyway to get feedback on the rest while I continue on the error handling.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -25,6 +25,7 @@
     "test": "jest"
   },
   "devDependencies": {
+    "@apollo/client": "^3.2.7",
     "@rollup/plugin-commonjs": "^14.0.0",
     "@rollup/plugin-image": "^2.0.5",
     "@rollup/plugin-node-resolve": "^8.4.0",
@@ -58,6 +59,7 @@
     "rollup-plugin-typescript2": "^0.27.1",
     "rollup": "^2.22.1",
     "sprotty": "0.7.0",
+    "subscriptions-transport-ws": "^0.9.18",
     "svg-url-loader": "^6.0.0",
     "terser": "^5.3.1",
     "ts-jest": "26.4.1",
@@ -66,6 +68,8 @@
     "universal-cookie": "^4.0.2"
   },
   "peerDependencies": {
+    "@apollo/client": "^3.2.7",
+    "subscriptions-transport-ws": "^0.9.18",
     "airbnb-prop-types": "^2.15.0",
     "graphql-tag": "^2.11.0",
     "graphql": "14.6.0",
--- a/frontend/src/diagram/DiagramWebSocketContainer.tsx
+++ b/frontend/src/diagram/DiagramWebSocketContainer.tsx
@@ -10,47 +10,45 @@
  * Contributors:
  *     Obeo - initial API and implementation
  *******************************************************************************/
-import { GraphQLClient } from 'common/GraphQLClient';
-import { useMutation, useLazyQuery } from 'common/GraphQLHooks';
+import { useLazyQuery, useMutation, useSubscription } from '@apollo/client';
 import { Text } from 'core/text/Text';
 import {
   COMPLETE__STATE,
   HANDLE_COMPLETE__ACTION,
-  HANDLE_CONNECTION_ERROR__ACTION,
   HANDLE_DATA__ACTION,
-  HANDLE_ERROR__ACTION,
   INITIALIZE__ACTION,
   LOADING__STATE,
   READY__STATE,
   SELECTED_ELEMENT__ACTION,
   SELECTION__ACTION,
+  SELECT_ZOOM_LEVEL__ACTION,
   SET_ACTIVE_TOOL__ACTION,
   SET_CONTEXTUAL_PALETTE__ACTION,
   SET_DEFAULT_TOOL__ACTION,
   SET_TOOL_SECTIONS__ACTION,
-  SELECT_ZOOM_LEVEL__ACTION,
   SWITCH_REPRESENTATION__ACTION,
 } from 'diagram/machine';
+import { ContextualPalette } from 'diagram/palette/ContextualPalette';
 import { initialState, reducer } from 'diagram/reducer';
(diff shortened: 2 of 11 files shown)


sbegaudeau (commented):
I think that pretty much everything in this file is wrong. You are adding more of the same "wrongdoing" to this file which may work but we will have to fix that.

Neither or dev dependencies nor our peer dependencies make sense.

sbegaudeau (commented):
I agree with this code but I wonder if it should not be in sirius-web instead.

For example, I don't see why we should have anything to do with credentials support. Especially since different users of `sirius-components` already have different authentication strategies.

frontend/package.json
```diff
@@ -25,6 +25,7 @@
     "test": "jest"
   },
   "devDependencies": {
+    "@apollo/client": "^3.2.7",
```
sbegaudeau, line 28: I think that pretty much everything in this file is wrong. You are adding more of the same "wrongdoing" to this file which may work but we will have to fix that.

Neither or dev dependencies nor our peer dependencies make sense.

frontend/src/common/ApolloGraphQLClient.ts
```diff
@@ -0,0 +1,42 @@
+/*******************************************************************************
```
sbegaudeau, line 1: I agree with this code but I wonder if it should not be in sirius-web instead.

For example, I don't see why we should have anything to do with credentials support. Especially since different users of `sirius-components` already have different authentication strategies.

frontend/src/diagram/DiagramWebSocketContainer.tsx
```diff
@@ -415,12 +417,15 @@ export const DiagramWebSocketContainer = ({ representationId, selection, setSele
...
 
   useEffect(() => {
-    if (!toolSectionData.loading && viewState === READY__STATE) {
-      if (toolSectionData?.data?.data?.viewer?.toolSections) {
-        dispatch({ type: SET_TOOL_SECTIONS__ACTION, toolSections: toolSectionData.data.data.viewer.toolSections });
+    if (!toolSectionLoading && viewState === READY__STATE) {
+      if (toolSectionData?.viewer?.toolSections) {
+        dispatch({
+          type: SET_TOOL_SECTIONS__ACTION,
+          toolSections: JSON.parse(JSON.stringify(toolSectionData.viewer.toolSections)),
```
sbegaudeau, line 421: The reducer should be modified to use what it needs from the server response instead of just modifying it inline.

JSON.parse + JSON.stringify is overkill to not use some basic rest and spread.

frontend/src/diagram/reducer.ts
```diff
@@ -207,8 +207,8 @@ const handleDataAction = (prevState, action) => {
   const { message } = action;
 
   let state = prevState;
-  if (message.payload && message.payload.data && message.payload.data.diagramEvent) {
-    const { diagramEvent } = message.payload.data;
+  if (message.data && message.data.diagramEvent) {
+    const { diagramEvent } = message.data;
```
sbegaudeau, line 211: No unit tests to update?

sbegaudeau (changes requested):

frontend/src/diagram/reducer.ts
pcdavid: Fixed by https://github.com/eclipse-sirius/sirius-components/pull/128/commits/0f6add0edf420bcbba4c3a325b7927a3c182746f

frontend/src/diagram/DiagramWebSocketContainer.tsx
pcdavid: Fixed by https://github.com/eclipse-sirius/sirius-components/pull/128/commits/44ce7aede446bb4bced5cd8e72c5bb2d66911ef3

frontend/package.json
pcdavid: Agreed, but if you're OK we'll handle this issue in a separate PR. It's not worse than what was done before, so I don't think this should block merging this as long as we plan to fix the global issue soon.

frontend/src/common/ApolloGraphQLClient.ts
pcdavid: I understand the concern, but this would lead to duplicate code in all apps based on `sirius-components`. I'll see if we can tweak the `ApolloClient` after it has been constructed, so that we can provide here one with some default config, and apps that need it could add the `credentials` part if they need it.

Note that the current custom hooks already mix in some authentication-related code, e.g.:
* https://github.com/eclipse-sirius/sirius-components/blob/master/frontend/src/common/GraphQLHooks.ts#L185
* https://github.com/eclipse-sirius/sirius-components/blob/master/frontend/src/common/GraphQLHttpClient.ts#L26 (hard-coded `credentials: 'include',  mode: 'cors'`).

pcdavid: Fixed by https://github.com/eclipse-sirius/sirius-components/pull/128/commits/4a118e75b505730d3251d0e596a894ce719a727e, and https://github.com/eclipse-sirius/sirius-web/pull/12/commits/8d298629cbd08798d4678fe0bedf158e4d732f6f on the sirius-web side.

sbegaudeau:
This seems fine, I'll just wait for the fix for the latest details on https://github.com/eclipse-sirius/sirius-web/pull/12 to merge them both.

sbegaudeau (approved):

sbegaudeau (changes requested):
I'll open an issue in order to activate the squash and merge option but in the meantime since everything is now ok, could you squash the whole PR and then I'll merge it. Thanks

sbegaudeau:
Forget about the last comment, I can do it myself. Since everything is now ok, I'll make sure that everything is merged regarding Apollo.

sbegaudeau (approved):

PR #155, opened 2020-12-02
[151] Add end to end tests

sbegaudeau:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/151


sample/test-sample/frontend/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@eclipse-sirius/sirius-web",
```
pcdavid, line 2: Should this use a different name than the "real" sirius-web app?

sample/test-sample/frontend/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "author": "Eclipse Sirius",
+  "version": "0.1.0",
+  "license": "EPL-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/eclipse-sirius/sirius-web"
```
pcdavid, line 8: Wrong URL, but I'm not sur this section makes sense for a sample app used for testing only.

sample/test-sample/frontend/package.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "publishConfig": {
+    "registry": "https://npm.pkg.github.com/"
+  },
+  "private": true,
```
pcdavid, line 13: Same remark: could probably be removed from this one. I don't think it will ever make sense to publish it, will it?

pcdavid (changes requested):

sample/test-sample/frontend/package.json
sbegaudeau: Done

sample/test-sample/frontend/package.json
sbegaudeau: Done

sample/test-sample/frontend/package.json
sbegaudeau: Done

pcdavid (approved):

PR #198, opened 2020-12-16
[173] Migrate NewProjectView to XState and MaterialUI

florianbarbin:
### Type of this PR 

- [ ] Bug fix
- [X] New feature or improvement
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)
WIP for issue 173
...

### What does this PR do?

...

...
 

### Potential side effects

...

### How to test this PR?

- [ ] Frontend Unit tests
- [ ] Backend Unit tests

### Checklist

- [ ] All tests pass.


frontend/src/views/new-project/NewProjectView.tsx
```diff
@@ -60,66 +60,101 @@ const createProjectMutation = gql`
...
       },
     };
+    const submitEvent: SubmitEvent = { type: 'SUBMIT' };
+    dispatch(submitEvent);
     createProject({ variables });
   };
 
   useEffect(() => {
     if (!loading && !error && data) {
       if (error) {
```
sbegaudeau, line 88: Something seems odd here, first of all this code is not executed if there is an error while we would like to display errors. Then inside there is an `if (error)`.

frontend/src/views/new-project/NewProjectView.tsx
```diff
@@ -60,66 +60,101 @@ const createProjectMutation = gql`
...
-        dispatch({ type: HANDLE_SUBMIT_FAILURE__ACTION, reponse: error });
+        let errorMessage = error.message;
+        const showToastEvent: ShowToastEvent = { type: 'SHOW_TOAST', message: errorMessage };
```
sbegaudeau, line 90: You could use something like that:

const { message } = error;
const showToastEvent: ShowToastEvent = { type: 'SHOW_TOAST', message };

frontend/src/views/new-project/NewProjectView.tsx
```diff
@@ -60,66 +60,101 @@ const createProjectMutation = gql`
...
+        dispatch(showToastEvent);
       } else if (data) {
```
sbegaudeau, line 92: Don't use an else if, you can have both errors and data! Ok in such a simple example there 99.999% chance that it can't occur but don't get used to this pattern. You can have both at the same time.

frontend/src/views/new-project/NewProjectView.tsx
```diff
@@ -60,66 +60,101 @@ const createProjectMutation = gql`
...
-        dispatch({ type: HANDLE_SUBMIT__ACTION, response: data });
+        let id = data.createProject?.project?.id;
+        if (id) {
```
sbegaudeau, line 94: You may want to discriminate on the type of the response instead of the presence of one field. Something like that:

const typename = data.createProject.__typename;
if (typename === 'CreateProjectSuccessPayload') {
  const { id } = data.createProject.project;
} else if (typename === 'ErrorPayload') {
  const { message } = data.createProject;
}
```

Inside of the first if, you know for a fact that project and id cannot be null and inside of the second one message cannot be null either. That's part of the GraphQL API, the server will make sure of it, otherwise you will receive an error.

File: frontend/src/views/new-project/NewProjectView.tsx [hunk 5]
```diff
@@ -60,66 +60,101 @@ const createProjectMutation = gql`
...
 
-  const canBeSubmitted = viewState === VALID__STATE;
-  const nameMessageContent = nameMessage ? (
-    <Message severity={nameMessageSeverity} data-testid="nameMessage" content={nameMessage} />
-  ) : null;
+  const canBeSubmitted = newProjectView === 'valid';
+  let validationMessage = '';
+  if (newProjectView === 'invalid') {
+    validationMessage = 'The name must contain between 3 and 20 characters.';
+  }
```
sbegaudeau, line 115: This logic should be in the state machine. After some discussion with Laurent, it would be better to have the following fields in your context : name, nameMessage, isNameError or something like that

frontend/src/views/new-project/NewProjectView.tsx
```diff
@@ -60,66 +60,101 @@ const createProjectMutation = gql`
...
-              data-testid="name"
-            />
-            <div className={styles.messageArea}>{nameMessageContent}</div>
-          </Label>
-          <Buttons>
-            <ActionButton type="submit" disabled={!canBeSubmitted} label="Create" data-testid="create-project" />
-            <SecondaryButton type="submit" to={`/`} label="Cancel" data-testid="back" />
-          </Buttons>
+          <TextField
+            error={newProjectView === 'invalid'}
```
sbegaudeau, line 122: It works because the example is really really small but this logic should be part of the state machine

frontend/src/views/new-project/NewProjectViewMachine.ts
```diff
@@ -0,0 +1,207 @@
...
+  | HideToastEvent;
+
+const isChangeNameEvent = (event: NewProjectEvent): event is ChangeNameEvent => {
+  return event.type === 'CHANGE_NAME';
+};
+const isHandleSuccessEvent = (event: NewProjectEvent): event is HandleSuccessEvent => {
+  return event.type === 'HANDLE_SUCCESS';
+};
+const isShowToastEvent = (event: NewProjectEvent): event is ShowToastEvent => {
+  return event.type === 'SHOW_TOAST';
```
sbegaudeau, line 67: This is not needed since, in a way, it double checks something that XState will check for us. We talked about it on zoom but I'm still adding this comment for others to see.

frontend/src/views/new-project/NewProjectViewMachine.ts
```diff
@@ -0,0 +1,207 @@
...
+};
+export const newProjectViewMachine = Machine<NewProjectViewContext, NewProjectViewStateSchema, NewProjectEvent>(
+  {
+    id: 'NewProjectView',
+    type: 'parallel',
+    context: {
+      name: '',
```
sbegaudeau, line 74: Just putting the same comment here too:

You would be better prepared for evolutions of your state machine with the following fields:
- name: ''
- nameMessage: 'The name must contain between 3 and 20 characters'
- isNameError: fase
- message: null
- newProjectId: null

sbegaudeau (changes requested):

frontend/src/views/new-project/NewProjectView.tsx
florianbarbin: done

frontend/src/views/new-project/NewProjectView.tsx
florianbarbin: done

sbegaudeau (changes requested):
- The file NewProjectView.module.css is not used anymore
- The file machine.ts in the same folder is not used anymore
- The file reducer.ts in the same folder is not used anymore

sbegaudeau (approved):

PR #212, opened 2020-12-22
Use Apollo in more places

pcdavid:


pcdavid:
The first commit in this PR is the same as the one in https://github.com/eclipse-sirius/sirius-components/pull/207

sbegaudeau (approved):

Issue #256, opened 2021-01-19
Fix support for EMF composed images


pcdavid:
`org.eclipse.sirius.web.emf.services.ObjectService` assumes that `org.eclipse.sirius.web.emf.services.ObjectService` can only return an `URL` or an `URI`. This is false for at least Ecore, which can return `ComposedImages` to overlay decorators. See `org.eclipse.emf.edit.provider.ItemProviderAdapter.overlayImage(Object, Object)` for example.

In that case ObjectService returns `null`, which causes NPEs when trying to use it.

PR #257, opened 2021-01-19
[256] Fix support for EMF composed images

pcdavid:
See #256.


sbegaudeau (approved):
+1, I'll rebase it and merge it in a second

Issue #300, opened 2021-02-10
Semantic changes cannot modify the size of an element anymore


sbegaudeau:
In flow, if I change the capacity of an element, it does not change its size anymore. The image ratio is not taken into account anymore.

sbegaudeau:
A fix has been contributed but with the current structure of the API it does not cover all use cases.

PR #305, opened 2021-02-10
[300] Fix size computation

wpiers:
Change-Id: If25f64661b655d71b8776796f66249dab13e0e22
Signed-off-by: William Piers <william.piers@obeo.fr>

### Type of this PR 

- [X] Bug fix
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

See #300

### What does this PR do?

Recompute the size of each node without children on refresh. This way nodes with conditional sizes are reevaluated.

### Potential side effects

The solution cannot apply to containers as their size is computed by ELK accordingly to its content and the incremental layout, in its current state, cannot do that.

### How to test this PR?

- [ ] Frontend Unit tests
- [ ] Backend Unit tests
- [X] Manual Test : activate manual layout (-Dsirius.layout.auto.deactivate=true), create a diagram with a fan, change its size => the node is resized

### Checklist

- [X] I have read CONTRIBUTING carefully.
- [X] I have signed the [Contributor License Agreement](http://www.eclipse.org/legal/CLA.php).
- [X] All my commits are signed-off (`-s`) with my mail address of my Eclipse Account.
- [X] I have covered my changes by unit tests or integration tests or manual tests.
- [X] All tests pass.
- [X] I have updated the documentation accordingly
- [X] New React components are availables in storybook

### DIFF
--- a/backend/sirius-web-diagrams/src/main/java/org/eclipse/sirius/web/diagrams/components/NodeComponent.java
+++ b/backend/sirius-web-diagrams/src/main/java/org/eclipse/sirius/web/diagrams/components/NodeComponent.java
@@ -105,10 +105,9 @@ public class NodeComponent implements IComponent {
     private Element doRender(VariableManager nodeVariableManager, String targetObjectId, Optional<Node> optionalPreviousNode) {
         NodeDescription nodeDescription = this.props.getNodeDescription();
         NodeContainmentKind containmentKind = this.props.getContainmentKind();
-        DiagramRenderingCache cache = this.props.getCache();
         NodePositionProvider nodePositionProvider = this.props.getNodePositionProvider();
-        NodeSizeProvider nodeSizeProvider = new NodeSizeProvider();
 
+        NodeSizeProvider nodeSizeProvider = new NodeSizeProvider();
         UUID nodeId = optionalPreviousNode.map(Node::getId).orElseGet(() -> this.computeNodeId(targetObjectId));
         Optional<Label> optionalPreviousLabel = optionalPreviousNode.map(Node::getLabel);
         String type = nodeDescription.getTypeProvider().apply(nodeVariableManager);
@@ -117,8 +116,6 @@ public class NodeComponent implements IComponent {
 
         INodeStyle style = nodeDescription.getStyleProvider().apply(nodeVariableManager);
 
-        IDiagramElementRequestor diagramElementRequestor = new DiagramElementRequestor();
-
         //@formatter:off
         Position position = optionalPreviousNode
                 .map(Node::getPosition)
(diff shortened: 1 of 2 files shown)


backend/sirius-web-diagrams/src/main/java/org/eclipse/sirius/web/diagrams/components/NodeComponent.java
```diff
@@ -171,7 +167,13 @@ private Element doRender(VariableManager nodeVariableManager, String targetObjec
...
         //@formatter:on
 
-        Size size = optionalPreviousNode.map(Node::getSize).orElseGet(() -> nodeSizeProvider.getSize(style, childNodes));
+        Size size;
+        if (childNodes.isEmpty()) {
+            size = nodeSizeProvider.getSize(style, childNodes);
+        } else {
+            size = optionalPreviousNode.map(Node::getSize).orElseGet(() -> nodeSizeProvider.getSize(style, childNodes));
+        }
+
```
sbegaudeau, line 176: Why are you computing the size when there are no children but keeping the old size if there are children? The issue was that semantic changes cannot impact the size anymore, regardless if there are children or not.

backend/sirius-web-diagrams/src/main/java/org/eclipse/sirius/web/diagrams/components/NodeSizeProvider.java
```diff
@@ -25,12 +25,16 @@
  * Provides the Size to apply to a new node.
  *
  * @author fbarbin
+ * @author wpiers
  */
 public class NodeSizeProvider {
 
+    private static final int ELK_SIZE_DIFF = 24;
```
sbegaudeau, line 32: Could you add a comment indicating what this magic number is, I personally do not remember.

sbegaudeau (changes requested):

backend/sirius-web-diagrams/src/main/java/org/eclipse/sirius/web/diagrams/components/NodeComponent.java
wpiers: In its current state, the rendering cannot compute the size of an element according to its children.
This is to be solved by https://github.com/eclipse-sirius/sirius-components/pull/299
For now the rendering can only recompute sizes for childless nodes (and solve the issue with fans with conditional sizes) but rely on ELK (so, keep the precomputed size) for nodes with children.

backend/sirius-web-diagrams/src/main/java/org/eclipse/sirius/web/diagrams/components/NodeSizeProvider.java
wpiers: comment added in https://github.com/wpiers/sirius-components/commit/7b33800e71e2bcc8b2cb048f31b5ebb70ec8e681

backend/sirius-web-diagrams/src/main/java/org/eclipse/sirius/web/diagrams/components/NodeComponent.java
```diff
@@ -108,15 +108,12 @@ private Element doRender(VariableManager nodeVariableManager, String targetObjec
...
         Optional<Label> optionalPreviousLabel = optionalPreviousNode.map(Node::getLabel);
         String type = nodeDescription.getTypeProvider().apply(nodeVariableManager);
         String targetObjectKind = nodeDescription.getTargetObjectKindProvider().apply(nodeVariableManager);
         String targetObjectLabel = nodeDescription.getTargetObjectLabelProvider().apply(nodeVariableManager);
-
         INodeStyle style = nodeDescription.getStyleProvider().apply(nodeVariableManager);
-
         IDiagramElementRequestor diagramElementRequestor = new DiagramElementRequestor();
 
         //@formatter:off
```
sbegaudeau, line 119: I was curious why those lines disappeared but now that I tried to restore them, I understand. If we keep them, we have a checkstyle error since the method is too long. The proper fix is not to remove empty lines and make the code less readable but instead to split the method in smaller ones.

sbegaudeau:
No need for changeIds on your commit but add a reference to the bug as in other commits.

sbegaudeau (approved):

PR #320, opened 2021-02-16
[318] Register all ELK layout meta data providers

### LINKED ISSUE #318: Add support for more ELK layout providers
We currently only support the "Layered" ELK layout provider, but it supports many others. Even if we do not use them in the default "Arrange All" algorithm, it would be useful to make them available programmatically.

pcdavid:
Only the layered algorithm was supported, this fix initializes the
layout metadata service with the following ELK algorithms:

- Force: Simulating forces that move the nodes into a balanced distribution
- MrTree: A spanning tree of the input graph and arranges all nodes according to the resulting parent-children hierarchy
- Radial: Radial layouter takes a tree and places the nodes in radial order around the root
- Stress: Simulating forces that move the nodes into a balanced distribution
- Polyomino
- DisCo: Layouter for arranging unconnected subgraphs
- Spore

Fixes #318.

Bug: eclipse-sirius/sirius-components#318
Signed-off-by: Mélanie Bats <melanie.bats@obeo.fr>

### DIFF
--- a/backend/sirius-web-diagrams-layout/pom.xml
+++ b/backend/sirius-web-diagrams-layout/pom.xml
@@ -73,6 +73,46 @@
 			<artifactId>org.eclipse.elk.alg.layered</artifactId>
 			<version>${elk.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>org.eclipse.elk</groupId>
+			<artifactId>org.eclipse.elk.alg.common</artifactId>
+			<version>${elk.version}</version>
+		</dependency>
+		<dependency>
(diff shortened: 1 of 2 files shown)


sbegaudeau:
I am not sure why this is necessary? If someone want to use a different layout algorithm, they can provide one and depend on the necessary ELK dependencies. Unless I'm missing something, we are adding a dependency to 8 new jars that we know that we are not using. This commit also brings in our classpath Xtext, Xtend, another version of EMF, Antlr, Log4J, the Eclipse platform and more. That a lot for something that we won't use.

pcdavid:
OK

PR #354, opened 2021-02-26
[346] Fix handling of Unset VSM operation

pcdavid:


backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/compatibility/modeloperations/UnsetOperationHandler.java
```diff
@@ -49,40 +50,65 @@ public UnsetOperationHandler(AQLInterpreter interpreter, ChildModelOperationHand
...
-                Boolean eIsMany = ecoreIntrinsicExtender.eIsMany(ownerObject, featureName);
-                if (eIsMany) {
-                    ecoreIntrinsicExtender.eAdd(ownerObject, featureName, elementsToUnset);
-                } else {
-                    ecoreIntrinsicExtender.eSet(ownerObject, featureName, elementsToUnset);
-                }
+    private void unset(EObject context, EStructuralFeature featureToEdit, List<EObject> elementsToUnset) {
+        if (featureToEdit.isMany()) {
+            EList<?> values = (EList<?>) context.eGet(featureToEdit);
+            if (elementsToUnset == null) {
```
sbegaudeau, line 101: Is that really the Sirius RCP behavior? I'm not an expert so I'm asking to be sure. If the element expression of the unset operation is blank then it removes everything from the multivalued attribute.

I would have imagined that it would instead nothing. From what I understand in your code, there is a real difference between nothing in the expression and an expression returning nothing (which would thus remove nothing).

sbegaudeau: I don't see any mention of this in the Sirius RCP documentation but it may not be documented here: https://www.eclipse.org/sirius/doc/specifier/general/Model_Operations.html#unset

pcdavid: You're right. The code in Sirius Desktop is spread accross several layers (`UnsetTask`, `ModelAccessor`, `CompositeMetamodelExtender`, and finally `EcoreIntrinsicExtender`), each with its own quirks. I *think* the new version matches the Sirius Desktop behavior:

* Not setting the `elements` expression at all means "clear the feature" (set to null is single valued).
* If the `elements` expression is set but happens to return an empty collection, we do nothing.
* If the `elements` expression returns actual elements to remove, Sirius Dekstop ends up calling `EcoreUtil.remove()` on each. For multi-valued features this corresponds to `remove()`-ing each one; for single-valued features, `EcoreUtil.remove()` ignores the actual "value" to remove and unsets the feature.

sbegaudeau (approved):

PR #356, opened 2021-02-26
[339] Let RenameDiagramEventHandler handle diagram renaming

pcdavid:


pcdavid:
Broken. Don't merge as is.

PR #374, opened 2021-03-04
Improve the event processors

sbegaudeau:


backend/sirius-web-spring-collaborative-diagrams/src/main/java/org/eclipse/sirius/web/spring/collaborative/diagrams/DiagramEventFlux.java
```diff
@@ -42,7 +48,11 @@ public DiagramEventFlux(Diagram currentDiagram) {
 
     public void diagramRefreshed(IInput input, Diagram newDiagram) {
         this.currentDiagram = newDiagram;
-        this.sink.tryEmitNext(new DiagramRefreshedEventPayload(input.getId(), this.currentDiagram));
+        EmitResult emitResult = this.sink.tryEmitNext(new DiagramRefreshedEventPayload(input.getId(), this.currentDiagram));
+        if (emitResult.isFailure()) {
+            String pattern = "An error has occurred while emitting a DiagramRefreshedEventPayload: {0}"; //$NON-NLS-1$
+            this.logger.warn(MessageFormat.format(pattern, emitResult));
```
pcdavid, line 53: SLF4J supports message formatting natively, and using that support prevents building the string if the logging level invoked is disabled. This kind of pattern should be written like this instead:

```java
this.logger.warn("An error has occurred while emitting a DiagramRefreshedEventPayload: {}", emitResult);
```

Granted, in this particular case (logging an error) it's not much of an improvement, but there are many other places where we should probably use that pattern instead.

backend/sirius-web-spring-collaborative-forms/src/main/java/org/eclipse/sirius/web/spring/collaborative/forms/PropertiesEventProcessor.java
```diff
@@ -1,187 +0,0 @@
-/*******************************************************************************
```
pcdavid, line 1: Should'nt `org.eclipse.sirius.web.collaborative.forms.api.IPropertiesEventProcessor` be removed too?

doc/adrs/016_remove_subscription_description.adoc
```diff
@@ -0,0 +1,42 @@
...
+It has given us the ability to track whether a representation could be disposed since nobody was using it anymore for example.
+While efficient, it is built on top of Reactor and GraphQL Java and it ties our GraphQL WebSocket integration with some of our collaborative classes such as `IEditingContextEventProcessorRegistry`.
+While it may not have been a critical issue in the past, it is getting in our way to provide some other feature.
+
+
+== Decision
+
+We will stop relying on `SubscriptionDescription` in order to retrieve the `Principal` of a subscription.
+Instead, we will make sure that the `Principal` is available at all time in our WebSocket integration.
+On top of that, we will leverage existing Reactor mecanisms to detect when a WebSocket session is subscribing or unsubscribing from our reactive publishers.
```
pcdavid, line 15: typo: "mec[h]anisms

backend/sirius-web-spring-collaborative/src/main/java/org/eclipse/sirius/web/spring/collaborative/projects/EditingContextEventProcessor.java
```diff
@@ -362,10 +354,16 @@ private void disposeRepresentation(UUID representationId) {
...
     @Override
     public void dispose() {
+        this.logger.trace(MessageFormat.format("Disposing the editing context \"{0}\"", this.editingContext.getId())); //$NON-NLS-1$
+
         this.executor.shutdown();
 
-        this.representationEventProcessors.values().stream().forEach(IRepresentationEventProcessor::dispose);
+        this.representationEventProcessors.values().stream().forEach(entry -> {
+            entry.getDisposable().dispose();
+            entry.getRepresentationEventProcessor().dispose();
```
pcdavid, line 380: `RepresentationDescriptionEntry` could have a `dispose()` itself which does this.

backend/sirius-web-spring-collaborative/src/main/java/org/eclipse/sirius/web/spring/collaborative/projects/RepresentationEventProcessorEntry.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.sirius.web.collaborative.api.services.IRepresentationEventProcessor;
+
+import reactor.core.Disposable;
+
+/**
+ * Entry used to store the representation event processor along with additional information.
+ *
+ * @author sbegaudeau
+ */
+public class RepresentationEventProcessorEntry {
```
pcdavid, line 26: I have a feeling that there is a better name/suffix than "Entry" for this concept, but can't find it at the moment.

backend/sirius-web-spring-collaborative/src/main/java/org/eclipse/sirius/web/spring/collaborative/projects/EditingContextEventProcessorRegistry.java
```diff
@@ -86,35 +94,54 @@ public EditingContextEventProcessorRegistry(IEditingContextSearchService editing
...
     @Override
-    public void dispose(UUID editingContextId) {
-        // @formatter:off
-        Optional.ofNullable(this.editingContextEventProcessors.remove(editingContextId))
-                .ifPresent(EditingContextEventProcessor::dispose);
-        // @formatter:on
+    public void disposeEditingContextEventProcessor(UUID editingContextId) {
+        Optional.ofNullable(this.editingContextEventProcessors.remove(editingContextId)).ifPresent(entry -> {
+            entry.getDisposable().dispose();
+            entry.getEditingContextEventProcessor().dispose();
```
pcdavid, line 131: Same remark as for the other, this should be the responsibility of the entry itself to know what/how to dispose and in which order.

backend/sirius-web-spring-collaborative/src/main/java/org/eclipse/sirius/web/spring/collaborative/projects/EditingContextEventProcessorRegistry.java
```diff
@@ -86,35 +94,54 @@ public EditingContextEventProcessorRegistry(IEditingContextSearchService editing
...
+            if (optionalEditingContextEventProcessor.isEmpty()) {
                 Optional<IEditingContext> optionalEditingContext = this.editingContextSearchService.findById(editingContextId);
                 if (optionalEditingContext.isPresent()) {
                     IEditingContext editingContext = optionalEditingContext.get();
-                    editingContextEventProcessor = new EditingContextEventProcessor(editingContext, this.editingContextPersistenceService, this.applicationEventPublisher, this.objectService,
+
+                    var editingContextEventProcessor = new EditingContextEventProcessor(editingContext, this.editingContextPersistenceService, this.applicationEventPublisher, this.objectService,
                             this.editingContextEventHandlers, this.representationEventProcessorComposedFactory);
-                    this.editingContextEventProcessors.put(editingContextId, editingContextEventProcessor);
+                    Disposable subscription = editingContextEventProcessor.canBeDisposed().delayElements(Duration.ofSeconds(30)).subscribe(canBeDisposed -> {
```
pcdavid, line 106: Magic number: should be at least a constant, maybe even configurable via a system property (I believe Spring supports properties which denote times with units).

pcdavid (changes requested):
See comments.

backend/sirius-web-spring-collaborative/src/main/java/org/eclipse/sirius/web/spring/collaborative/projects/RepresentationEventProcessorEntry.java
sbegaudeau: I had the same feeling :)

backend/sirius-web-spring-collaborative/src/main/java/org/eclipse/sirius/web/spring/collaborative/representations/SubscriptionManager.java
```diff
@@ -42,20 +43,22 @@
...
+    private final Map<String, AtomicInteger> username2subscriptionCount = new ConcurrentHashMap<>();
 
     @Override
     public void add(IInput input, String username) {
-        int subscriptionCount = this.username2subscriptionCount.getOrDefault(username, 0).intValue();
-        this.username2subscriptionCount.put(username, subscriptionCount + 1);
+        var subscriptionCount = this.username2subscriptionCount.getOrDefault(username, new AtomicInteger());
+        subscriptionCount.getAndIncrement();
+
+        this.username2subscriptionCount.put(username, subscriptionCount);
```
pcdavid: Actually re-reading this I think this is not thread-safe either. From the initial state, two threads can enter `add`, both calling `getOrDefault`will return 2 `AtomicInteger`, increment them both to 1, and both but `1` into the map. The last one to call `put` will "win" but we still end with `1` for this username.

I think we should use `ConcurrentMap.computeIfAbsent()` instead of `getOrDefault` and then `put` to atomically get/initialize the `AtomicInteger` and then atomically mutate it.

Same logic for `remove`.

pcdavid (approved):

PR #414, opened 2021-03-26
[397] Be more specific when looking for potential domain or view documents

pcdavid:
@sbegaudeau I'm not entirely sure about the new API this introduces: `Iterable<DocumentEntity> findAllByType(String name, String uri)`.

Another option would be to have explicit `findAllDomainDefinitions` and `findAllViewDefinitions`, but that seems overly specific and IMO `IDocumentRepository` should stay agnostic about the kind of documents we store. If tomorrow we want to hadd new kinds which are treated somewhat specially by other parts of the system, we shouldn't have to touch this API every time.

OTOH, the `(String name, String uri)` also leak EMF-specific concepts, even if they are just strings here. Between the name (`findAll**ByType**`) and these arguments, it implicitly introduces a notion of document type identified by name/uri, which we do not have at the moment.

Otherwise the actual technique you suggested works :+1:, and prevents loading & parsing unrelated documents when we are looking for domains or views, so it's also a nice performance improvement.


sbegaudeau:
Truth to be told, this specific way of finding the relevant documents will soon move out of Sirius Components into Sirius Web only. I have no problem with such specific logic to find the relevant documents in Sirius Web. I don't think that Sirius Web needs to be "that" agnostic and it can evolve quite easily.

sbegaudeau (approved):

PR #471, opened 2021-04-29
[441] Add support for custom images in the View DSL 

pcdavid:
With this change, the nodes's style in the View DSL get a new `shape` attribute, which must match the `id` of a custom image stored inside the database. The studio maker does not need to handle these ids, a UI (combo-box) is provided to propose all the available images by name.

`ImagesController` serves these images using a hard-coded prefix to distinguish them from the others.

Note that there is no mechanism at this point for the end-user to add new images, they must be added manually or programmatically by a developer, for example with code like this:

```java
Files.list(sourcePath).forEach(imgPath -> {
    try {
        CustomImageEntity customImageEntity = new CustomImageEntity();
        customImageEntity.setLabel(this.getImageLabel(imgPath));
        customImageEntity.setFileName(imgPath.getFileName().toString());
        customImageEntity.setContent(Files.readAllBytes(imgPath));
        customImageEntity = this.customImageRepository.save(customImageEntity);
    } catch (IOException e) {
        e.printStackTrace();
    }
});
```


backend/sirius-web-services-api/src/main/java/org/eclipse/sirius/web/services/api/images/ICustomImagesService.java
```diff
@@ -0,0 +1,52 @@
...
+        @Override
+        public List<CustomImage> getAvailableImages() {
+            return List.of();
+        }
+
+        @Override
+        public Optional<CustomImage> findById(UUID id) {
+            return Optional.empty();
+        }
+    }
```
sbegaudeau, line 51: Why have an interface containing its implementation?

backend/sirius-web-services-api/src/main/java/org/eclipse/sirius/web/services/api/images/CustomImage.java
```diff
@@ -0,0 +1,63 @@
...
+import org.eclipse.sirius.web.annotations.graphql.GraphQLNonNull;
+import org.eclipse.sirius.web.annotations.graphql.GraphQLObjectType;
+
+/**
+ * Lightweight DTO representing a custom image (without the content).
+ *
+ * @author pcdavid
+ */
+@Immutable
+@GraphQLObjectType
```
sbegaudeau, line 30: Does this need to appear in the GraphQL API? Why?

backend/sirius-web-persistence/src/main/java/org/eclipse/sirius/web/persistence/entities/CustomImageEntity.java
```diff
@@ -0,0 +1,81 @@
...
+@Table(name = "CustomImage")
+public class CustomImageEntity {
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private UUID id;
+
+    private String label;
+
+    @Column(name = "filename")
+    private String fileName;
```
sbegaudeau, line 40: Why do we have both a filename and an id? Why is the filename needed?

- Since we have a label I imagine that the label will be displayed to the end user, not the filename
- Since the identify of the entity is defined by its id, why the filename?

Is it used only to check that we didn't contribute the same image twice for hardcoded images? For those images, the id could be derived from the filename.

backend/sirius-web-persistence/src/main/resources/db/changelog/sirius-web-schema.sql
```diff
@@ -93,6 +93,13 @@ CREATE TABLE Modeler (
     CONSTRAINT pk_modeler_id PRIMARY KEY (id),
     CONSTRAINT fk_modeler_project_id_id FOREIGN KEY (project_id) REFERENCES project(id) ON DELETE CASCADE
 );
+
+CREATE TABLE CustomImage (
+	id UUID DEFAULT gen_random_uuid() NOT NULL,
+	label TEXT NOT NULL,
+	filename TEXT NOT NULL,
+	content BYTEA NOT NULL
```
sbegaudeau, line 101: What is the primary key?

backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/view/ViewPropertiesConfigurer.java
```diff
@@ -0,0 +1,171 @@
...
+        // @formatter:on
+    }
+
+    private SelectDescription createShapeSelectionField() {
+        // @formatter:off
+        return SelectDescription.newSelectDescription("nodestyle.shapeSelector") //$NON-NLS-1$
+                                .idProvider(variableManager -> "nodestyle.shapeSelector") //$NON-NLS-1$
+                                .labelProvider(variableManager -> "Shape") //$NON-NLS-1$
+                                .valueProvider(variableManager -> variableManager.get(VariableManager.SELF, NodeStyle.class).map(NodeStyle::getShape).orElse(EMPTY))
+                                .optionsProvider(variableManager -> this.customImagesService.getAvailableImages().stream().collect(Collectors.toList()))
```
sbegaudeau, line 152: The following call returns a list : `this.customImagesService.getAvailableImages()` so `.stream().collect(Collectors.toList())` is not useful unless I missed something

backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/view/ViewConverter.java
```diff
@@ -160,18 +168,29 @@ private NodeDescription convert(org.eclipse.sirius.web.view.NodeDescription view
...
                 .targetObjectLabelProvider(this.semanticTargetLabelProvider)
                 .semanticElementsProvider(this.getSemanticElementsProvider(viewNodeDescription))
                 .synchronizationPolicy(synchronizationPolicy)
-                .typeProvider(variableManager -> NodeType.NODE_RECTANGLE)
+                .typeProvider(variableManager -> nodeType)
                 .labelDescription(this.getLabelDescription(viewNodeDescription))
-                .styleProvider(variableManager -> this.stylesFactory.createNodeStyle(viewNodeDescription.getStyle().getColor()))
+                .styleProvider(variableManager -> {
+                    String shapeId = viewNodeDescription.getStyle().getShape();
+                    String shapeFileName = this.customImagesService.findById(UUID.fromString(shapeId)).map(CustomImage::getFileName).orElse(DEFAULT_SHAPE_FILE);
```
sbegaudeau, line 188: I'm still puzzled with this id + filename identity. Why aren't you using the id here? The id could be used as the sole identity of the image everywhere without using the filename at all. Since it's not displayed to the end user, we could remove it entirely.

backend/sirius-web-services-api/src/main/java/org/eclipse/sirius/web/services/api/images/ICustomImagesService.java
pcdavid: It was kind of a test/proposal to avoid the current situation where we have to create multiple NoOp implementations of our APIs (e.g. 2 verions of `NoOpEditingContext`, 3 `NoOpEditService`, etc.).
Granted, in this particular case this is used at a single location, but it was more to see your reaction about this approach :smiley: 
I can use the current pattern instead and move `NoOpCustomImageService` in the single location it is used at the moment (close to `ImagesControllerTestCases`).

backend/sirius-web-persistence/src/main/resources/db/changelog/sirius-web-schema.sql
pcdavid: The id. I forgot to complete the schema after experimenting (hence the DRAFT).
I'll add it.

backend/sirius-web-persistence/src/main/java/org/eclipse/sirius/web/persistence/entities/CustomImageEntity.java
pcdavid: I hesitated about this. Having the filename was mostly to have "cleaner" URLs when accessing the images from `ImagesController`, but we probably don't care.
One issue though is that currently the filename is used to identify the file type and this the content-type (via the extension).
If we only keep the id, label and content, direct URLs would by of the form `/api/images/custom/${someUUID}` how would `ImageController` determine the content-type?

Or we can add a `content-type` column directly in the table and set it on import, but it feels like adding a subset of the information we already have with the filename (through the extension).

Or we can decide to only support SVGs, but that seems wrong.

backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/view/ViewPropertiesConfigurer.java
pcdavid: I get a `Type mismatch: cannot convert from List<CustomImage> to List<Object>` without the collect. Anyway, I've modified this line a little since to return the images sorted by label, so the intermediate step is needed anyway.

backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/view/ViewConverter.java
pcdavid: I know, the current design is problematic.
I added a `id UUID` because it's the convention we follow, and I guess it's needed by Hibernate? Not sure.
But I also needed _something_ to be able to determine the file type (and thus which `MediaType` to return in `ImageController`), hence the `fileName` (yes, just the image type/extension would be enough).

backend/sirius-web-services-api/src/main/java/org/eclipse/sirius/web/services/api/images/CustomImage.java
pcdavid: Careless copy/paste. I'll remove it.

backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/view/ViewConverter.java
pcdavid: Another issue with using the id is that they are randomly generated (at least how we do it in other entities), and this means re-importing images breaks any existing diagrams that refer to the id in some of hteir nodes' styles.
This would also make such diagrams non-portable between servers (upload + download) even if they have the "same" custom images imported, as they would have different ids on different instances.

pcdavid:
So, I agree that the current situation with both the `id` and the `fileName` is ugly/broken. However, it seems to me that if we want to keep only one, it should be `fileName`:
1. It makes images URIs cleaner/more readable. It's a very marginal gain as in practice no user will probably ever see these URLs, but it could make debugging easier for us.
2. having the full file name allows us to determine the content-type (through the extension)
3. more importantly, it makes the URLs and primary key of the custom images stable. Currently if we re-import the same images on the server, they will get different random ids and break any models which reference the old ids that could have been exported before. This also make such models non-portable across instances via download/upload, as even if the target instance has the "same" custom images installed, they would have diffrent ids.

sbegaudeau:
I agree that the situation is broken but I think we should only keep the id **but definitely not right now** (only filename or both is ok for me right now).

- I think that the URL should not contain the filename because when users will be able to upload their own images, it may reveal some sensitive information. On top of that, we would have to handle conflicts if two users upload an image with the same name. We really really don't care about those issues right now which is why I'm fine with filename for now. We already have [this issue](https://github.com/eclipse-sirius/sirius-components/issues/115) for EMF icons and [a fix was started by Hugo](https://github.com/eclipse-sirius/sirius-components/commit/18749602b020e7e0d6779ecf20536f3ed1b0029b) to switch to opaque URLs entirely (IIRC one issue with his solution was that it was stateful which can't really work).
- That's what is convincing me the most to keep the filename for now. We should not determine the content type this way and we should instead rely on `Files.probeContentType` or `URLConnection.getContentType` or something like that because the filename may not represent the proper content type of the file but now is not the time to change that.
- We have not really had the time to discuss on the lifecycle of those images but I'm not sure that we want to allow reimporting the same images (it may only be a development issue, in this case we could compute the UUID of those resources from their filename to make it stable in this specific use case).

pcdavid:
> we could compute the UUID of those resources from their filename to make it stable in this specific use case).

This would cause the same possible conflicts as you mention in the case we rely on user-supplief file names though.

mbats (approved):

PR #506, opened 2021-05-21
🧹 [cleanup] Remove unused dependencies in the frontend

pcdavid:
- [cleanup] Remove storybook dependency
- [cleanup] Remove Cypress dependency
- [cleanup] Remove other unused dependencies


sbegaudeau (changes requested):
While I understand the removal of Storybook, Cypress, SVG URL Loader and terser, I'm not completely convinced by the jest one since we are still using jest-dom. Why remove the jest typescript definition while keeping jest based tests and thus jest, jest-junit-reporter, etc. We could use it more and should probably do so. I think, I'll just update your PR to start merging the other removals.

PR #507, opened 2021-05-26
🐛 [480] Fix inheritance support in DomainConverter

pcdavid:
The previous condition did not account for the sentinel element.


sbegaudeau (approved):
I did not take a good look at this method but I find it odd that we need to make multiple passes to convert types. Why can't we just convert everything and after that add inheritance links? I am not sure to understand what the sentinel is supposed to be either. The code does not seem complex but I don't understand why this approach is needed.

I'll approve the fix but I have not idea if it's correct or not since I don't understand the goal of the method in general.

PR #516, opened 2021-05-27
✨ [510] Add support for rounded corners

pcdavid:


frontend/src/diagram/sprotty/views/ListView.tsx
```diff
@@ -55,6 +55,7 @@ export class ListView extends RectangularNodeView {
       default:
         break;
     }
+    const radius = nodeStyle.borderRadius || 0;
```
sbegaudeau, line 58: how can the field be null when GraphQL says it's required? No need to use ` || 0`

frontend/src/diagram/sprotty/views/RectangleView.tsx
```diff
@@ -61,6 +61,8 @@ export class RectangleView extends RectangularNodeView {
       height: Math.max(0, node.bounds.height),
     };
 
+    const radius = nodeStyle.borderRadius || 0;
```
sbegaudeau, line 64: Same comment, this cannot be null

sbegaudeau (changes requested):

sbegaudeau (approved):

PR #521, opened 2021-05-28
✨ Add support for custom tool definitions

pcdavid:


backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/view/ToolInterpreter.java
```diff
@@ -0,0 +1,200 @@
...
+            }
+        } else {
+            elementsToUnset.stream().forEach(value -> EcoreUtil.remove(context, featureToEdit, value));
+        }
+    }
+
+    private Optional<VariableManager> executeCreateInstance(VariableManager variableManager, CreateInstance creatInstanceOperation) {
+        var optionalSelf = variableManager.get(VariableManager.SELF, EObject.class);
+        if (optionalSelf.isPresent()) {
+            var newInstance = this.ecore.createInstance(creatInstanceOperation.getTypeName());
```
pcdavid, line 180: This does not work in the context of a dynamically create EPackage. I have a patch which uses the same workaround as in `CanonicalBehaviors.createSemanticInstance` instead.

sbegaudeau (approved):

PR #522, opened 2021-05-28
✨ [520] Add support for providing Java services to dynamic modelers

pcdavid:


sbegaudeau (changes requested):
(I'll just review the last commit related to Java services)

If I understand your commit correctly, you are adding to the View DSL the concept of Java extensions (just like in Sirius RCP)? This is not what I understood when we discussed about this topic (I think a small ADR could be useful here to make sure that we are all on the same page).

For me, such support would simply be an API which would let a specifier provide additional Java services to a set of "RepresentationDescription" not an API in which end users would be able to specify Java classes to be loaded from their web browsers. That seems like a very dangerous approach isn't it?

An how could the end user know the qualified name of the Java class on top of that?

pcdavid:
> (I'll just review the last commit related to Java services)
> 
> If I understand your commit correctly, you are adding to the View DSL the concept of Java extensions (just like in Sirius RCP)? This is not what I understood when we discussed about this topic (I think a small ADR could be useful here to make sure that we are all on the same page).

I had a first implementation that did that, but it forces the user who creates a modeler and needs to use such a service to also:
1. be a Java programmer;
2. build and deploy a custom Sirius Web application.

Both of which go against the objective of targeting non-experts for web-based modelers.

Here the idea would have been to propose some generic service classes (e.g. based on java-faker), document them and allow studio makers to choose which ones to add to their modelers.

> For me, such support would simply be an API which would let a specifier provide additional Java services to a set of "RepresentationDescription" not an API in which end users would be able to specify Java classes to be loaded from their web browsers.
>  That seems like a very dangerous approach isn't it?

Agreed.

> An how could the end user know the qualified name of the Java class on top of that?

Documentation?

I'll dig my previous attempt from my reflog and propose that instead.

sbegaudeau:
> > (I'll just review the last commit related to Java services)
> > If I understand your commit correctly, you are adding to the View DSL the concept of Java extensions (just like in Sirius RCP)? This is not what I understood when we discussed about this topic (I think a small ADR could be useful here to make sure that we are all on the same page).
> 
> I had a first implementation that did that, but it forces the user who creates a modeler and needs to use such a service to also:
> 
> 1. be a Java programmer;
> 2. build and deploy a custom Sirius Web application.
> 
> Both of which go against the objective of targeting non-experts for web-based modelers.

In both visions, the Java services will have to be written by a Java programmer but we need to decide who will chose where and when such service will be available. Since the lifecycle of the Java services cannot be the same as the lifecycle of the Domain I fail to see use cases in which this feature will really be useful but that's another topic. When the Java service will be written it can only be using generic EMF APIs (EObject, eSet, eGet, etc) and it will not be able to make any assumptions on the Domain.

> Here the idea would have been to propose some generic service classes (e.g. based on java-faker), document them and allow studio makers to choose which ones to add to their modelers.

We could simply create an API to add services to every representations. We could simply reuse the same pattern as we did with `emfServices` and the `input` variable in EEF to make such services available. Something like `faker.doStuff()`. Could we force it to contribute "general purpose Java services"?

> > For me, such support would simply be an API which would let a specifier provide additional Java services to a set of "RepresentationDescription" not an API in which end users would be able to specify Java classes to be loaded from their web browsers.
> > That seems like a very dangerous approach isn't it?
> 
> Agreed.

I am glad to see that we agree on that 😄 

> > An how could the end user know the qualified name of the Java class on top of that?
> 
> Documentation?
> 
> I'll dig my previous attempt from my reflog and propose that instead.

The only issue that I have is to ask the end user for the qualified name of a Java class. If at least it was an API to select something in a predetermined list it would be better.

sbegaudeau:
Could something like this work, let's say I create a Java service like that:

```
@JavaService
public class Faker {
    public String getName(EObject eObject) {
        return "random stuff";
    }
}
```

and we would register it in a way which would create the following API in AQL `aql:faker.getName(self)` or something like that?

sbegaudeau (approved):

PR #525, opened 2021-05-31
✨ [512] View DSL: Add support for layout mode configuration

pcdavid:


sbegaudeau (approved):
It seems ok but a rebase is necessary due to some conflicts on the metamodel (the code needs to be generated once again)

PR #531, opened 2021-06-02
[fix] Force the use of guava 30

gcoutable:
### Type of this PR 

- [x] Bug fix
- [x] Build/Releng

### What does this PR do?

In its plug-in dependencies, ELK requires guava [15,19), but, since a
recent version ELK uses guava Streams which have been introduced in
guava 21.
To prevent ClassNotFoundException to occur because of guava Streams we
force the use of guava 30.

We will be able to revert this commit when ELK 0.8.0 will be released

### DIFF
--- a/backend/sirius-web-diagrams-layout/pom.xml
+++ b/backend/sirius-web-diagrams-layout/pom.xml
@@ -38,8 +38,22 @@
 			<url>https://maven.pkg.github.com/eclipse-sirius/sirius-components</url>
 		</repository>
 	</distributionManagement>
+	
+	<dependencyManagement>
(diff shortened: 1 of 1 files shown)


sbegaudeau (changes requested):
Could you make sure to use the same whitespaces as the rest of the file? (I think it's tab but in your PR there is a mix of tabs and spaces)

gcoutable:
> 
> 
> Could you make sure to use the same whitespaces as the rest of the file? (I think it's tab but in your PR there is a mix of tabs and spaces)

Done

sbegaudeau (approved):

PR #533, opened 2021-06-02
[532] Improve explorer child creation

gcoutable:
### Type of this PR 

- [x] Bug fix

### Issue(s)

https://github.com/eclipse-sirius/sirius-components/issues/532

### What does this PR do?

- Makes the CreateChildEventHandler returning an ErrorPayload when the
child that has been created has no owner.
- Handles ErrorPayload in NewObjectModal

### How to test this PR?

- [x] Manual Test:

1. Creates a view model, a diagram description, and a node description
2. Creates a first node style under the node description
3. Tries to create a second node style under the  same node description

The modal used to create the node description child should still be displayed, and the Material-UI Snackbar should appear with an error message.

### DIFF
--- a/backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/services/EditService.java
+++ b/backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/services/EditService.java
@@ -202,7 +202,10 @@ public class EditService implements IEditService {
         editingDomain.getCommandStack().execute(createChildCommand);
         Collection<?> result = createChildCommand.getResult();
         if (result.size() == 1) {
-            objectOptional = Optional.of(result.iterator().next());
+            Object child = result.iterator().next();
+            if (child instanceof EObject && EcoreUtil.isAncestor(eObject, (EObject) child)) {
+                objectOptional = Optional.of(child);
+            }
         }
         return objectOptional;
     }
--- a/frontend/src/modals/new-object/NewObjectModal.tsx
+++ b/frontend/src/modals/new-object/NewObjectModal.tsx
@@ -27,6 +27,8 @@ import { useMachine } from '@xstate/react';
 import gql from 'graphql-tag';
 import {
   GQLCreateChildMutationData,
+  GQLCreateChildPayload,
+  GQLErrorPayload,
   GQLGetChildCreationDescriptionsQueryData,
   GQLGetChildCreationDescriptionsQueryVariables,
   NewObjectModalProps,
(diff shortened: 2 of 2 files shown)


backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/services/EditService.java
```diff
@@ -202,7 +202,10 @@ public EditService(IEditingContextEPackageService editingContextEPackageService,
         editingDomain.getCommandStack().execute(createChildCommand);
         Collection<?> result = createChildCommand.getResult();
         if (result.size() == 1) {
-            objectOptional = Optional.of(result.iterator().next());
+            Object child = result.iterator().next();
+            if (child instanceof EObject && EcoreUtil.isAncestor(eObject, (EObject) child)) {
+                objectOptional = Optional.of(child);
+            }
```
sbegaudeau, line 208: Shouldn't the child be deleted in case it is not an ancestor?

sbegaudeau:
I have removed the reference to "project" from the PR and commit since the explorer may be used without the concept of project.

backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/services/EditService.java
gcoutable: I don't think, because the "child" is a blank EObject (in our case a blank NodeStyle) which has been provided by AdapterFactoryEditingDomain#getNewChildDescriptors (NodeDescriptionItemprovider#collectNewChildDescriptor) and held by the CommandParameter.
Then, the "child" is given to the CreateChildCommand which only adds, sets, or does nothing whether the feature is multi-valued, single-valued, or already set for a single-valued feature.

gcoutable: So, if we have to delete the child, we also need to delete other potential "children" provided by AdapterFactoryEditingDomain#getNewChildDescriptors.

sbegaudeau: So you are saying that the child is not in the resource at all?

gcoutable: If the create child command did nothing, `child.eContainer() == null`, so in that case, yes, the child is not in a resource, and thus, we should return an ErrorPayload.
If the create child command did add or did set (depending on the feature bounds), `child.eContainer() == eObject`, meaning the child has been "created", and thus, we should return a CreateChildSuccessPayload.

sbegaudeau (approved):

PR #534, opened 2021-06-03
[fix] Fix label padding for nodes

gcoutable:
### Type of this PR 

- [x] Bug fix

### What does this PR do?

It fixes the node label padding for the incremental layout to be the same as the automatic layout (ELK).


pcdavid (changes requested):
I get strange inconsistencies with the patch, with nodes changing size when moving another node and then getting their original size back (which seems too high to me) after an "Arrange All".

https://user-images.githubusercontent.com/10608/120650886-7d2da980-c47e-11eb-870a-ed9c7e674f80.mp4

pcdavid (approved):

PR #542, opened 2021-06-08
[fix] Fix open existing representation from onboarding area

gcoutable:


sbegaudeau (approved):
@pcdavid should try to run this code but it makes sense to have to change this to fix the bug

PR #545, opened 2021-06-10
Use nicer SVG icons for Domain and View elements

pcdavid:
### Type of this PR 

### What does this PR do?

Use distinct and nice-looking SVG icons for the View and Domain DSL elements and tools, and also for default elements (including dynamic instances) where we used a not-so-nice GIF.


sbegaudeau (approved):
I have not executed the code but it looks good

PR #553, opened 2021-06-16
[515] Build artifact's sources

gcoutable:


pcdavid:
It's probably OK given that they do not contain any Java source, but for the record the projects `sirius-web-graphiql` and `sirius-web-graphql-voyager` do not produce source JARs with this PR.

sbegaudeau (approved):

PR #573, opened 2021-06-21
[572] Move the object type name in a tooltip for tree items

pcdavid:
Make the labels more compact and readable, while still preserving access to the type information when needed.

* Left: before (e.g. "Processor Radar_Capture")
* Right: after ("Radar_Capture" with "Processor" in the tooltip)

compact-labels


sbegaudeau:
I've changed the title of the tree item since by displaying the kind directly could show the concept `Document` in the user interface.

sbegaudeau (approved):

PR #581, opened 2021-06-23
[releng] Bump version to v0.3.2

sbegaudeau:


sbegaudeau:
A bug has occurred during the deployment of the v0.3.1 and since we can't retry to publish this version, we have to switch to v0.3.2

gcoutable (approved):

PR #624, opened 2021-07-08
Avoid resetting edge's routing in some cases when it is not needed

pcdavid:
- [550] Consider the same padding as ELK's in the incremental layout
- [550] Refactor for clarity
- [550] Only change container size if it is too small to hold its content

### DIFF
--- a/backend/sirius-web-diagrams-layout/src/main/java/org/eclipse/sirius/web/diagrams/layout/incremental/updater/ContainmentUpdater.java
+++ b/backend/sirius-web-diagrams-layout/src/main/java/org/eclipse/sirius/web/diagrams/layout/incremental/updater/ContainmentUpdater.java
@@ -215,7 +215,7 @@ public class ContainmentUpdater {
     }
 
     private double getPaddingHeight(IContainerLayoutData container) {
-        double nodeLabelPaddingHeight = 0;
+        double nodeLabelPaddingHeight = LayoutOptionValues.DEFAULT_ELK_PADDING;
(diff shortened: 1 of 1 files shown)


backend/sirius-web-diagrams-layout/src/main/java/org/eclipse/sirius/web/diagrams/layout/incremental/updater/ContainmentUpdater.java
```diff
@@ -158,24 +158,11 @@ private void updateNodeListItem(NodeLayoutData nodeLayoutData) {
...
-        for (NodeLayoutData child : container.getChildrenNodes()) {
-            maxWidth = Math.max(maxWidth, child.getPosition().getX() + child.getSize().getWidth() + this.getPaddingWidth(container));
-            maxHeight = Math.max(maxHeight, child.getPosition().getY() + child.getSize().getHeight() + this.getPaddingHeight(container));
-        }
-
-        if (maxWidth != width || maxHeight != height) {
-            container.setSize(Size.of(maxWidth, maxHeight));
+        // Only update the conttainer's size if it is too small for its content.
+        if (container.getSize().getWidth() < contentSize.getWidth() || container.getSize().getHeight() < contentSize.getHeight()) {
+            container.setSize(contentSize);
```
gcoutable, line 165: I think a NodeList should be resized if the sum of its NodeListItems size is smaller than the NodeList size.

sbegaudeau (approved):

PR #626, opened 2021-07-08
Use a Select widget to display EEnum properties

pcdavid:
- [625] Use a Select widget to display EEnum properties
- [625] Only show 'None' as a value for optional properties

### Type of this PR 

- [ ] Bug fix
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

...

### What does this PR do?

...

...
 

### Potential side effects

...

### How to test this PR?

- [ ] Frontend Unit tests
- [ ] Backend Unit tests

### Checklist

- [ ] All tests pass.

### DIFF
--- a/backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/compatibility/properties/EEnumIfDescriptionProvider.java
+++ b/backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/compatibility/properties/EEnumIfDescriptionProvider.java
@@ -26,6 +26,7 @@ import org.eclipse.emf.ecore.EDataType;
 import org.eclipse.emf.ecore.EEnum;
 import org.eclipse.emf.ecore.EEnumLiteral;
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.EStructuralFeature;
 import org.eclipse.emf.ecore.util.EcoreUtil;
 import org.eclipse.emf.edit.provider.ComposedAdapterFactory;
 import org.eclipse.sirius.web.compat.forms.WidgetIdProvider;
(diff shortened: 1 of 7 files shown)


backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/compatibility/properties/EEnumIfDescriptionProvider.java
```diff
@@ -83,13 +84,25 @@ private SelectDescription getSelectDescription() {
                 .labelProvider(this.getLabelProvider())
                 .valueProvider(this.getValueProvider())
                 .optionsProvider(this.getOptionsProvider())
+                .valueRequiredProvider(this.getValueRequiredProvider())
```
sbegaudeau, line 87: Do we need to have such special use case, couldn't it be done using simply another value in the value provider and thus without hardcoding "null" in the frontend? I've always believed that this hardcoded "null" was odd and while this change may work, it may be a hack on top of a previous hack.

sbegaudeau (changes requested):

backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/compatibility/properties/EEnumIfDescriptionProvider.java
pcdavid: I don't know. It seems to me that _unsetting_ a feature is fundamentally different from setting it to a value, even a "special" one. Even on the frontend it wouldn't shock me to have a clearly distinct way to indicate that a property is _not set_ and to _unset_ it (not that I'm advocating for it right now).

See for example text fields, where it is impossible to distinguish in the UI if the value is unset of set explicitly to `""`. And once someone has set a value to a text attribute once it is impossible in the UI to unset it (you can only set it to empty, but that is not always the same thing).

I know it's a different widget that this one, but it illustrates that IMO "no value" is something semantically very different from "a value which can be interpreted as 'empty'", and the UI should support disinguishing between the two.

PR #631, opened 2021-07-16
Disable move/resize of nodes on auto-layout diagrams

pcdavid:
- [514] Disable move/resize of nodes on auto-layout diagrams
- [514] Disable the arrange all button for auto-layouted diagrams


backend/sirius-web-diagrams/src/main/java/org/eclipse/sirius/web/diagrams/Diagram.java
```diff
@@ -50,6 +50,8 @@
 
     private Size size;
 
+    private boolean autoLayout;
```
sbegaudeau, line 53: Why provide this info on the GraphQL API of the diagram POJO while it already exists on the GraphQL API of DiagraDescription?

sbegaudeau (changes requested):

sbegaudeau (approved):

PR #681, opened 2021-08-09
[doc] Add various ADRs to the 0.5.0 release of Sirius Components

sbegaudeau:

### DIFF
--- a/doc/adrs/024_add_support_for_view_deletion_requests.adoc
+++ b/doc/adrs/024_add_support_for_view_deletion_requests.adoc
@@ -1,0 +1,50 @@
+= ADR-024 - Add support for view deletion requests
+
+== Context
+
+We have added support for the creation of unsynchronized nodes but we did not add the ability to delete those nodes without deleting the semantic element used to create them.
+For that, we will need to add a new concept, the `ViewDeletionRequest`.
+
+== Decision
+
+A new concept, named `ViewDeletionRequest` will be introduced in order to indicate that a specific node should be deleted from the next rendering without any impact on the semantic elements.
+It will have the following API:
+
+```
+@Immutable
+public final class ViewDeletionRequest {
+
+    private UUID elementId;
+
+    private ViewDeletionRequest() {
+        // Prevent instantiation
+    }
+
+    public UUID getElementId() {
+        return this.elementId;
+    }
+
+    public static Builder newViewDeletionRequest() {
+        return new Builder();
+    }
+
+    // Additional code...
+}
+```
+
+The identifier of the element will be used in order to match it in the `NodeComponent` to make sure that the node will not be rendered during the next rendering.
+The method `org.eclipse.sirius.web.diagrams.components.NodeComponent.shouldRender(String, Optional<Node>)` should be updated in order to take care of a potential view deletion request.
+In this method we will have to compare all the given `elementId` with the identifier of the previously rendered node to find out if we still need to render it.
+
+The props of the `NodeComponent` and the `DiagramComponent` will thus be updated to included the `viewDeletionRequests` to perform.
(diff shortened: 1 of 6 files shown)


doc/adrs/024_add_support_for_view_deletion_requests.adoc
```diff
@@ -0,0 +1,50 @@
...
+```
+
+The identifier of the element will be used in order to match it in the `NodeComponent` to make sure that the node will not be rendered during the next rendering.
+The method `org.eclipse.sirius.web.diagrams.components.NodeComponent.shouldRender(String, Optional<Node>)` should be updated in order to take care of a potential view deletion request.
+In this method we will have to compare all the given `elementId` with the identifier of the previously rendered node to find out if we still need to render it.
+
+The props of the `NodeComponent` and the `DiagramComponent` will thus be updated to included the `viewDeletionRequests` to perform.
+The `IDiagramContext` will also need to be updated to include said `ViewDeletionRequest`.
+
+We may need to clean existing `ViewCreationRequest` and `ViewDeletionRequest` from the `DiagramContext` when the latter is reseted.
```
gcoutable, line 42: `ViewCreationRequest`: Does that mean it may have more than one view element for one semantic element?

doc/adrs/027_make_the_explorer_independent_from_modeling_concepts.adoc
```diff
@@ -0,0 +1,76 @@
...
+
+Those various use cases may require interaction with a server and some custom user interface.
+The same concerns are relevant to support downloading documents today (custom user interface + custom endpoint).
+
+In order to support those use cases, we will need to have the ability to contribute custom entries in the context menus of the TreeItem.
+In a similar fashion as the ability to contribute additional representations, the Sirius Components frontend API will provide the ability to add entries in the context menu.
+Several examples will be provided by Sirius Web:
+
+- A direct download link to download a document
+- An entry to create new objects thanks to the `NewObjectModal`
```
gcoutable, line 46: You wrote right above that "The creation of a new child may require specific information". How will the `NewObjectModal` ask for this specific information?

gcoutable (changes requested):

doc/adrs/027_make_the_explorer_independent_from_modeling_concepts.adoc
sbegaudeau: All I am saying is that creating child right now requires a dedicated user interface with some dedicated GraphQL endpoints etc. Thus, we can't and won't provide a generic way to create child but instead let anyone provide their own user interface to provide, for example, the current NewObjectModal, the NewRepresentationModal, etc. Right now, the NewObjectModal ask for the child creation descriptions for example.

doc/adrs/025_switch_to_a_schema_first_approach_for_our_graphql_schema.adoc
```diff
@@ -0,0 +1,67 @@
...
+As a result, it will slow down our ability to make massive changes in the GraphQL schema, on purpose, and make evolutions more visible.
+
+It will also dramatically reduce the relevance of our custom GraphQL / Spring integration and give us the ability to leverage other GraphQL tools and frameworks more easily.
+
+== Decision
+
+=== Core Schema
+
+We will create a file named `core.graphqls` containing all the core concepts of the GraphQL schema used to manipulate Sirius Components.
+This file will contain all the concepts related to our representations (Diagram, Form, Tree, Properties, Selection, Validation and more), the of editing context and the viewer at least.
```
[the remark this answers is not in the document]
sbegaudeau: Fixed in a second commit

gcoutable (approved):

PR #749, opened 2021-09-22
[746] Delay form event processor disposal to avoid lost events

pcdavid:
Just a draft for now so that the existing workaround can be discussed.
* The delay is hard coded, and the choice of 5 seconds is arbitrary.
* The change only is only applied to Forms, but if we want to to apply more globally it should be move up into the stack.


pcdavid:
Maybe in `EditingContextEventProcessor.acquireRepresentationEventProcessor()`:
```java
Disposable subscription = representationEventProcessor.canBeDisposed().delayElements(Duration.ofSeconds(5)).subscribe(canBeDisposed -> {
    if (canBeDisposed.booleanValue()) {
        this.disposeRepresentation(configuration.getId());
    }
});
```
I have not tested.

sbegaudeau:
I'm not merging this review since I found two issues, I fixed one but I still have to think about the other one.

gcoutable (approved):

PR #836, opened 2021-11-17
[822] Add delete from diagram generic tool in diagrams

AxelRICHARD:
### Type of this PR 

### Issue(s)

Bug: https://github.com/eclipse-sirius/sirius-components/issues/822

### DIFF
--- a/backend/sirius-web-spring-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
+++ b/backend/sirius-web-spring-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
@@ -27,6 +27,7 @@ type Diagram implements Representation {
   nodes: [Node!]!
   edges: [Edge!]!
   toolSections: [ToolSection!]!
+  unsynchronizedDiagramElementsDescriptionIds: [ID!]!
 }
 
 type Node {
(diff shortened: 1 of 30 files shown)


backend/sirius-web-spring-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
```diff
@@ -27,6 +27,7 @@ type Diagram implements Representation {
   nodes: [Node!]!
   edges: [Edge!]!
   toolSections: [ToolSection!]!
+  unsynchronizedDiagramElementsDescriptionIds: [ID!]!
```
sbegaudeau, line 30: As discussed this morning, instead of adding such field, it would be better to have the relevant information on `NodeDescription`. The current approach has several drawbacks:

- Since it is a computed information, it requires everyone using Sirius Components to provide a data fetcher with this behavior
- We don't want to start adding custom fields for small use cases like that

frontend/src/workbench/Workbench.types.ts
```diff
@@ -16,6 +16,7 @@ export type Selection = {
   id: string;
   label: string;
   kind: string;
+  descriptionId?: string;
```
sbegaudeau, line 19: As discussed this morning too, this should wait for the contribution related to the ADR-36 in order to leverage some additional information on the kind of a graphical selection for example.

sbegaudeau (changes requested):
I have not performed an extensive review but there are two issues which should be re-evaluated after the integration of the work on metadata and on the `Selection`

PR #848, opened 2021-11-30
Fix the diagram selection that should be updated while the workbench selection is changed.

gcoutable:


sbegaudeau (approved):

sbegaudeau:
With this PR we will revert to the old behavior but it may be temporary since the old code is quite odd

PR #854, opened 2021-12-02
Use multi-selection api in workbench

gcoutable:
### Type of this PR 

- [ ] Bug fix
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

https://github.com/eclipse-sirius/sirius-components/issues/134

### What does this PR do?
This commit does not add behavior, it only prepares the workbench and
all workbench components using the selection to support the multiple
selection.

### DIFF
--- a/frontend/src/onboarding/NewRepresentationArea.tsx
+++ b/frontend/src/onboarding/NewRepresentationArea.tsx
@@ -52,6 +52,7 @@ export const NewRepresentationArea = ({
   };
   const [state, setState] = useState(initialState);
   const { message } = state;
+  const selectedItem = selection.entries.length > 0 ? selection.entries[0] : null;
 
   // Representation creation
   const [createRepresentation, { loading, data, error }] = useMutation(createRepresentationMutation);
@@ -60,7 +61,7 @@ export const NewRepresentationArea = ({
       const { createRepresentation } = data;
       if (createRepresentation.representation) {
         const { id, label, kind } = createRepresentation.representation;
-        setSelection({ id, label, kind });
+        setSelection({ entries: [{ id, label, kind }] });
       } else if (createRepresentation.__typename === 'ErrorPayload') {
         setState((prevState) => {
           const newState = { ...prevState };
@@ -72,7 +73,7 @@ export const NewRepresentationArea = ({
   }, [loading, data, error, setSelection]);
   const onCreateRepresentation = (representationDescriptionId) => {
     const selected = representationDescriptions.find((candidate) => candidate.id === representationDescriptionId);
-    const objectId = selection.id;
+    const objectId = selectedItem.id;
     const input = {
       id: uuid(),
       editingContextId,
@@ -121,8 +122,8 @@ export const NewRepresentationArea = ({
     return <AreaContainer title={title} subtitle="You need edit access to create representations" />;
   } else {
     let subtitle =
-      selection && representationDescriptions.length > 0
-        ? 'Select the representation to create on ' + selection.label
+      selectedItem && representationDescriptions.length > 0
+        ? 'Select the representation to create on ' + selection.entries[0].label
         : 'There are no representations available for the current selection';
     return (
       <AreaContainer title={title} subtitle={subtitle} banner={message}>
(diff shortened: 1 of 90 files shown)


frontend/src/onboarding/NewRepresentationArea.tsx
```diff
@@ -121,8 +122,8 @@ export const NewRepresentationArea = ({
     return <AreaContainer title={title} subtitle="You need edit access to create representations" />;
   } else {
     let subtitle =
-      selection && representationDescriptions.length > 0
-        ? 'Select the representation to create on ' + selection.label
+      selectedItem && representationDescriptions.length > 0
+        ? 'Select the representation to create on ' + selection.entries[0].label
```
sbegaudeau, line 126: `selectedItem.label` to stay coherent but it works

frontend/src/workbench/Workbench.tsx
```diff
@@ -62,8 +62,14 @@ export const Workbench = ({
   const { selection, representations, displayedRepresentation } = context;
 
   const setSelection = (selection: Selection) => {
-    const isRepresentation = registry.isRepresentation(selection.kind);
-    const updateSelectionEvent: UpdateSelectionEvent = { type: 'UPDATE_SELECTION', selection, isRepresentation };
+    const representations: Representation[] = selection?.entries.filter((entry) =>
```
sbegaudeau, line 65: no need for `?` the selection cannot be null here, can it?

frontend/src/workbench/WorkbenchMachine.ts
```diff
@@ -20,21 +20,25 @@ export interface WorkbenchStateSchema {
...
+  representations: Representation[];
+};
 export type WorkbenchEvent = UpdateSelectionEvent | ShowRepresentationEvent | HideRepresentationEvent;
 
 export const workbenchMachine = Machine<WorkbenchContext, WorkbenchStateSchema, WorkbenchEvent>(
   {
     initial: 'initial',
     context: {
-      selection: null,
+      selection: { entries: [] },
```
sbegaudeau, line 41: Now that the object `selection` on which everything is connected cannot be `null` anymore, there should not be any occurrences of `selection?` in the code again. There are still some in your PR.

sbegaudeau (changes requested):
You just need to review usages of `selection?` and I'll need to test that with a complete product like Sirius Web but otherwise everything seems ok. Could you please update the `CHANGELOG` too? Thanks

frontend/src/onboarding/NewRepresentationArea.tsx
gcoutable: Done

frontend/src/workbench/Workbench.tsx
gcoutable: It should not. Removed.

frontend/src/workbench/WorkbenchMachine.ts
gcoutable: Done

sbegaudeau:
I'll need to review the impact on consumers like Sirius Web and add an entry in the CHANGELOG but it looks good.

sbegaudeau:
There seems to have a very minor change in behavior in the lifecycle of the details view. I'll investigate.

frontend/src/properties/PropertiesWebSocketContainer.tsx
```diff
@@ -88,9 +88,13 @@ export const PropertiesWebSocketContainer = ({
...
-    if (currentSelection?.id !== selection?.id) {
-      const isRepresentation = registry.isRepresentation(selection.kind);
-      const switchSelectionEvent: SwitchSelectionEvent = { type: 'SWITCH_SELECTION', selection, isRepresentation };
+    if (selection.entries.length > 0 && selection.entries[0].id !== currentSelection?.id) {
+      const isRepresentation = registry.isRepresentation(selection.entries[0].kind);
+      const switchSelectionEvent: SwitchSelectionEvent = {
+        type: 'SWITCH_SELECTION',
+        selection: selection.entries[0],
+        isRepresentation,
+      };
```
sbegaudeau, line 97: The current selection of the PropertiesWebSocketContainer cannot be null anymore so it does not have the same behavior when there is no selection

sbegaudeau:
I'll close the PR since it is not made against `master` and as such it is not relevant. We will recreate the PR from scratch and merge it.

PR #872, opened 2021-12-06
[871] Allow the customization of the executor service of the editing context

sbegaudeau:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/871

### DIFF
--- a/backend/sirius-web-spring-collaborative/src/main/java/org/eclipse/sirius/web/spring/collaborative/projects/EditingContextEventProcessorParameters.java
+++ b/backend/sirius-web-spring-collaborative/src/main/java/org/eclipse/sirius/web/spring/collaborative/projects/EditingContextEventProcessorParameters.java
@@ -1,0 +1,171 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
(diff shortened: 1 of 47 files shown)


backend/sirius-web-spring-collaborative/src/main/java/org/eclipse/sirius/web/spring/collaborative/projects/EditingContextEventProcessorParameters.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        public Builder executorServiceProvider(IEditingContextEventProcessorExecutorServiceProvider executorServiceProvider) {
+            this.executorServiceProvider = Objects.requireNonNull(executorServiceProvider);
+            return this;
+        }
+
+        public EditingContextEventProcessorParameters build() {
+            EditingContextEventProcessorParameters parameters = new EditingContextEventProcessorParameters();
+            parameters.messageService = Objects.requireNonNull(this.messageService);
```
Camork, line 160: Why there needs the not-null assertion when it is never null?

sbegaudeau: Someone could forget to set it, it would thus be null

gcoutable (commented):
It may need an entry in the changelog

sbegaudeau:
I have updated the PR with an entry in the CHANGELOG

gcoutable (approved):

PR #883, opened 2021-12-14
View DSL: Add support for defining border nodes

pcdavid:
This is on top of https://github.com/eclipse-sirius/sirius-components/pull/857, the actual commits specific to this change are the last 3:

- [596] Add border nodes in the View DSL
- [596] Handle border nodes in viewconverter
- [596] Update the CHANGELOG


sbegaudeau (approved):
I'll merge it once the build is green. It works nicely but it shows that we have some improvements to make on layouting border nodes. It will help me review the other PRs on that topic 👍

PR #832, opened 2021-11-16
[359] Fix border node's size computation

Camork:
Signed-off-by: Charles Wu <camork1@gmail.com>

### Type of this PR 

- [x] Bug fix
- [ ] New feature or improvement
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

The border node size  is computed incorrectly during incremental layout process.

### What does this PR do?

The border node no need to add padding size, So this PR remove this padding for border node

### Screenshot/screencast of this PR

Below is the screencast that move a node with three border nodes.

https://user-images.githubusercontent.com/19710731/141887023-3a671316-a85f-431d-8192-d923ccde406b.mov

### Potential side effects

...

### How to test this PR?

- [ ] Frontend Unit tests
- [ ] Backend Unit tests
- [ ] Cypress : please specify
- [ ] Manual Test : please specify

### Checklist

- [ ] I have read CONTRIBUTING carefully.
- [ ] I have signed the [Contributor License Agreement](http://www.eclipse.org/legal/CLA.php).
- [ ] All my commits are signed-off (`-s`) with my mail address of my Eclipse Account.
- [ ] I have covered my changes by unit tests or integration tests or manual tests.
- [ ] All tests pass.
- [ ] I have updated the documentation accordingly
- [ ] New React components are availables in storybook

PR #889, opened 2021-12-15
[fix] Fix border node size computation

pcdavid:
I'm not sure yet how this is related to #832.


sbegaudeau (approved):
It's less bad than before :)

PR #892, opened 2021-12-16
[cleanup] Remove log message from Result

pcdavid:
The API is already designed to indicate to the caller if the result
can be interpreted as an int or not. It's up to the client code to
decide what to do if not.

With the log at such a low level, there is no way to avoid "scary"
stack traces in the backend logs in case of a malformed user-supplied
expression for example, which should not be considered even a warning.

### Type of this PR 

- [ ] Bug fix
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

...

### What does this PR do?

...

...
 

### Potential side effects

...

### How to test this PR?

- [ ] Frontend Unit tests
- [ ] Backend Unit tests

### Checklist

- [ ] All tests pass.

### DIFF
--- a/backend/sirius-web-interpreter/src/main/java/org/eclipse/sirius/web/interpreter/Result.java
+++ b/backend/sirius-web-interpreter/src/main/java/org/eclipse/sirius/web/interpreter/Result.java
@@ -21,17 +21,12 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 /**
  * The evaluation result.
  *
  * @author sbegaudeau
  */
 public class Result {
-
-    private Logger logger = LoggerFactory.getLogger(Result.class);
 
     private final Optional<Object> rawValue;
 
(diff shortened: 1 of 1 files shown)


sbegaudeau (changes requested):
I don't see why we would like to remove our ability to log a potential error. One could stop logging this error with the relevant configuration property but I really like the idea of logging everything bad which can happen on the server. This could help detect badly configured expressions.

pcdavid:
Expressions are user-supplied, especially in a dynamic studio setting, and users do not have access to the backend logs. If a user makes a mistake on e.g. a `sizeComputationExpression` which does not return a valid int, I don't see how filling the backend logs with potentially thousands if these stack traces (for every evaluation of the expression) would be helpful to the user, or to the admin.

The calling code, which knows the context of where the expression comes from and for which purpose it is evaluated, is better placed to decide what to do if the result is not an `int`:

* maybe log a warning, yes, in some cases (for example for statically deployed modelers, which can be tested and vetted before being deployed to end-users);
* maybe try to interpret it as something else ("OK, its not an int, maybe it is some other kind of value that I also know how to interpret"), in which case it is was not an error at all;
* maybe surface the error in some user-visible way (which can only depend on the context in which the expression is used, and can not be hard-coded in `Result` itself).

sbegaudeau:
- Expressions can be specifier-supplied if the expressions is coming from the odesign compatibility layer. The specifier can see while building its product that something is wrong and fix it. This is the most popular use case of Sirius Components right now.
- This can be enabled or disabled by the one integrating Sirius Components and running the server with the relevant log configuration. It does not have to fill anything. If it is such an issue that users are filling thousands of lines of log with this exception then it probably means that something is very wrong in the expression or in the way the user interface allows end users to edit the data.

PR #909, opened 2021-12-23
[wip] Switch to a URL based kind for the selection

sbegaudeau:
This change needs to be performed in:
- the explorer description
- the representations descriptions (TO BE DONE)
- the EditingContextRepresentation(s) data fetchers (TO BE DONE on the Sirius Web side)
- the representations in the subscriptions (diagram, representations at least)


sbegaudeau:
I'm creating this PR to save my current work in progress before the holidays

sbegaudeau:
Additional things to document in the CHANGELOG:

- The argument `classId` in the GraphQL schema, never made much sense, it's the `kind` of the selection. It has been renamed in this PR on both the frontend and backend side. It makes things easier to understand
- Switch to both `URLSearchParams` on the frontend side and the newly added `IKindParser` service on the backend side to parse the kind of an object. On the backend side, it will of course rely on `java.net.URI` and `java.net.URLDecoder`

I still have an issue with the retrieval of the tools of a diagram or with the selection in a diagram but it will be easy to fix. Overall I'm almost done with this PR which will allow us to distinguish objects in a more efficient manner.

sbegaudeau:
As noted by @gcoutable, we may use `UriComponentsBuilder` and `UriComponents` from Spring to manipulate URIs but I'll have to review first how much coupling it would add to Spring first. On top of that, we will know before hand the structure of the URIs that we will parse and thus can make some assumptions that Spring cannot.

sbegaudeau:
> I still have an issue with the retrieval of the tools of a diagram or with the selection in a diagram but it will be easy to fix. 

I have updated the PR with a fix

sbegaudeau:
I won't use `UriComponents` for now since it requires adding a dependency to `spring-web` in projects that do not have that dependency directly but I'll have to have another look at that in a second time when I'll be back.

gcoutable (approved):

PR #950, opened 2022-01-12
[949] Fix the position of the begin/end labels on edges

pcdavid:
Fix the way we use ELK to layout begin/end labels.

Before:

After:

The PR also removes some `toString()` calls that are redundant now that the corresponding ids are already strings (instead of UUIDs as they were before).


sbegaudeau (approved):
Waiting for the build to merge it

PR #951, opened 2022-01-12
[943] Add support for begin/end labels on edges

pcdavid:
Note that while not a strict dependency technically, the new feature does not make much sense without the fix in https://github.com/eclipse-sirius/sirius-components/pull/950.


backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/view/ViewConverter.java
```diff
@@ -487,7 +487,9 @@ private EdgeDescription convert(org.eclipse.sirius.web.view.EdgeDescription view
                                      .targetObjectIdProvider(this.semanticTargetIdProvider)
                                      .targetObjectKindProvider(this.semanticTargetKindProvider)
                                      .targetObjectLabelProvider(this.semanticTargetLabelProvider)
-                                     .centerLabelDescription(this.getLabelDescription(viewEdgeDescription, interpreter))
+                                     .centerLabelDescription(this.getSpecificEdgeLabelDescription(viewEdgeDescription, viewEdgeDescription.getLabelExpression(), "_centerlabel", interpreter)) //$NON-NLS-1$
+                                     .beginLabelDescription(this.getSpecificEdgeLabelDescription(viewEdgeDescription, viewEdgeDescription.getBeginLabelExpression(), "_beginlabel", interpreter)) //$NON-NLS-1$
+                                     .endLabelDescription(this.getSpecificEdgeLabelDescription(viewEdgeDescription, viewEdgeDescription.getEndLabelExpression(), "_endlabel", interpreter)) //$NON-NLS-1$
```
sbegaudeau, line 492: Wouldn't this try to create a begin and end label description even if there are no label expressions?

sbegaudeau (changes requested):

backend/sirius-web-emf/src/main/java/org/eclipse/sirius/web/emf/view/ViewConverter.java
pcdavid: Same as for the center one, which is optional. But I'll update the PR to make this cleaner in all 3 cases.

sbegaudeau (approved):
Waiting for the build to merge this PR

PR #954, opened 2022-01-12
[952] The contextual palette can use many lines to display its tools

gcoutable:
### Type of this PR 

- [ ] Bug fix
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

https://github.com/eclipse-sirius/sirius-components/issues/952

### What does this PR do?

This PR prevents the contextual palette to be too long by displaying tools on many lines if needed.

### DIFF
--- a/frontend/src/diagram/palette/ContextualPalette.tsx
+++ b/frontend/src/diagram/palette/ContextualPalette.tsx
@@ -211,7 +211,13 @@ export const ContextualPalette = ({
   );
 
   // Find how many items should be display per line.
-  let itemsPerLine = getNumberOfColumns(contextualToolSections, !!invokeLabelEdit, !!invokeDelete, atLeastOneEdgeTool);
+  let itemsPerLine = getNumberOfColumns(
+    contextualToolSections,
+    !!invokeLabelEdit,
+    !!invokeDelete,
+    isTargetElementSynchronized,
+    atLeastOneEdgeTool
+  );
   const props = { itemsPerLine };
   const classes = useContextualPaletteStyle(props);
 
@@ -296,6 +302,9 @@ export const ContextualPalette = ({
     }
     if (!!invokeDelete) {
       firstlineToolElementToTake--;
+      if (!isTargetElementSynchronized) {
+        firstlineToolElementToTake--;
+      }
     }
     if (!!connectorToolContent) {
       firstlineToolElementToTake--;
@@ -340,6 +349,7 @@ const getNumberOfColumns = (
   toolSections: ToolSectionType[],
   hasDirectEdit: boolean,
   hasDeleteTool: boolean,
+  isTargetElementSynchronized: boolean,
   hasAtLeastOneConnector: boolean
 ): number => {
   let numberOfColumns = toolSections.length + 1;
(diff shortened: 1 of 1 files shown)


CHANGELOG.adoc
```diff
@@ -51,6 +52,7 @@ Currently it is not possible to compute different values for width and height.
 - https://github.com/eclipse-sirius/sirius-components/issues/565[#565] [diagram] Improve the layout of multiple edges between the same nodes
 - https://github.com/eclipse-sirius/sirius-components/issues/914[#914] [diagram] Add the graphical selection to the semantic ones while clicking on diagram elements
 - https://github.com/eclipse-sirius/sirius-components/issues/925[#925] [diagram] Perform a fit to screen after the first render of a diagram
+- https://github.com/eclipse-sirius/sirius-components/issues/952[#952] [diagram] The contextual palette can be displayed with many lines
```
pcdavid, line 55: It doesn't seem precise enough. I *can*, but when/why? Proposal:

> When there are more than 15 tools to display, the contextual palette is split in multiple lines.

sbegaudeau:
Could you rebase this PR please, it will be easier for you to preserve your changes while handling the conflict

sbegaudeau:
I have an issue related to keys with your PR:

```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ContextualPalette`. See https://reactjs.org/link/warning-keys for more information.
    at div
    at ContextualPalette (http://localhost:3000/static/js/vendors~main.chunk.js:23912:35)
    at div
    at div
    at div
    at DropArea (http://localhost:3000/static/js/vendors~main.chunk.js:19553:32)
    at div
```

I still don't know how to reproduce it all the time

frontend/src/diagram/palette/ContextualPalette.tsx
```diff
@@ -207,54 +256,120 @@ export const ContextualPalette = ({
...
+   *   - The remaing elements are tool sections
+   *
+   * All other lines are composed of remaining tool sections.
+   */
+  const toolPaletteContent: JSX.Element[] = [];
+  if ((!!toolSectionsContent && toolSectionsContent.length > 0) || invokeLabelEdit || invokeDelete) {
+    // Fill the first line
+    const firstLine: JSX.Element[] = [];
+    let firstlineToolElementToTake = TOOL_PER_LINE - 1;
+    if (!!invokeLabelEdit) {
```
sbegaudeau, line 300: Why all those `!!`? Why not `if(invokeLabelEdit)`?

frontend/src/diagram/palette/ContextualPalette.tsx
```diff
@@ -207,54 +256,120 @@ export const ContextualPalette = ({
...
+      firstLine.push(renameEntry);
+    }
+    if (!!invokeDelete) {
+      firstLine.push(deleteEntry);
+    }
+    firstLine.push(closeEntry);
+    toolPaletteContent.push(<>{firstLine}</>);
+
+    // Fill other lines
+    while (!!toolSectionsContent && toolSectionsContent.length > 0) {
```
sbegaudeau, line 329: Can toolSectionsContent even be `null` or `undefined` here?

sbegaudeau (changes requested):

sbegaudeau:
Capture d’écran 2022-01-18 à 09 58 50


- I don't have 15 tools per line
- The hardcoded tools do not appear on the first line
- All the tools do not have the same dimension
- The icons are not centered

Personally, I don't really care about this constraint that the hardcoded tools should appear on the first line, It does not make a lot of sense and makes the code really complex for no real benefit. I think we would be way better with a plain simple CSS grid with no magic ordering what so ever.

PR #964, opened 2022-01-14
[958] Use the right edge start and edge end position

gcoutable:
### Type of this PR 

- [x] Bug fix
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

https://github.com/eclipse-sirius/sirius-components/issues/958

### What does this PR do?

Use positions relative to the diagram origin and not positions relative
to the viewport.

### How to test this PR?

- [x] Manual Test : Using the palette, create an edge between two elements. 
It can be using a create edge tool or the connector tool.


sbegaudeau:
I'm not sure why you are changing what we are using to compute the start/end position of the edges but the result is better 👍

sbegaudeau (approved):

PR #989, opened 2022-01-20
New ADRs for the environment variable and the support of the post selection

sbegaudeau:
- [985] Add an ADR for the environment variable
- [988] Provide an ADR for the post selection support

### DIFF
--- a/doc/adrs/039_provide_a_variable_to_detect_the_environment.adoc
+++ b/doc/adrs/039_provide_a_variable_to_detect_the_environment.adoc
@@ -1,0 +1,39 @@
+= ADR-039 - Provide a variable to detect the environment
+
+== Context
+
+We need to be able to distinguish the environment in which our representation descriptions are executed.
+This would help us create odesign which works both in the desktop environment and in a web environment.
+
+== Decision
+
+We will add a new variable named `environment` using the newly created type `Environment` in all the variable managers created to execute operations.
+
+```
+package org.eclipse.sirius.web.core.api;
+
+public class Environment {
+
+  public static final String ENVIRONMENT = "environment";
+
+  private final String name;
+
+  public Environment(String name) {
+    this.name = Objects.requireNonNull(name);
+  }
+
+  public String getName() {
+    return this.name;
+  }
+}
+```
+
+It will be provided like other variables: `variableManager.put(Environment.ENVIRONMENT, new Environment("siriusComponents");`
+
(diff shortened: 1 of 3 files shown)


doc/adrs/039_provide_a_variable_to_detect_the_environment.adoc
```diff
@@ -0,0 +1,39 @@
+= ADR-039 - Provide a variable to detect the environment
+
+== Context
+
+We need to be able to distinguish the environment in which our representation descriptions are executed.
+This would help us create odesign which works both in the desktop environment and in a web environment.
+
+== Decision
+
+We will add a new variable named `environment` using the newly created type `Environment` in all the variable managers created to execute operations.
```
pcdavid, line 10: AFAIK, referencing a non-existant variable in an AQL expression creates an evalutation error.
Given the Sirius Desktop does not provide such a variable (yet?), using an expression which references it inside an `odesign` will fail in a Sirius Desktop env.

```
org.eclipse.sirius.common.tools.api.interpreter.EvaluationException: An error has appeared during the evaluation of an expression:
Expression: aql:if environment = 'desktop' then self.eClassifiers else Sequence{} endif
Diagnostic: Diagnostic ERROR source=org.eclipse.acceleo.query code=0 Couldn't find the 'environment' variable data=[null]
```

sbegaudeau: And what is the result of an `if` model operation if the results fails? Is it considered as `false`? If yes, then it works in our case 😄

sbegaudeau: In other words, what is the result of [that](https://github.com/pcdavid/org.eclipse.sirius/blob/aee60e43454d04ff568b9ebd4e12f47d7661f19e/plugins/org.eclipse.sirius/src/org/eclipse/sirius/business/internal/helper/task/operations/IfTask.java#L58) in case of an error. Wouldn't it work thanks to [that](https://github.com/pcdavid/org.eclipse.sirius/blob/38f51995cd1539a16cd7848033abd2428fd8be6b/plugins/org.eclipse.sirius.common/src/org/eclipse/sirius/common/tools/internal/interpreter/DefaultConverter.java#L36) since the result would be `null` and thus it would be converted to `false`

gcoutable (approved):

PR #1000, opened 2022-01-21
[985] Provide a variable to detect which environment is used

AxelRICHARD:
### Type of this PR 

### Issue(s)

https://github.com/eclipse-sirius/sirius-components/issues/985

### What does this PR do?

We need to be able to distinguish the environment in which our
representation descriptions are executed.
This would help us create odesign which works both in the desktop
environment and in a web environment.

### Potential side effects

Custom variable created in odesign with name "environment" could lead to collision name issue.

### How to test this PR?

- [x] Manual Test : please specify

Create a generic tool in an odesign, add an IF element under the Begin element with the following aql expression `aql:environment <> null`.
Then add a CreateInstance element that creates the semantic element of your choice.


backend/sirius-components-compatibility/src/main/java/org/eclipse/sirius/components/compatibility/services/diagrams/DiagramDescriptionNodeAndEdgeDescriptionsPopulator.java
```diff
@@ -162,7 +164,9 @@ private boolean isEnabledByDefault(Layer layer) {
                     childVariableManager.put(NEW_CONTAINER, newContainer.get());
                     childVariableManager.put(NEW_VIEW_CONTAINER, newViewContainer.get());
                     childVariableManager.put(ToolProvider.ELEMENT, self);
-
+                    if (optionalEnvironment.isPresent()) {
+                        childVariableManager.put(Environment.ENVIRONMENT, optionalEnvironment.get());
+                    }
```
sbegaudeau, line 169: The changes in this file are not necessary, a child variable manager can access the variable of its parent variable manager. We don't propagate `self` everywhere for example

sbegaudeau (changes requested):
This variable is missing at least from the EditLabelEventHandler. Could you also mention in the `CHANGELOG` that:

- The value of the variable will change to something specific for the integrating application (for example siriusWeb)
- This is only available for the diagram for now. We will see how we can make it a bit more automatic in order to support all the representations, maybe with a VariableManagerFactory implementation provided by the integrating application or something like that.

sbegaudeau (approved):
Just waiting for the build to be green to merge it

Issue #1022, opened 2022-01-26
Update our naming convention to distinguish events from their imagined behavior


sbegaudeau:
We have concepts such as:

- CreateNodeTool
- CreateEdgeTool
- NodeCreationEvent
- EdgeCreationEvent

None of those concept does exactly what it describe. A "CreateNodeTool" can create edges and no nodes at all, a `CreateEdgeTool` can rename a label and not create any edge, a CreateNodeTool could actually even be used to delete nodes, etc. We need to make those concepts more factual with regard with what they do. We could for example start by renaming `NodeCreationEvent` to `SinglePositionEvent` and `EdgeCreationEvent` to `DoublePositionEvent`.

I think that a first step necessary in order to start considering the impact of the existing NodeCreationEvent when it triggers the creation of edges for example.

sbegaudeau:
Closing this issue for now since the first part of this work has been integrated.

PR #1023, opened 2022-01-26
[959] Remove the concept of DeleteTool

sbegaudeau:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/959


backend/sirius-components-compatibility/src/main/java/org/eclipse/sirius/components/compatibility/services/diagrams/ToolProvider.java
```diff
@@ -362,7 +361,7 @@ private CreateEdgeTool convertEdgeCreationDescription(Map<UUID, NodeDescription>
         // @formatter:on
     }
 
-    private DeleteTool convertDeleteElementDescription(Map<UUID, NodeDescription> id2NodeDescriptions, AQLInterpreter interpreter, DeleteElementDescription deleteElementDescription) {
+    private CreateNodeTool convertDeleteElementDescription(Map<UUID, NodeDescription> id2NodeDescriptions, AQLInterpreter interpreter, DeleteElementDescription deleteElementDescription) {
```
pcdavid, line 364: We should probably rename `CreateNodeTool` into simply `NodeTool`? (and similarly for edges)

pcdavid (approved):

backend/sirius-components-compatibility/src/main/java/org/eclipse/sirius/components/compatibility/services/diagrams/ToolProvider.java
gcoutable: [#1022](https://github.com/eclipse-sirius/sirius-components/issues/1022) will handle that.

gcoutable (approved):

backend/sirius-components-compatibility/src/main/java/org/eclipse/sirius/components/compatibility/services/diagrams/ToolProvider.java
sbegaudeau: Yep and I think that it should not even be a `NodeTool`, it's just a tool in the palette activated by a single click as opposed to tools that are activated in two steps by opening the palette on the source and then selecting the target. Technically what we are calling today a `CreateNodeTool` can be applied on an edge.

I think I would be in favor of something radical like `SingleSelectionTool` (for the current `CreateNodeTool`) and `DoubleSelectionTool` (for the current `CreateEdgeTool`) which could then trigger a `SinglePositionEvent` and a `DoublePositionEvent` for the new rendering. This way, we would have something clearly distinct from some expected behavior.

@AxelRICHARD will bring support for the `DoubleClickTool` soon which has the advantage of doing exactly what it says

PR #1035, opened 2022-02-03
[779] Add support for tools' preconditions

AxelRICHARD:
### Type of this PR 

- [x] Documentation

### Issue(s)

https://github.com/eclipse-sirius/sirius-components/issues/779

### What does this PR do?

Add an ADR for tools' preconditions


doc/adrs/041_add_support_for_tools_preconditions.adoc
```diff
@@ -0,0 +1,87 @@
...
+public interface IToolSectionProvider {
+  boolean canHandle(..);
+  List<ToolSection> handle(Object targetElement,...);
+}
+```
+
+The code of both `ViewConverter.createToolSections(AQLInterpreter)` and `ToolProvider.getToolSections(Map<UUID, NodeDescription>, List<EdgeDescription>, DiagramDescription, List<Layer>)` would thus be transferred to two `IToolSectionProvider`
+
+The precondition of the tool will be evaluated by the `IToolSectionProvider` dedicated to compatibility layer (with the old code from `ToolProvider#getToolSections`)
+
```
pcdavid, line 81: If we start supporting tools' preconditions in the API, it should also be supported in Views:
* add a `preconditionExpression` to `org.eclipse.sirius.components.view.Tool`
* update `org.eclipse.sirius.components.emf.view.ViewConverter.createToolSections(AQLInterpreter)` to pass the precondition to the Node/Edge tools it creates.

pcdavid (changes requested):
The ADR only talks about tools which appear in the contextual palette (Node and Edge tools).
Should we also add precondition support to:

* direct edition;
* delete tools;
* drop tools.

Technically they are not currently "tools" (`ITool`), but plain Java callbacks/handlers.
Adding precondition support to them would mean either doubling the number of callbacks (`labelEditHandler` + `canEditLabelHandler`, etc.) or reifying them into proper `ITool`s.

I think supporting these would be nice in the future, but probably too much for this ADR.
They should still be mentioned though, at least as a known limitation that we considered and decided to defer for later.

doc/adrs/041_add_support_for_tools_preconditions.adoc
```diff
@@ -0,0 +1,87 @@
...
+== Decision
+
+=== Frontend
+
+Rename `getToolSectionsQuery` in operation.ts to `getDiagramDescriptionQuery`` (along with the renaming of the associated mutation and results in `DiagramWebSocketContainer``)
+
+Tools are part of the DiagramDescription, they will be computed on-the-fly from now on. Remove `DiagramDescription.getToolSections() and `GQLDiagramDescription#toolSections``
+
+Remove `toolSections` retrieving part from `getDiagramDescriptionQuery` and add it on `ContextualPalette`` openning.
```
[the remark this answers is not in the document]
sbegaudeau: I think that `getDiagramDescriptionQuery` can be removed entirely, we need the `getToolSectionQuery` to retrieve the tools because we need to retrieve them at some point 😄

sbegaudeau:
> The ADR only talks about tools which appear in the contextual palette (Node and Edge tools). Should we also add precondition support to:
> 
> * direct edition;
> * delete tools;
> * drop tools.

While it's not specified directly, one of the key goal of this work is to control the appearance of the direct edit / delete tools which are hardcoded on the frontend. The drop tool is not really a tool, the user drag some elements from the explorer to the diagram and we ask the server to do something about that. We don't know if the drag and drop of those elements is supported. The concept of DropTool only exists in the View DSL and in Sirius RCP, it does not exist in the core API nor in the GraphQL API so I don't think that it's relevant here.

> Technically they are not currently "tools" (`ITool`), but plain Java callbacks/handlers. Adding precondition support to them would mean either doubling the number of callbacks (`labelEditHandler` + `canEditLabelHandler`, etc.) or reifying them into proper `ITool`s.

The direct edit and delete tools will almost certainly  be represented as `ITool` sent by the server to the frontend during this work. We will probably ignore their handler to keep the existing behavior with their dedicated GraphQL API.

pcdavid:
> The drop tool is not really a tool, the user drag some elements from the explorer to the diagram and we ask the server to do something about that.

This looks like a tool to me (but I think our "misunderstanding" is on terminoloy only, not on the actual behavior):
* The user performs some UI interaction on the frontend;
* This result in some input event sent to the backend;
* The backend performs whichever operation is configured for this input in this particular modeler.

The "drop tool" does not have an explicit piece of UI to represent it, but that would be the same for a "double-click tool".

Maybe we should distinguish more clearly (in the API/terminology)  between:

* UI interactions which result in an event sent to the backend. The event includes all the information related to the UI interaction (single or double click? mouse position? key pressed? etc.) but does not assume anything about the backend's response.
* Actual behaviors on the backend which are compatible with some kinds of input events (based on the parameters they need to do heir work, i.e. which variables they need to be provided)
* Bindings between the two, possibly contextual/conditional.

The "tool" would be only the last part:

```java
public interface ITool {
    boolean handles(VariableManager inputVariables);
    Function<VariableManager, IStatus> getHandler();
}
```

1. the backend receives an input event, which corresponds to data gathered from a UI interaction on the front;
2. it builds a `VariableManager` with that data;
3. it asks the available tools if they can handle this input;
4. if a tool's `handles` returns `true`, we execute its handler.

sbegaudeau:
I agree that we need to distinguish things more clearly and I think we are not that far off.

We kind of already have the distinction in user interface interactions with the drop handler or the `del` key binding but most of our interactions rely on clickling on tools in the palette retrieved from the server. I think that all the possible user interactions on the frontend should be our tools and on the backend, we have our handlers and we are matching them thanks to the description and the event handlers.

I think that the tool should be the first part because it's almost the case already. We could have something like:

- tools are the user interface interaction capabilities provided to the frontend by the backend
- handlers are the behavior executed when we want to trigger a tool, they can do anything they want
- and between the two, we have the various `IDiagramEventHandler` to find the proper handler and resolve the variables (for example change an `objectId` into a proper `Object` or change a `diagramElementId` into a proper `Node` or `Edge`)

I don't think we will have a single generic `IDiagramEventHandler` in the future since the variables that need to be resolved are quite different for each tool. It could be possible to have something like a naming convention stating that if a variable from a tool is named `targetObjectId` then we should resolve it using `IObjectService#getObject(...)` etc but we are quite far from that for now.

I think that we could leverage some of our existing patterns to have something like:

```
public interface ITool {
  private String id;
  private String kind;
  private String selectionDescriptionId;
}

public interface IPaletteTool {
  private String imageURL;
}
public class SingleClickPaletteTool implements IPaletteTool {}
public class SingleClickOnTwoElementsPaletteTool implements IPaletteTool {
  private List<EdgeCandidate> edgeCandidates;
}
public class DoubleClickTool implements ITool {}
public class DropTool implements ITool {}
public class KeyBindingTool implements ITool {
  private String key;
  private List<String> modifiers; 
}
```

I think that we would still require a class hierarchy for our tools because the various user interactions may require different details (`imageURL` for the tools of the palette, the potential candidates for the "edge" tool, etc). We could express those using something else like a `kind` but in the end, it's the same result with more or less typing.

With the `kind` of the tool, we could have something like this:

- `siriusComponents://tool?interaction=SingleClick` for regular tools in the palette
- `siriusComponents://tool?interaction=SingleClickOnTwoElements` for our current edge creation tools
- `siriusComponents://tool?interaction=Drop` for drop tools
- `siriusComponents://tool?interaction=DirectEdit`
- `siriusComponents://tool?interaction=KeyBinding&key=Del`
- `siriusComponents://tool?interaction=Doubleclick` for future "double click" tools

I'm not really a fan since it would involve parsing stuff that can easily be expressed with specific attributes without issues. It's not like we really need to have a generic tool that could be connected to multiple user interface interaction. For that, one could create a handler and use it for multiple tools.

Our handlers will probably end up like that:

```
interface IToolHandler {
  boolean canHandle(Tool tool, VariableManager variableManager);
  IStatus handle(Tool tool, VariableManager variableManager);
}
```

Which have been created to match specific tools.
It would help make tools serializable (a first step in order to be able to serialize the whole diagram description one day) and would help extending diagram descriptions by being able to easily provide new tool handlers.

I think for this work, we could just improve things by having the `Delete from model`, `Delete from diagram` and `Direct Edit` being tools computed by the backend. The first two can simply be `NodeCreationTools` (renaming that will be done elsewehere) and the last one be a new kind of tool `DirectEditTool` because it triggers a different user interface interaction.

frontend/src/diagram/palette/ContextualPalette.types.ts
```diff
@@ -10,14 +10,43 @@
...
+  representationId: string;
+  diagramElement: any;
   invokeTool: (tool: Tool) => void;
   invokeConnectorTool: () => void;
   invokeLabelEdit: () => void | null;
   invokeDelete: (deletionPolicy: GQLDeletionPolicy) => void | null;
   invokeClose: () => void;
 }
+
+export interface GQLGetToolSectionsVariables {}
```
gcoutable, line 25: It should include variables needed by the query. In that case, the definition should be:
```
export interface GQLGetToolSectionsVariables {
  editingContextId: string;
  diagramId: string;
  diagramElementId: string;
}
```

frontend/src/diagram/palette/ContextualPalette.types.ts
```diff
@@ -10,14 +10,43 @@
...
  *******************************************************************************/
-import { GQLDeletionPolicy, GQLDiagramDescription, Tool, ToolSection } from 'diagram/DiagramWebSocketContainer.types';
+import { GQLDeletionPolicy, GQLToolSection, Tool } from 'diagram/DiagramWebSocketContainer.types';
 export interface ContextualPaletteProps {
-  diagramDescription: GQLDiagramDescription;
-  toolSections: ToolSection[];
-  targetElement: any;
+  editingContextId: string;
+  representationId: string;
+  diagramElement: any;
```
gcoutable, line 17: Should be a _SModelElement_

gcoutable (changes requested):

frontend/src/diagram/palette/ContextualPalette.types.ts
AxelRICHARD: Done

frontend/src/diagram/palette/ContextualPalette.types.ts
AxelRICHARD: Done

gcoutable (approved):

pcdavid (approved):
I have not really followed Axel's changes, but removing my "veto" not to block the merge.
My previous remarks were about the ADR and we exchanged on that in the comments.

AxelRICHARD:
The work is not done yet on this enhancement. More commits are coming.

backend/sirius-components-compatibility/src/main/java/org/eclipse/sirius/components/compatibility/services/ImageConstants.java
```diff
@@ -24,6 +24,12 @@
 
     public static final String DEFAULT_SVG = IMAGES_ROOT_FOLDER + "/Default.svg"; //$NON-NLS-1$
 
+    public static final String EDIT_SVG = IMAGES_ROOT_FOLDER + "/edit.svg"; //$NON-NLS-1$
+
+    public static final String SEMANTIC_DELETE_SVG = IMAGES_ROOT_FOLDER + "/semanticDelete.svg"; //$NON-NLS-1$
+
+    public static final String GRAPHICAL_DELETE_SVG = IMAGES_ROOT_FOLDER + "/graphicalDelete.svg"; //$NON-NLS-1$
+
```
sbegaudeau, line 32: Should probably be in a diagram-related project

backend/sirius-components-compatibility/src/test/java/org/eclipse/sirius/components/compatibility/diagrams/DomainBasedSourceNodesProviderTests.java
```diff
@@ -61,7 +62,18 @@ public void testComputeSourceNodes() {
 
         variableManager.put(DiagramDescription.CACHE, cache);
 
-        IIdentifierProvider identifierProvider = element -> nodeMapping.getName();
+        IIdentifierProvider identifierProvider = new IIdentifierProvider() {
```
sbegaudeau, line 65: Extend the `NoOp` one

backend/sirius-components-compatibility/src/test/java/org/eclipse/sirius/components/compatibility/diagrams/MappingConverterTests.java
```diff
@@ -103,7 +103,18 @@ public void testContainerStylePropertiesFromConditionalStyle() {
         containerMapping.getConditionnalStyles().add(this.createConditionalContainerStyle(EXPRESSION_TRUE, secondConditionalStyle));
         containerMapping.getConditionnalStyles().add(this.createConditionalContainerStyle(EXPRESSION_TRUE, thirdConditionalStyle));
 
-        IIdentifierProvider identifierProvider = element -> UUID.randomUUID().toString();
+        IIdentifierProvider identifierProvider = new IIdentifierProvider() {
```
sbegaudeau, line 106: Extend the `NoOp` one

backend/sirius-components-compatibility/src/test/java/org/eclipse/sirius/components/compatibility/diagrams/RelationBasedSourceNodesProviderTests.java
```diff
@@ -56,7 +57,18 @@ public void testComputeSourceNodes() {
 
         variableManager.put(DiagramDescription.CACHE, cache);
 
-        IIdentifierProvider identifierProvider = element -> nodeMapping.getName();
+        IIdentifierProvider identifierProvider = new IIdentifierProvider() {
```
sbegaudeau, line 60: Extend the `NoOp` one

backend/sirius-components-emf/src/test/java/org/eclipse/sirius/components/emf/compatibility/diagrams/EdgeMappingConverterTests.java
```diff
@@ -73,7 +74,18 @@ public void testEdgeFromNodeToContainer() {
 
         IObjectService objectService = new IObjectService.NoOp();
         IRepresentationMetadataSearchService representationMetadataSearchService = new IRepresentationMetadataSearchService.NoOp();
-        IIdentifierProvider identifierProvider = element -> containerMappingUUID.toString();
+        IIdentifierProvider identifierProvider = new IIdentifierProvider() {
```
sbegaudeau, line 77: Extend the `NoOp` one

backend/sirius-components-emf/src/test/java/org/eclipse/sirius/components/emf/compatibility/operations/OperationTestContext.java
```diff
@@ -62,7 +63,18 @@ public OperationTestContext() {
 
         this.objectService = new IObjectService.NoOp();
         this.representationMetadataSearchService = new IRepresentationMetadataSearchService.NoOp();
-        this.identifierProvider = element -> UUID.randomUUID().toString();
+        this.identifierProvider = new IIdentifierProvider() {
```
sbegaudeau, line 66: Extend the `NoOp` one

backend/sirius-components-emf/src/test/java/org/eclipse/sirius/components/emf/compatibility/properties/FormRendererTests.java
```diff
@@ -74,7 +74,18 @@ public AQLInterpreter create(ViewExtensionDescription viewExtensionDescription)
 
         IObjectService objectService = new IObjectService.NoOp();
         IRepresentationMetadataSearchService representationMetadataSearchService = new IRepresentationMetadataSearchService.NoOp();
-        IIdentifierProvider identifierProvider = element -> UUID.randomUUID().toString();
+        IIdentifierProvider identifierProvider = new IIdentifierProvider() {
```
sbegaudeau, line 77: Extend the `NoOp` one

doc/adrs/041_add_support_for_tools_preconditions.adoc
```diff
@@ -0,0 +1,92 @@
+= ADR-041 - Add support for tools preconditions
```
sbegaudeau, line 1: The ADR number should be updated now

sbegaudeau (changes requested):
I'll take over the PR to simplify things and fix the small issues detected here and there

sbegaudeau (approved):

PR #1038, opened 2022-02-04
[1022] Contribute an ADR on the renaming of some event and tools

sbegaudeau:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/1022

### DIFF
--- a/doc/adrs/041_rename_concepts_related_to_tools.adoc
+++ b/doc/adrs/041_rename_concepts_related_to_tools.adoc
@@ -1,0 +1,29 @@
+= ADR-041 - Rename concepts related to tools
+
+== Context
+
+The concepts related to our tools are very tightly related to the effect that we think they will have most of the time.
+As a result, we have concepts such as the `CreateNodeTool` and the `CreateEdgeTool` in the GraphQL API.
+We also have the internal concepts of `NodeCreationEvent` and the `EdgeCreationEvent` in the internal diagram API.
+
+This create an issue since, by design, we cannot predict the behavior of any of those tool or event.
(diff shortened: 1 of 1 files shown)


pcdavid (commented):
I understand the need for the position information, but it still seems strange to have this inside the name (instead of something like `SingleElementTool`).
If the "Delete (from diagram/model)" become actual tools (as mentioned in https://github.com/eclipse-sirius/sirius-components/pull/1035#issuecomment-1029757202), they do not necessarily have a position associated to them (e.g. when triggered from the keyboard).

With this naming convention how would we distinguish a "double-click tool" on a single element/position from a "single-click tool"?

sbegaudeau:
Yeah, the name is not great but I feel like `SingleElementTool` creates the expectation that a single element will be created. The `Delete from model` and `Delete from diagram` tools tools could also provide the position from which they have been triggered but that doesn't really work with a delete triggered by the `Del` key or the direct edit...

We may need something like that:

- `SingleClickOnElementTool` with a single position which triggers a `SinglePositionEvent` during the refresh
- `SingleClickOnTwoElementsTool` with a `DoublePositionEvent` for the refresh
- `DoubleClickOnElementTool` also with a single position and a `SinglePositionEvent`

😄

So the current `CreateEdgeTool` would become a `SingleClickOnTwoElementsTool`, those two elements can be the same one but we need two distinct "single click" to use it, not a "double click". The current `CreateNodeTool` would be a `SingleClickOnElementTool` and a future double click tool would be a `DoubleClickOnElementTool`.

sbegaudeau:
Closing this PR since part of this work has been done in https://github.com/eclipse-sirius/sirius-components/pull/1035

PR #1042, opened 2022-02-07
[releng] Update to sprotty 0.11.0

gcoutable:
### Type of this PR 

- [ ] Bug fix
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [x] Build/Releng

### What does this PR do?

It updates sprotty to 0.11.1 and removes deprecated code as much as possible. There is one deprecation remaining on our custom _SGraphFactory_ which does not have any replacement in _sprotty-protocol_.


frontend/__mocks__/styleMock.js
```diff
@@ -0,0 +1,13 @@
...
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+module.exports = {};
```
sbegaudeau, line 13: Could you add a small comment here indicating that it's to mock CSS imports required by Sprotty during the Jest tests as described here: https://jestjs.io/docs/webpack#handling-static-assets

frontend/package.json
```diff
@@ -58,7 +58,8 @@
     "rollup-plugin-postcss": "4.0.2",
     "rollup-plugin-terser": "7.0.2",
     "rollup-plugin-typescript2": "0.31.1",
-    "sprotty": "0.9.0",
+    "sprotty": "0.11.1",
+    "sprotty-protocol": "0.11.0",
```
sbegaudeau, line 62: Why the difference in `devDependencies` vs `peerDependencies`?

gcoutable: _sprotty-protocol_ is not needed in `peerDependencies` as it comes with Sprotty.
If I don't put _sprotty-protocol_ in `devDependencies`, _sprotty-protocol_ imports work but the auto-completion will not propose _sprotty-protocol_.

sbegaudeau: I have tested on my end while updating the version of sprotty used and it does work. Could you just remove `spotty-protocol` here, I understand the appeal of having the completion support but it won't prevent anyone from typing `import Stuff from 'sporty'` and it could create problems if we don't update it properly by having incompatible versions of sprotty and sprotty-protocol (we already have different version number here). If you prefer it the other way, just make sure that everything is aligned in all the package.json in all the locations.

I don't really mind that you do that one way or the other, I just want consistency in the package.json (for example, in the devDependencies of Sirius Components and the dependencies of Sirius Web). Could you move on with the submission of the other PRs so that I can merge all that and release Sirius Components 2022.01.10, thanks :)

sbegaudeau (changes requested):
Ok everything seems great, could you just update the CHANGELOG and then create another commit for the version bump? Thanks

sbegaudeau (approved):

PR #1044, opened 2022-02-08
[1045] Providers should return List<?> instead of List<Object> whenever possible

ldelaigue:
Miscellaneous improvements to make it easier for applications to use sirius-components.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -7,6 +7,7 @@
 - [ADR-039] Provide a variable to detect the environment
 - [ADR-040] Add support for a post selection
 - [ADR-041] Add the ability to contribute additional services to the EMFQueryService
+- [ADR-042] Use wildcard collections instead of List<Object> in Providers
 
 === Deprecation warning
 - [core] The various DTOs related to the creation and renaming of both documents and representations will be removed from the project at some point. The only reason why we will keep them for the moment is that some of them are used to trigger some specific behavior in the `EditingContextEventProcessor`. The split of the representation metadata should help us remove those special use cases
@@ -24,6 +25,7 @@
 - https://github.com/eclipse-sirius/sirius-components/issues/1019[#1019] [core] Services now have access to the context of the request triggering their execution. Consumer of Sirius Components will have to provide the `RequestAttributes`` in the `RequestContextHolder`` to make the `EditingContextEventProcessor`` work. It should be very easy for those in a Spring environment, but it is not automatic for those outside of a Spring environment
 - https://github.com/eclipse-sirius/sirius-components/issues/699[#699] [core] The `IRepresentationImageProvider` will now use the `kind` of the representation instead of the instance to compute the image. It will allow users to ask for the image of a representation with either its instance or its metadata
 - https://github.com/eclipse-sirius/sirius-components/issues/699[#699] [core] Add the `targetObjectId` to the `RepresentationMetadata`
+- https://github.com/eclipse-sirius/sirius-components/issues/1045[#1045] [core] Providers will now return List<?> instead of List<Object>. This makes it possible for applications to reuse existing services to implement providers without making useless copies of lists.
 
 === Dependency update
 
--- a/backend/sirius-components-trees/src/main/java/org/eclipse/sirius/components/trees/description/TreeDescription.java
+++ b/backend/sirius-components-trees/src/main/java/org/eclipse/sirius/components/trees/description/TreeDescription.java
@@ -51,9 +51,9 @@ public final class TreeDescription implements IRepresentationDescription {
 
     private Function<VariableManager, Boolean> deletableProvider;
 
-    private Function<VariableManager, List<Object>> elementsProvider;
+    private Function<VariableManager, List<?>> elementsProvider;
 
-    private Function<VariableManager, List<Object>> childrenProvider;
+    private Function<VariableManager, List<?>> childrenProvider;
 
     private Function<VariableManager, Boolean> hasChildrenProvider;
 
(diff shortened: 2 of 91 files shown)


backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/services/EMFValidationService.java
```diff
@@ -39,6 +40,7 @@
  * @author gcoutable
  */
 @Service
+@ConditionalOnMissingBean
```
sbegaudeau, line 43: Why do you need that to be conditional? If you are using EMF, why wouldn't you use the EMF validation service?

backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/services/ObjectService.java
```diff
@@ -47,6 +48,7 @@
  * @author hmarchadour
  */
 @Service
+@ConditionalOnMissingBean
```
sbegaudeau, line 51: Same question here, if you need the EMF part of Sirius Components, why not use its object service?

sbegaudeau (changes requested):

backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/services/EMFValidationService.java
ldelaigue: I'm not using EMF. The addition of these annotations make it possible for my application not to depend on EMF, that's the point :-)

backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/services/ObjectService.java
ldelaigue: I think I begin to see what you mean... I'm trying to not depend on EMF. How can I avoid having the services & other Spring components of sirius-components-emf scanned? Marking them as @ConditionalOnMissingBean works but may not be the best answer.
However, I still believe that even if I want to depend on EMF, it's be legitimate for an app to be able to inject their own implementation (which can be achieved by annotating it with @Primary)

backend/sirius-components-trees/src/main/java/org/eclipse/sirius/components/trees/description/TreeDescription.java
```diff
@@ -51,9 +51,9 @@
 
     private Function<VariableManager, Boolean> deletableProvider;
 
-    private Function<VariableManager, List<Object>> elementsProvider;
+    private Function<VariableManager, List<?>> elementsProvider;
```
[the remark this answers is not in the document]
ldelaigue: I've created bug #1045; I will update the changelog as well

doc/adrs/041_use_wildcard_collections.adoc
```diff
@@ -0,0 +1,55 @@
...
+** `SelectDescription`
+*** `optionsProvider`: Only used in `SelectComponent`, to transform into a list
+
+=== Compatibility layer
+
+N/A
+
+=== Frontend
+
+N/A
```
sbegaudeau, line 49: Those parts are not required in every ADR, you can delete them

backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/services/EMFValidationService.java
ldelaigue: I removed these useless @ConditionalOnMissingBean.

doc/adrs/041_use_wildcard_collections.adoc
ldelaigue: Okay

doc/adrs/041_use_wildcard_collections.adoc
```diff
@@ -0,0 +1,55 @@
...
+** `ListDescription`
+*** `itemsProvider`: Only used in `ListComponent`, to transform into a list
+** `MultiSelectDescription`
+*** `optionsProvider`: Only used in `MultiSelectComponent`, to transform into a list
+** `PageDescription`
+*** `semanticElementsProvider`: Only used in `PageComponent`, to transform into a list
+** `RadioDescription`
+*** `optionsProvider`: Only used in `RadioComponent`, to transform into a list
+** `SelectDescription`
+*** `optionsProvider`: Only used in `SelectComponent`, to transform into a list
```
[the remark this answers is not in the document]
ldelaigue: I'll update this with the exact list of the following commit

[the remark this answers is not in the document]
ldelaigue: I'll also take into account `SelectionDescription.objectsProvider` & `ValidationDescription.diagnosticsProvider`

CHANGELOG.adoc
```diff
@@ -6,6 +6,7 @@
 
 - [ADR-039] Provide a variable to detect the environment
 - [ADR-040] Add support for a post selection
+- [ADR-041] Use wildcard collections instead of List<Object> in Providers
```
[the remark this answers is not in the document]
ldelaigue: Done. I've updated this PR and pushed PRs to depending projects, including sirius-web.

doc/adrs/042_use_wildcard_collections.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+* Using `Iterable<?>`
+** Clarifies that what's returned is only intended to be iterated over, but does not make it possible to easily know the number of elements, or if there's an element at all.
+* Using `Collection<?>`
+** Better than `List<Object>`, does not impose ordering of elements, and makes it easy to check isEmpty() or size().
+* Using `List<?>`
+** Better than `List<Object>`, conveys that ordering of elements matters, and makes it easy to check isEmpty() or size().
+
+== Decision
+
```
pcdavid, line 20: You mention several alternatives above, but the actrual "Decision" section does not make it clear which one was chosen.
Yes it's more or less implied by the properties of the alternatives, but it does not cost much to make the choice more explicit.

pcdavid:
This changes APIs, so it will require a version bump (and the corresponding switch on downstream projects like Sirius Web).

sbegaudeau (approved):

PR #1058, opened 2022-02-10
[1052] Add the ability to contribute additional services in the EMFQueryService

RaphaelPageObeo:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/1052

### DIFF
--- a/backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/query/EMFQueryService.java
+++ b/backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/query/EMFQueryService.java
@@ -18,6 +18,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
+import java.util.stream.Collectors;
 
 import org.eclipse.emf.ecore.EPackage;
 import org.eclipse.sirius.components.collaborative.api.IQueryService;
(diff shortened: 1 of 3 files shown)


backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/query/EMFQueryService.java
```diff
@@ -51,8 +53,11 @@
 
     private final IEditingContextEPackageService editingContextEPackageService;
 
-    public EMFQueryService(IEditingContextEPackageService editingContextEPackageService) {
+    private final List<IQueryJavaServiceProvider> queryJavaServiceProviders;
+
+    public EMFQueryService(IEditingContextEPackageService editingContextEPackageService, List<IQueryJavaServiceProvider> queryJavaServiceProviders) {
```
pcdavid, line 58: Minor, but an alternate constructor could have reduced the impacts:

```java
public EMFQueryService(IEditingContextEPackageService editingContextEPackageService) {
    this(editingContextEPackageService, List.of());
}
```

pcdavid (approved):

backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/query/EMFQueryService.java
RaphaelPageObeo: Indeed :)

sbegaudeau: Why would we need to keep the old constructor?

pcdavid: To avoid impacting existing code which does not need to provide additional service classes.
Yes, it's only our own test code in this case, that's why I said it was a minor remark, not that it was a required change.

PR #1079, opened 2022-02-21
[1070] Reveal the currently selected representation in the explorer

pcdavid:
### Type of this PR 

- [ ] Bug fix
- [ ] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

https://github.com/eclipse-sirius/sirius-components/issues/1070

### What does this PR do?

- Switch to a flat tree format for the GraphQL message so that the frontend does not need to keep track of try to guess the depth of the tree (and remove the now obsolete maxDepth).
- Ask the backend to reveal the currently selected elements by expanding their ancestors.
- Update TreeRenderer to convert elements to reveal (TreeEventInput.revealed) into ancestors to expand using the new TreeDescription.getAncestorsProvider().


pcdavid:
Note that I still need to add at least one ADR (or maybe two) before this can be reviewed & merged.

pcdavid:
> Note that I still need to add at least one ADR (or maybe two) before this can be reviewed & merged.

The new patch includes the ADR and can be reviewed.

sbegaudeau:
There were two main approaches to solve this issue:

- One one side, we could have asked the server for the list of the elements to expand for a given selection along with the max depth of the element to reveal and then keep the existing lifecycle of the tree representation
- On the other side, this solution, in which the elements to reveal are computed as part of the refresh of the tree representation and thus given in the `TreeEventInput`

You went for the second approach because you had some concerns regarding the performance of the first approach.
After some testing that I have realized, I think that the second approach comes with way more performance regressions.


##### Lifecycle change

With this solution, each time the selection is modified, we can't possibly know what the new tree will look like. Since we don't know the new elements which will be expanded, we need to unsubscribe from the current tree to subscribe to a new tree representation. Even if the tree is the same, we have to recompute the whole tree from scratch after any click. In this use case, the list of elements to expand and the depth of the tree have not changed at all.

https://www.youtube.com/watch?v=dM9HdAUkswA

On top of that, since the selection of each user is now part of the `TreeEventInput`, it is a key part of the identity of the tree computed on the backend. As a result, multiple users whatching the same tree but with a different selection will now require a dedicated tree instance on the backend which will consume more memory and which will require more refresh operations after any semantic change in the editing context.

On master, with three users watching the same tree but with a different selection:

[](https://i.imgur.com/nUor4fw.png)

With this pull request, the three users will require three tree instances on the backend:

[](https://i.imgur.com/VjOH4Ll.png)


I don't see how we could possibly prevent those performence regressions by keeping the selection as part of the identity of the tree. As a result, I don't see how we could move forward with this solution while computing the elements to reveal during the rendering.

You can see here a comparison of the before and after of the impact on the lifecycle of the tree representation:

- Before: https://www.youtube.com/watch?v=Bnw7ZgJSpww
- After: https://www.youtube.com/watch?v=nfQZK1-QXPw


##### Datastructure change

Since you can't possibly know the new depth of the tree after the refresh, you have changed the data structure of the tree in the GraphQL layer. Doing so at the GraphQL layer, instead of the service layer and thus doing so outside of Sirius Components will require anyone integrating Sirius Components to do the same computation.

Event if we did integrate that part in Sirius Components, unless I missed something, it seems that each time a refresh of a tree will occur, we will need to navigate the whole tree data structure way more than before.

Currently on master, when a tree is refreshed, we need to:

- Navigate the data structure to be rendered (the containment relationships of the `EMFEditingContext`) in the `TreeRenderer` to build the tree
- Navigate the tree on the frontend to render it using `ExplorerWebSocketContainer`

Now, if I understand correctly, we need to:

- Navigate the data structure to be rendered (the containment relationships of the `EMFEditingContext`) in the `TreeRenderer` to build the tree
- Navigate once again the tree to create a flat tree on the backend
- Navigate that flat tree on the frontend to recreate the real tree
- Navigate the tree on the frontend to render it using `ExplorerWebSocketContainer`

That seems to require a lot of additional traversal of our tree data structure even in use cases where it's not necessary. For example, if some users are watching the same project with various selections and one of those user is editing something in the details view (nothing related to a change in selection), now everybody has to perform those additional traversal of our tree.

To conclude, I think that the impact of this contribution should not be underestimated. It comes with quite sizable changes to number of instances of the explorer we will keep in memory and the number of times those instances will be re-rendered. I don't think that we can afford to go this way.

sbegaudeau (changes requested):
See the comment above

sbegaudeau:
I've created a [draft of an ADR](https://github.com/eclipse-sirius/sirius-components/commit/8bab3449a5370240c865e2910b1926b502117ff5) on the subject to indicate what the other approach would look like. I may have missed a thing or two.

frontend/src/explorer/ExplorerWebSocketContainer.tsx
```diff
@@ -36,7 +42,36 @@ export const ExplorerWebSocketContainer = ({
...
+    GQLGetTreePathData,
+    GQLGetTreePathVariables
+  >(getTreePathQuery);
+
+  useEffect(() => {
+    if (tree) {
+      const variables: GQLGetTreePathVariables = {
+        editingContextId,
+        treeId: tree.id,
+        selectionEntryIds: synchronized ? selection.entries.map((entry) => entry.kind + '&id=' + entry.id) : [],
```
sbegaudeau, line 57: This should be an id, not a custom made data. We can't trust GraphQL calls to create the proper data format, the frontend should not be trusted to create coherent derived information. The backend should figured out the type of the data from its id. It gave it to the frontend after all.

sbegaudeau (commented):
I'll update the PR with the necessary fixes

sbegaudeau (approved):

PR #1098, opened 2022-03-11
[837] Fix dropped element position

RaphaelPageObeo:
### Type of this PR 

- [x] Bug fix
- [ ] Documentation
- [x] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

closes https://github.com/eclipse-sirius/sirius-components/issues/837

### What does this PR do?

This PR refactors the algorithm used to provide the position of the nodes during the incremental layout. The reasoning behind the refactoring will be explained with comments on the modified files.

It also changes this algorithm so that it can compute the position of a new parent node containing children. This allows to fix the bug described in [837](https://github.com/eclipse-sirius/sirius-components/issues/837).

It also cleans up existing tests.

Before : 

https://user-images.githubusercontent.com/61457677/157916872-82860644-97e6-4e25-b79b-142163baf802.mp4

After : 

https://user-images.githubusercontent.com/61457677/157916931-62c28345-4517-4487-80d6-3fbe6e171028.mp4

### Potential side effects

I am not sure if this would qualify as a side effect, but there is an existing problem that this PR does not fix.
Currently, a node is considered to have an undefined position if its position is equal to Position.UNDEFINED = Position.at(-1, -1).

So a node who is actually placed at (-1, -1) would be considered as not positioned, and the layout will certainly have undesirable effects on the diagram.

### How to test this PR?

- [ ] Frontend Unit tests
- [x] Backend Unit tests


backend/sirius-components-diagrams-layout/src/main/java/org/eclipse/sirius/components/diagrams/layout/incremental/IncrementalLayoutEngine.java
```diff
@@ -82,7 +82,6 @@ public IncrementalLayoutEngine(NodeSizeProvider nodeSizeProvider) {
     }
 
     public void layout(Optional<IDiagramEvent> optionalDiagramElementEvent, DiagramLayoutData diagram, ISiriusWebLayoutConfigurator layoutConfigurator) {
-        this.nodePositionProvider.reset();
```
RaphaelPageObeo, line 85: Previously, the nodePositionProvider was reset every time we required the layout of a diagram.

The effect of this reset was exclusively to set to null a field, named `last`. This field was used to remember the last node whose position was computed by the nodePositionProvider. This node can not be affected by the rest of the layout. 
The provider requires this information when several siblings need to be positioned so that they do not overlap. 

This reset mechanism caused two main issues : 
- The first one is probably personal, but I found it _odd_ that the nodePositionProvider was not 'self contained'. In this state, the provider actually depended on the layout algorithm to decide the position of the nodes. In my opinion, this goes against the logic that this component is supposed to be the one providing the position of the nodes.
- The second one is what caused the bug [837](https://github.com/eclipse-sirius/sirius-components/issues/837). It is easier to explain why with an example.
`
Take a parent node and two children. They are all being created at the same time, so none of those nodes have a position. 
The node provider pins the first child. Then it uses its field "last" to position the second child. The second child becomes "last".
Now we are positioning the parent. Since "last" is not reset, the parent is positioned after child two. 
`

This is why I remove this reset mechanism, and let the provider find the last positioned sibling by itslef.

backend/sirius-components-diagrams-layout/src/main/java/org/eclipse/sirius/components/diagrams/layout/incremental/provider/NodePositionProvider.java
```diff
@@ -51,116 +54,160 @@
...
-                if (this.isEventPositionInParentBounds(node.getParent(), Position.at(eventX, eventY))) {
-                    // The node has been created by a tool and has a fixed position
-                    double xPosition = eventX - parentPosition.getX();
-                    double yPosition = eventY - parentPosition.getY();
-                    position = Position.at(xPosition, yPosition);
-                    this.last = node;
-                } else {
-                    // The new node should appear in the parent, without moving the parent unless necessary
-                    position = Position.at(10, 10);
-                    this.last = node;
```
RaphaelPageObeo, line 73: The main interesting thing here is that I create a special case for the NodeType.NODE_LIST_ITEM nodes, instead of having them mingled with the rest. I believe it helps to understand what the algorithm does.

backend/sirius-components-diagrams-layout/src/test/java/org/eclipse/sirius/components/diagrams/layout/incremental/DiagramLayoutTest.java
```diff
@@ -157,6 +157,6 @@ public void testSimpleDiagramLayout() throws IOException {
         Optional<Node> optionalLayoutedThirdParent = this.getNode(layoutedDiagram.getNodes(), "Third Parent"); //$NON-NLS-1$
         assertThat(optionalLayoutedThirdParent).isPresent();
         Node layoutedThirdParent = optionalLayoutedThirdParent.get();
-        assertThat(layoutedThirdParent.getPosition()).isEqualTo(Position.at(301, 101));
+        assertThat(layoutedThirdParent.getPosition()).isEqualTo(Position.at(300, 100));
```
RaphaelPageObeo, line 160: This was actually a bug, there should not haven been a 1 pixel difference here.

backend/sirius-components-diagrams-layout/src/test/java/org/eclipse/sirius/components/diagrams/layout/incremental/NodeCreationTests.java
```diff
@@ -137,20 +137,89 @@ public void testNodeCreationOutsideItsParent() {
...
+    }
+
+    /**
+     * The is the exact same method than {@link NodePositionProvider} except that it is for {@link Node}.
+     *
+     * @param node
+     *            The node
+     * @return its default position
+     */
+    private Position getDefaultPosition(Node parent, Node child) {
```
RaphaelPageObeo, line 200: Copy pasting this algorithm just to change the type of the parameter is not very elegant, but I did not have a better idea.

backend/sirius-components-diagrams-layout/src/test/java/org/eclipse/sirius/components/diagrams/layout/incremental/NodePositionProviderTests.java
```diff
@@ -48,9 +48,9 @@
 
     private static final double START_Y_WITHIN_PARENT = 35;
 
-    private static final double START_X_OUTSIDE_PARENT = 1000;
```
RaphaelPageObeo, line 51: This position was not outside the parent, since the parent was 1000x1000.

backend/sirius-components-diagrams-layout/src/test/java/org/eclipse/sirius/components/diagrams/layout/incremental/NodePositionProviderTests.java
```diff
@@ -86,41 +85,32 @@ public void testDiagramNewNodesOneByOne() {
...
-        Position nextPosition = nodePositionProvider.getPosition(Optional.empty(), nodeLayoutData);
-        nodeLayoutData.setPosition(nextPosition);
-
-        assertThat(nextPosition).extracting(Position::getX).isEqualTo(Double.valueOf(0));
-        assertThat(nextPosition).extracting(Position::getY).isEqualTo(Double.valueOf(0));
-
-        nodePositionProvider = new NodePositionProvider();
-        nodeLayoutData = this.createNodeLayoutData(Position.UNDEFINED, DEFAULT_NODE_SIZE, diagramLayoutData, NodeType.NODE_RECTANGLE);
-        nodes.add(nodeLayoutData);
-        nextPosition = nodePositionProvider.getPosition(Optional.empty(), nodeLayoutData);
```
RaphaelPageObeo, line 98: I am not sure what this was testing.

backend/sirius-components-diagrams/src/main/java/org/eclipse/sirius/components/diagrams/components/DiagramComponent.java
```diff
@@ -89,12 +89,19 @@ public Element render() {
...
+                .orElse(Size.UNDEFINED);
+
         DiagramElementProps diagramElementProps = DiagramElementProps.newDiagramElementProps(diagramId)
                 .targetObjectId(targetObjectId)
                 .descriptionId(diagramDescription.getId())
                 .label(label)
-                .position(Position.UNDEFINED)
-                .size(Size.UNDEFINED)
+                .position(position)
+                .size(size)
```
RaphaelPageObeo, line 104: This is what was causing the "1 pixel difference bug" since the position of the diagram was being rest to (-1, -1) every time the layout algorithm was applied.

backend/sirius-components-diagrams-layout/src/main/java/org/eclipse/sirius/components/diagrams/layout/incremental/provider/NodePositionProvider.java
```diff
@@ -51,116 +54,160 @@
...
+    private Optional<Position> getPositionRelativeToSibling(NodeLayoutData node) {
+        Optional<NodeLayoutData> lastPositionedSiblingOptional = this.getLastPositionedSibling(node);
+        if (lastPositionedSiblingOptional.isPresent()) {
+            NodeLayoutData lastPositionedSibling = lastPositionedSiblingOptional.get();
+            Position lastPosition = lastPositionedSibling.getPosition();
+            Size lastSize = lastPositionedSibling.getSize();
+            double x = lastPosition.getX();
+            double y = lastPosition.getY() + lastSize.getHeight() + IncrementalLayoutEngine.NODES_GAP;
+            return Optional.of(Position.at(x, y));
         }
```
RaphaelPageObeo, line 180: This is what is replacing the "last" field. It will not be as performant as before but it works with the siblings. Does it seems ok with you?

sbegaudeau (approved):

PR #1101, opened 2022-03-14
[cleanup] TreeItemContextMenu cleanup

RaphaelPageObeo:
### Type of this PR 

- [ ] Bug fix
- [ ] Documentation
- [x] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)

...

### What does this PR do?

- Move TreeItem#deleteItem in TreeItemContextMenu
- Opens snackbar if an error occurs during the deletion
- Type deleteItem

### How to test this PR?

- [ ] Frontend Unit tests
- [ ] Backend Unit tests


frontend/src/tree/TreeItemContextMenu.tsx
```diff
@@ -10,16 +10,38 @@
...
-import { TreeItemContextMenuComponentProps } from 'index';
-import React, { useContext } from 'react';
+import gql from 'graphql-tag';
+import {
+  GQLDeleteTreeItemData,
+  GQLDeleteTreeItemInput,
+  GQLDeleteTreeItemVariables,
+  TreeItemContextMenuComponentProps,
+  TreeItemContextMenuState,
+} from 'index';
```
sbegaudeau, line 30: Do not import things from "Index", import them from the relevant file. Here `tree/TreeItemContextMenu.types`

frontend/src/tree/TreeItemContextMenu.tsx
```diff
@@ -33,9 +55,16 @@ export const TreeItemContextMenu = ({
   selection,
   setSelection,
   enterEditingMode,
-  deleteItem,
   onClose,
 }: TreeItemContextMenuProps) => {
+  const [state, setState] = useState<TreeItemContextMenuState>({ toast: 'hidden', message: null });
```
sbegaudeau, line 61: No need to keep the state of the toast, if the message is not null than display the toast.

frontend/src/tree/TreeItemContextMenu.tsx
```diff
@@ -44,60 +73,114 @@ export const TreeItemContextMenu = ({
...
+      id: uuid(),
+      editingContextId,
+      representationId: treeId,
+      treeItemId: item.id,
+    };
+    deleteTreeItem({ variables: { input } });
+    onClose;
+  };
+
+  const handleError = (loading, data, error) => {
```
sbegaudeau, line 91: The build is right, the whole content of the method should be inside of the useEffect, don't separate the code in another method.

frontend/src/tree/TreeItemContextMenu.tsx
```diff
@@ -44,60 +73,114 @@ export const TreeItemContextMenu = ({
...
+    if (!loading) {
+      if (error) {
+        openToast;
```
sbegaudeau, line 94: Perform the setState here, remove the method openToast

frontend/src/tree/TreeItemContextMenu.tsx
```diff
@@ -44,60 +74,114 @@ export const TreeItemContextMenu = ({
...
-      {item.deletable ? (
-        <MenuItem key="delete-tree-item" onClick={deleteItem} data-testid="delete" disabled={readOnly} aria-disabled>
-          <ListItemIcon>
-            <DeleteIcon fontSize="small" />
-          </ListItemIcon>
-          <ListItemText primary="Delete" />
-        </MenuItem>
-      ) : null}
-    </Menu>
+    <div>
```
sbegaudeau, line 114: Use `<> </>` instead of a div. You do not need an additional div which may change the layout

sbegaudeau (changes requested):

frontend/src/tree/TreeItemContextMenu.tsx
```diff
@@ -44,60 +68,114 @@ export const TreeItemContextMenu = ({
...
+
+  const deleteItem = () => {
+    const input: GQLDeleteTreeItemInput = {
+      id: uuid(),
+      editingContextId,
+      representationId: treeId,
+      treeItemId: item.id,
+    };
+    deleteTreeItem({ variables: { input } });
+    onClose;
```
sbegaudeau, line 82: `onClose` is a function, use it as such `onClose()` otherwise it does not work. I'm pretty sure that if you test that, the context menu is never closed while deleting something

frontend/src/tree/TreeItemContextMenu.tsx
```diff
@@ -44,60 +68,114 @@ export const TreeItemContextMenu = ({
...
+
+  useEffect(() => {
+    if (!deleteTreeItemLoading) {
+      if (deleteTreeItemError) {
+        setState({
+          message: 'An error has occurred while executing this action, please contact the server administrator',
+        });
+      }
+      if (deleteTreeItemData) {
+        const keys = Object.keys(deleteTreeItemData);
```
sbegaudeau, line 93: The code should look like this:

```
const { deleteTreeItem } = deletTreeItemData;
if (isErrorPayload(deleteTreeItem)) {
  const { message } = deleteTreeItem;
  setState({ message });
} 
```

You are not using TypeScript and you are not using the message from the error payload

frontend/src/tree/TreeItemContextMenu.types.ts
```diff
@@ -24,6 +24,31 @@ export interface TreeItemContextMenuProps {
...
+export interface GQLDeleteTreeItemData {
+  deleteTreeItem: GQLDeleteTreeItemPayload;
+}
+export interface GQLDeleteTreeItemPayload {
+  __typename: string;
+}
+
+export interface GQLDeleteTreeItemSuccessPayload extends GQLDeleteTreeItemPayload {
+  id: string;
+}
```
sbegaudeau, line 39: Unless I missed something, you should have a GQLErrorPayload extending GQLDeleteTreeItemPayload with a message and you should use it

sbegaudeau (changes requested):

frontend/src/tree/TreeItemContextMenu.tsx
```diff
@@ -44,60 +73,109 @@ export const TreeItemContextMenu = ({
...
+        anchorOrigin={{
+          vertical: 'bottom',
+          horizontal: 'right',
+        }}
+        open={state.message != null}
+        autoHideDuration={3000}
+        onClose={() => closeToast}
+        message={state.message}
+        action={
+          <IconButton size="small" aria-label="close" color="inherit" onClick={() => closeToast}>
```
sbegaudeau, line 173: `closeToast` is a function so the code should either be `onClick={closeToast}` or `{onClick=() => closeToast()}`. This does not work.

frontend/src/tree/TreeItemContextMenu.tsx
```diff
@@ -44,60 +73,109 @@ export const TreeItemContextMenu = ({
...
+        ) : null}
+      </Menu>
+      <Snackbar
+        anchorOrigin={{
+          vertical: 'bottom',
+          horizontal: 'right',
+        }}
+        open={state.message != null}
+        autoHideDuration={3000}
+        onClose={() => closeToast}
```
sbegaudeau, line 170: `closeToast` is a function, the code should either be `onClose={closeToast}` or `onClose={() => closeToast()}`. This does not work

sbegaudeau (changes requested):

PR #1111, opened 2022-03-16
[cleanup] Express TreeItem css with material ui

RaphaelPageObeo:
### Type of this PR 

- [x] Cleanup

### What does this PR do?

Express css of Tree Item with material ui


frontend/src/tree/TreeItem.tsx
```diff
@@ -35,39 +35,75 @@ const renameTreeItemMutation = gql`
   }
 `;
 
+const useTreeItemStyle = makeStyles((theme) => ({
+  treeItem: {
+    iconColor: '#66808a',
```
sbegaudeau, line 40: Is that properly considered as a variable?

frontend/src/tree/TreeItem.tsx
```diff
@@ -78,6 +114,38 @@ export const TreeItem = ({
...
+            height="20"
+            onClick={onClick}
+            data-testid={dataTestid}
+          />
+        );
+      }
+    }
+    return <div></div>;
+  };
+
```
sbegaudeau, line 148: No need to move this code, the composent gets re-created every time a refresh is performed. That has a cost.

sbegaudeau (changes requested):
I'll have a look at the comments and update the PR by myself

sbegaudeau (approved):

PR #1114, opened 2022-03-16
[1112] [1116] [1117] Improve support for multiple selection entries in the workbench

sbegaudeau:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/1112


gcoutable (changes requested):
I think some entries in the Changelog could be added and an ADR could be written at least for #1116

PR #1129, opened 2022-03-28
Select representation opened from a tab

pcdavid:
- [1128] Select the representation opened from a tab
- [1128] Remove dead code in the workbench's state machine

There might be more "aggressive" cleanups/refactoring possible, but given that we approach the release I prefered limiting the impacts.


sbegaudeau (approved):
I'll merge it just after the build

PR #1150, opened 2022-04-11
Add the ability to minimize the content of a site

pcdavid:


sbegaudeau:
Tell me if I missed something but this PR and [that one](https://github.com/eclipse-sirius/sirius-components/pull/1157) both share a commit named `Use vertically aligned buttons to select the view to display` but that's just because you pushed you current work on this PR and the other one is the real one that can be reviewed.

If I'm not mistaken then I'll mark this PR as a draft, feel free to change that if I'm not looking at the right one ^^'

pcdavid:
Yes, I had left this one open because it contains a commit with the WIP ADR about the actual minimization feature.

pcdavid:
See https://github.com/eclipse-sirius/sirius-components/pull/1191 instead.

PR #1158, opened 2022-04-13
Add support for a tree widget in forms

pcdavid:
See https://github.com/eclipse-sirius/sirius-components/issues/1139

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -9,6 +9,7 @@
 - [ADR-62] Switch to a monorepo layout
 - [ADR-63] Simplify the Sprotty integration
 - [ADR-64] Add support for a ToggableAreaContainer in forms
+- [ADR-65] Add support for a tree widget in forms
 
 === Deprecation warnings
 
@@ -16,6 +17,8 @@
 === Breaking changes
 
 === Dependency update
+
+- The frontend now depends on `@material-ui/lab` to support the new https://github.com/eclipse-sirius/sirius-components/issues/1139[tree widget] (see ADR-065).
 
 === Bug fixes
 
@@ -41,6 +44,7 @@
 - https://github.com/eclipse-sirius/sirius-components/issues/1248[#1248] [charts] Add support for pie-chart in Form representation
 - https://github.com/eclipse-sirius/sirius-components/issues/1255[#1255] [form] Add support for charts in form descriptions editor
 - https://github.com/eclipse-sirius/sirius-components/issues/1244[#1244] [form] Add support for flexbox containers on FormDescriptions
+- https://github.com/eclipse-sirius/sirius-components/issues/1139[#1139] [form] Add support for a new Tree widget
 - https://github.com/eclipse-sirius/sirius-components/issues/1140[#1140] [form] Widget groups now support a new setting (`displayMode = TOGGLEABLE_AREAS`) which allows end-users to select which of the widgets in the group are visible, and adapts their layout to the available horizontal space.
 
 == v2022.5.0
--- a/backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/components/TreeComponent.java
+++ b/backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/components/TreeComponent.java
@@ -1,0 +1,125 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.components.forms.components;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import org.eclipse.sirius.components.forms.TreeNode;
+import org.eclipse.sirius.components.forms.description.TreeDescription;
+import org.eclipse.sirius.components.forms.elements.TreeElementProps;
+import org.eclipse.sirius.components.forms.validation.DiagnosticComponent;
+import org.eclipse.sirius.components.forms.validation.DiagnosticComponentProps;
+import org.eclipse.sirius.components.representations.Element;
+import org.eclipse.sirius.components.representations.IComponent;
+import org.eclipse.sirius.components.representations.VariableManager;
+
+/**
+ * The component used to create the tree widget and its nodes.
+ *
+ * @author pcdavid
+ */
+public class TreeComponent implements IComponent {
+
+    public static final String ROOT_VARIABLE = "root"; //$NON-NLS-1$
+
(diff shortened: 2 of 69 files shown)


pcdavid:
The commit titled "[1139] Hard code a pseudo "semantic browser" in the default rules for testing" is just to check how the APIs work, and in particular to ensure they support the definition of the sections of a generic semantic browser (or something like it) without begin too convoluted.

Some version of the `SemanticBrowserDescriptionProvider` it defines will probably be reused for the tests.

frontend/src/properties/propertysections/TreePropertySection.tsx
```diff
@@ -0,0 +1,100 @@
...
+ *******************************************************************************/
+
+import { makeStyles } from '@material-ui/core/styles';
+import { ClassNameMap } from '@material-ui/core/styles/withStyles';
+import Typography from '@material-ui/core/Typography';
+import ChevronRightIcon from '@material-ui/icons/ChevronRight';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import { TreeItem } from '@material-ui/lab';
+import TreeView from '@material-ui/lab/TreeView';
+import { httpOrigin } from 'common/URL';
```
sbegaudeau, line 21: Do not use `httpOrigin` directly, rely on `useContext(ServerContext)`. Long story short, with this code your component would only work in an environment where the dev/prod use cases are the same as in Sirius Web. For example, this code cannot work in the VS Code extension.

frontend/src/properties/propertysections/TreePropertySection.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const updateSelection = (node: TreeNode) => {
+    if (node.selectable) {
+      const newSelection: SelectionEntry = {
+        id: node.id,
+        label: node.label,
+        kind: node.kind,
+      };
+      setSelection({ entries: [newSelection] });
+    }
+  };
```
sbegaudeau, line 71: This code could be moved where it needs to be used

frontend/src/properties/Group.tsx
```diff
@@ -109,3 +111,15 @@ export const Group = ({ editingContextId, formId, group, widgetSubscriptions, se
...
+    propertySection = (
+      <TreePropertySection
+        editingContextId={editingContextId}
+        formId={formId}
+        widget={widget}
+        key={widget.id}
+        subscribers={subscribers}
+        setSelection={setSelection}
+      />
+    );
```
sbegaudeau, line 125: There are a couple of rebasing issues, in this file at least, which prevent the code from compiling. I'll handle them

frontend/src/form/FormEventFragments.ts
```diff
@@ -265,6 +265,19 @@ export const formRefreshedEventPayloadFragment = gql`
...
+              expandedNodesIds
+              nodes {
+                id
+                parentId
+                label
+                kind
+                imageURL
+                selectable
+              }
+            }
```
sbegaudeau, line 280: This cannot work. I suspect that this was the result of a rebasing issue.

```
Validation error of type InvalidFragmentType: Fragment cannot be spread here as objects of type FlexboxContainer can never be of type TreeWidget @ 'flexboxContainerFields/children/children'
```

frontend/src/properties/propertysections/TreePropertySection.types.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+import { Subscriber, Tree } from 'form/Form.types';
+import { Selection } from 'workbench/Workbench.types';
+
+export interface TreePropertySectionProps {
+  editingContextId: string;
+  formId: string;
```
sbegaudeau, line 19: Those props are not used at all.

frontend/src/properties/propertysections/TreePropertySection.types.ts
```diff
@@ -0,0 +1,23 @@
...
+  widget: Tree;
+  subscribers: Subscriber[];
+  setSelection: (selection: Selection) => void;
```
sbegaudeau, line 22: It's a bit odd to ask for setSelection but not for the selection to highlight the current selection. It only works in one direction.

doc/adrs/065_add_tree_widget_in_forms.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+The backend API to describe a tree widget will follow the same pattern as the other `AbstractWidgetDescription`.
+It will have the following fields:
+
+- `Function<VariableManager, String> idProvider, labelProvider, iconURLProvider`: used to provide the unique id, label and optional icon URL of the actual `TreeWidget` being rendered.
+- `Function<VariableManager, List<?>> childrenProvider`: used to provide the ordered children of a particular candidate element.
+For a given render, this will be invoked *once* with both `self` and `candidate` variables set to the target element of the widget itself, and `ancestors` set to an empty list to determine the root elements to display.
+It will then be invoked recursively on every element returned with:
+  * `self` always set the target element of the widget itself;
+  * `candidate` set to each element in turn;
```
sbegaudeau, line 72: It's very odd to use the name `candidate` here since the current element for the various providers / expressions has always been `self`. The name `candidate` has been introduced for the candidates in the select / radio widgets since it represents a candidate for the value.

Something like `root` to identify the root target element.

backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/components/TreeComponent.java
```diff
@@ -0,0 +1,135 @@
...
+        List<TreeNode> nodes = new ArrayList<>();
+        for (var entry : childrenByParentId.entrySet()) {
+            String parentId = entry.getKey();
+            List<?> semanticChildren = entry.getValue();
+            for (Object semanticChild : semanticChildren) {
+                VariableManager itemVariableManager = variableManager.createChild();
+                itemVariableManager.put(CANDIDATE_VARIABLE, semanticChild);
+                nodes.add(this.renderNode(itemVariableManager, treeDescription, parentId));
+            }
+        }
```
sbegaudeau, line 73: Why the need to iterate twice on the tree?

CHANGELOG.adoc
```diff
@@ -17,6 +18,8 @@
 
 === Dependency update
 
+- The frontend now depends on `@material-ui/lab` to support the new https://github.com/eclipse-sirius/sirius-components/issues/1139[tree widget] (see ADR-061).
```
sbegaudeau, line 21: ADR 65

backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/components/TreeComponent.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+public class TreeComponent implements IComponent {
+
+    public static final String CANDIDATE_VARIABLE = "candidate"; //$NON-NLS-1$
+
+    public static final String ANCESTORS_VARIABLE = "ancestors"; //$NON-NLS-1$
+
+    public static final String NODES_VARIABLE = "nodes"; //$NON-NLS-1$
+
+    private TreeComponentProps props;
```
sbegaudeau, line 43: should be final

sbegaudeau (approved):

Issue #1152, opened 2022-04-12
Add support for form descriptions in the view converter


florianbarbin:
* [x] **I have checked that this feature has not yet been suggested by someone else.**

### Expected behavior
It will be possible to create new FormDescription in the DSL View from the frontend.
...

### Cuttings backs

...

### Potential side effects

...

### Describe alternatives you've considered

...

PR #1162, opened 2022-04-14
[doc] Add ADR for form descriptions in the view converter support

florianbarbin:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/1152
Signed-off-by: Florian Barbin <florian.barbin@obeo.fr>

### Type of this PR 

- [ ] Bug fix
- [x] Documentation
- [ ] Cleanup
- [ ] Test
- [ ] Build/Releng

### Issue(s)
#1152 
...

### What does this PR do?
Add the ADR for the support for form descriptions in the view converter
...

...
 

### Potential side effects

...

### How to test this PR?

- [ ] Frontend Unit tests
- [ ] Backend Unit tests

### Checklist

- [x] I have signed the [Contributor License Agreement](http://www.eclipse.org/legal/CLA.php).
- [x] All my commits are signed-off (`-s`) with my mail address of my Eclipse Account.
- [ ] All tests pass.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -6,7 +6,8 @@
 
 - [ADR-50] Add support for multiple selection entries in the details view
 - [ADR-51] Add support for form descriptions in the View DSL
-- [ADR-52] Add Support for EditingContext actions
+- [ADR-52] Add support for EditingContext actions
+- [ADR-53] Add support for form descriptions in the view converter
 
 === Deprecation warning
 
--- a/doc/adrs/053_add_support_for_form_descriptions_in_the_view_converter.adoc
+++ b/doc/adrs/053_add_support_for_form_descriptions_in_the_view_converter.adoc
@@ -1,0 +1,44 @@
+= ADR-053 - Add support for form descriptions in the view converter
+
+== Context
+
+The RepresentationDescriptions in the View DSL needs to be converted in the supported type `org.eclipse.sirius.components.representations.IRepresentationDescription`.
+
+For now, only `DiagramDescription` are converted by the `ViewConverter`.
+
+== Decision
+
+We need to create and add a `FormDescriptionConverter` to the `ViewConverter`.
+The `ViewConverter.convert()` method statically converts DiagramDescription:
+
+```            
+result = view.getDescriptions().stream()
+    .filter(org.eclipse.sirius.components.view.DiagramDescription.class::isInstance)
+    .map(org.eclipse.sirius.components.view.DiagramDescription.class::cast)
+    .map(viewDiagramDescription -> this.convertDiagramDescription(viewDiagramDescription, interpreter))
+    .collect(Collectors.toList());
+                         
+```
+
+We need to modify this logic to dispatch to the correct converter according to the `IRepresentationDescription` concrete type.
+We will create a common interface `IRepresentationDescriptionConverter` for `FormDescriptionConverter` and `DiagramDescriptionConverter`. The `ViewConverter` will no longer reference a specific converter implementation.
+
+The `FormDescriptionConverter` will create a new `org.eclipse.sirius.components.forms.description.FormDescription`. The DSL does not yet handle Pages and Groups. The converter will create a default page and group.
+
+As a first step, we will only convert `TextfieldDescription`. The Textfield Widget will be read only for now.
+
+Since the View DSL does not yet allow defining behavior on the Textfield Widget, the following fields from `TextfieldDescription` will be provided with default values:
+
+* `textfieldDescription.newValueHandler`: Will do nothing for now.
+* `textfieldDescription.diagnosticsProvider`: Will return an empty list.
+* `textfieldDescription.kindProvider`: Will return an empty String.
+* `textfieldDescription.messageProvider`: Will return an empty String.
+
+
+== Status
+
+WIP
+
+== Consequences
+
+We will be able to create a new Form Representation based on a defined FormDescription. This representation will just allow to display Textfield with a computed value but without editing capabilities in a first step.


CHANGELOG.adoc
```diff
@@ -31,6 +32,7 @@
 - https://github.com/eclipse-sirius/sirius-components/issues/1116[#1116] [form] Add support for multiple selection entries in the details view
 - https://github.com/eclipse-sirius/sirius-components/issues/1117[#1117] [diagram] Add support for multiple selections in a diagram. This work only display the various selection entries in the diagram. It does not support Ctrl+click or ⌘+click in a diagram. Support for this feature would require additional work with major improvements in the lifecycle of the `DiagramServer``
 - https://github.com/eclipse-sirius/sirius-components/issues/1146[#1146] [core] Add support for EditingContext actions. 
+- https://github.com/eclipse-sirius/sirius-components/issues/1152[#1152] [form] Add support for form descriptions in the view converter
```
sbegaudeau, line 35: Just use the tag [doc] to commit the ADR and don't mention the issue here. Because this commit is in no way adding support for form descriptions in the view converter.

doc/adrs/053_add_support_for_form_descriptions_in_the_view_converter.adoc
```diff
@@ -0,0 +1,42 @@
+= ADR-053 - Add support for form descriptions in the view converter
+
+== Context
+
+The RepresentationDescriptions in the Vew DSL needs to be converted in the supported type `org.eclipse.sirius.components.representations.IRepresentationDescription`.
```
sbegaudeau, line 5: `Vew` -> `View`

doc/adrs/053_add_support_for_form_descriptions_in_the_view_converter.adoc
```diff
@@ -0,0 +1,42 @@
...
+    .filter(org.eclipse.sirius.components.view.DiagramDescription.class::isInstance)
+    .map(org.eclipse.sirius.components.view.DiagramDescription.class::cast)
+    .map(viewDiagramDescription -> this.convertDiagramDescription(viewDiagramDescription, interpreter))
+    .collect(Collectors.toList());
+                         
+```
+
+We need to modify this logic to dispatch to the correct converter according to the `RepresentationDescription` concret type.
+
+The `FormDescriptionConverter` will create a new `org.eclipse.sirius.components.forms.description.FormDescription`. The DSL does not handle yet Pages and Groups. The converter will create a default page and group.
```
sbegaudeau, line 24: Please do not extend it in an hardcoded way, just make it take some `IRepresentationDescriptionConverter` and provide a `FormDescriptionConverter` along the `DiagramDescriptionConverter`. A key part to make it support new kinds of representation descriptions is to allow it to transform representation descriptions that it does not know.

sbegaudeau (changes requested):

doc/adrs/053_add_support_for_form_descriptions_in_the_view_converter.adoc
florianbarbin: OK, I will clarify this part.

CHANGELOG.adoc
florianbarbin: ok

doc/adrs/053_add_support_for_form_descriptions_in_the_view_converter.adoc
```diff
@@ -0,0 +1,43 @@
...
+result = view.getDescriptions().stream()
+    .filter(org.eclipse.sirius.components.view.DiagramDescription.class::isInstance)
+    .map(org.eclipse.sirius.components.view.DiagramDescription.class::cast)
+    .map(viewDiagramDescription -> this.convertDiagramDescription(viewDiagramDescription, interpreter))
+    .collect(Collectors.toList());
+                         
+```
+
+We need to modify this logic to dispatch to the correct converter according to the `IRepresentationDescription` concrete type.
+We will create a common interface `IRepresentationDescriptionConverter` for `FormDescriptionConverter` and `DiagramDescriptionConverter`. The `ViewConverter` will no longer reference a specific converter implementation.
```
pcdavid, line 23: Use a new line for each sentence (it helps when reading diffs).

doc/adrs/053_add_support_for_form_descriptions_in_the_view_converter.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+The `FormDescriptionConverter` will create a new `org.eclipse.sirius.components.forms.description.FormDescription`. The DSL does not yet handle Pages and Groups. The converter will create a default page and group.
+
+As a first step, we will only convert `TextfieldDescription`. 
+
+Since the View DSL does not yet allow defining behavior on the Textfield Widget, the following fields from `TextfieldDescription` will be provided with default values:
+
+-`textfieldDescription.newValueHandler`
```
pcdavid, line 31: `TextfieldDescription` does not seem to be tied to a specific attribute, but use an arbitrary `valueExpression`.
Can you clarify what would be the default behavior of the `newValueHandler`? I.e. how do you determine which attribute of which semantic element to `eSet` the new value?

pcdavid: OK, so I just read the last sentence below:

> without editing capabilities in a first step

So my remark does not apply to this version, but the fact that the current scope does not include edition could be mentioned earlier.

florianbarbin: Good remark, I will clarify the "default" behavior for each one of them.

doc/adrs/053_add_support_for_form_descriptions_in_the_view_converter.adoc
florianbarbin: ok

doc/adrs/053_add_support_for_form_descriptions_in_the_view_converter.adoc
sbegaudeau: > `TextfieldDescription` does not seem to be tied to a specific attribute, but use an arbitrary `valueExpression`.
> Can you clarify what would be the default behavior of the `newValueHandler`? I.e. how do you determine which attribute of which semantic element to `eSet` the new value?

That's the same behavior as in EEF / Sirius Desktop. I'm not sure that we want to change that to tightly coupled an attribute with a widget.

We could decide that we don't want to have the same behavior as in Sirius and couple widgets and attributes together (which would probably make simple use cases easier and make advanced use cases more complex) but that's a major change from the way things have been working so far. The label of a node or and edge in the view is not linked to an attribute either for example.

florianbarbin: From what I understand:
* The value of the textfield is computed according to the value expression provided by the specifier.
*  The behavior when changing this value will be specified in an another PR (by updating this ADR ?) but the specifier will also describe (with view::Operation probably), what is done with the new value entered by the end-user.

pcdavid: @sbegaudeau yes I understand, my remark was that on first reading "we will provide a default value for the newValueHandler" begged for the question: what kind of default behavior can you determine given that the text value is *not* bound to a specific attribute but computed? The ADR mentions that the first iteration will be read-only, but this is only mentioned later. I'm not asking for changing the behavior, just to clarify the text, for example just making it clear that the default behavior for newValueHandler does nothing.

sbegaudeau: I have not read the ADR yet, so I did not see that part in the ADR:  "we will provide a default value for the newValueHandler" :)

I want to be also clear, I do not know for a fact that we want the existing Sirius Desktop / EEF behavior, I need to clarify that first. We could very well end up with widgets tightly connected to features and a Form Description view DSL made with beginners in mind.

I'm pretty sure that we did not write anywhere what we want on that subject (on purpose to have this conversation, if I remember correctly). So in the end, I just know that we don't know... or at least that I don't know 😄

sbegaudeau (approved):

sbegaudeau:
I'll merge it after the build since I've rebased it on top of master

PR #1186, opened 2022-04-26
Draft ADR for user-supplied render variables

pcdavid:

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/adrs/adr-render-variables.adoc
+++ b/doc/adrs/adr-render-variables.adoc
@@ -1,0 +1,35 @@
+= ADR-XXX - Add support for additional variables during representation rendering
+
+== Context
+
+When rendering a representation, Sirius Components will invoke many different user-supplied functions.
+The execution context of these functions is controlled by the corresponding `IComponent` through the `VariableManager` passed to the function.
+The only variables available to the user-supplied functions are the ones defined by Sirius itself.
+
+As a consequence, every invocation of such a function starts "from scratch" and needs to compute everything it needs from only the Sirius-supplied variables each time.
+During a given rendering of a representation, the state of the underlying semantic model, the previous state of the representation (if there is one) and the representation description itself can be assumed to be fixed.
+
+=> opportunity to pre-compute some global information once (per render) to speed up the repeated invocations of the functions.
+
+== Decision
+
+We will add:
+
+```java
+interface IRepresentationDescription
+  default Function<VariableManager, VariableManager> getRenderVariablesProvider() {
+    return (variableManager) -> new VariableManager();
+  }
+}
+```
+
+The renderVariablesProvider will be guaranteed to be invoked exactly once per representation rendering, before any other user-supplied function.
+The variables it returns will be added to the evaluation context of all successive user-supplied function invocation (during the same refresh).
+If such a variable uses a name which conflicts with a system-provided variable, it will be ignored and a error will be logged so that the developer can change the name of his variable.
+
+== Status
+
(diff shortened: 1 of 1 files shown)


doc/adrs/adr-render-variables.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+We will add:
+
+```java
+interface IRepresentationDescription
+  default Function<VariableManager, VariableManager> getRenderVariablesProvider() {
+    return (variableManager) -> new VariableManager();
+  }
+}
+```
```
sbegaudeau, line 24: Just a couple things to keep in mind when we will move forward with this ADR and its implementation. We don't really need to custom every representation description like that, we can simply add an independent service like that:

```
public interface IRenderingVariableManagerProvider {
  VariableManager get(VariableManager variableManager, ...);
}
```

Which can be injected using Spring at the desired location. Since we try not to add more stuff on the description if it's not required and since it's not strictly necessary in projects like `sirius-components-diagrams` etc.

We will also have to think about the various variables available for the provider but we may have to change the lifecycle of some existing variables like:

- previous representation
- view creation request
- view deletion request
- representation context
- editing context

To start working really on this ADR, we should probably also document the variables which should be available in the rendering VariableManager of the various representations and answer some questions like "should `viewCreationRequests` be a `props` or a variable of `DiagramComponent`" and why :)

sbegaudeau (changes requested):
I fully agree with the philosophy of this ADR but as stated in my comment, it needs additional discussions to ensure that it matches properly some of our use cases.

pcdavid:
The proposed `IRenderingVariableManagerProvider` would need some kind of `boolean handles(IRepresentationDescription)` method.

On the one hand, I proposed putting `getRenderVariablesProvider` directly on `IRepresentationDescription` (or maybe just some subtypes at first) to avoid the need for this kind of `handles` method, and also because the additional variables this would provide are tightly coupled with the rest of the representation description: it only makes sense to provide them if the other user-supplied functions use them, and for a user-supplied function to exploit it must be sure they will be available. Putting it all together would help ensure this consistency.

On the other hand, having an externally supplied `IRenderingVariableManagerProvider` (or several?) opens an interesting possibility: use such custom rendering variables in modelers defined using `odesign` files (just provide the required `IRenderingVariableManagerProvider` as a bean along with the `odesign`).

If we wanted to be able to use this mechanism for studios, we could have something like this:

In `view.ecore`:

```
class RepresentationDescription {
  owned ref renderVariables: VariableDeclaration[]
}

class VariableDeclaration {
  attr name: EString;
  attr valueExpression: EString;
}
```

Then the platform itself could provide a single `IRenderingVariableManagerProvider` which, for representation descriptions defined in a studio, would evaluate said expressions to populate the render variables.
There is currently now way to know if a specific instance of e.g. `DiagramDescription` comes from a studio definition but https://github.com/eclipse-sirius/sirius-components/issues/1088 is supposed to allow that I think.

pcdavid:
See https://github.com/eclipse-sirius/sirius-web/blob/master/doc/adrs/200_add_configurable_render_variables_to_tree.adoc

PR #1188, opened 2022-04-26
[1176] Fix Exception in ViewDiagramDescriptionConverter

florianbarbin:
ViewDiagramDescriptionConverter is now a service but two attributes
convertedNodes and convertedEdges were mutable and used during the
conversion.

# Pull request template

## General purpose
- [x] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

This issue has been detected from this attached 
[Company DSL.zip](https://github.com/eclipse-sirius/sirius-components/files/8563334/Company.DSL.zip).
* Import the DSL
* Create a new project
* Create a new "Company" model
* Create a new 'Employee" diag
* Create several employees diagrams

- [ ] Has the Kiwi TCMS test suite been updated with tests for this contribution
?

### DIFF
--- a/backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/view/diagram/ViewDiagramDescriptionConverter.java
+++ b/backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/view/diagram/ViewDiagramDescriptionConverter.java
@@ -13,7 +13,6 @@
 package org.eclipse.sirius.components.emf.view.diagram;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -105,10 +104,6 @@ public class ViewDiagramDescriptionConverter implements IRepresentationDescripti
         return UUID.nameUUIDFromBytes(EcoreUtil.getURI(diagramElementDescription).toString().getBytes());
     };
 
-    private Map<org.eclipse.sirius.components.view.NodeDescription, NodeDescription> convertedNodes;
-
-    private Map<org.eclipse.sirius.components.view.EdgeDescription, EdgeDescription> convertedEdges;
-
     public ViewDiagramDescriptionConverter(IObjectService objectService, IEditService editService) {
         this.objectService = Objects.requireNonNull(objectService);
         this.editService = Objects.requireNonNull(editService);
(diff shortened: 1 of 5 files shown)


backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/view/diagram/ViewDiagramDescriptionConverter.java
```diff
@@ -572,4 +560,19 @@ private IDiagramContext getDiagramContext(VariableManager variableManager) {
         return variableManager.get(IDiagramContext.DIAGRAM_CONTEXT, IDiagramContext.class).orElse(null);
     }
 
+    /**
+     * A structure representing the current state of the conversion.
+     *
+     * @author fbarbin
+     */
+    private class ConvertedState {
```
pcdavid, line 568: I personally do not have any issue with this kind of private helper classes, but:

* if should be `static` unless it needs a hidden reference to the enclosing instance;
* it should still follow our coding practices (`final` fields, getters);
* in this case, instead of being a "dumb" wrapper for two collections, maybe it should expose a more meaningful and slightly high-level API which matches how the collections are used in the converter. See for example `org.eclipse.sirius.components.diagrams.renderer.DiagramRenderingCache` which handles a similar problem (keep track of the mutable state during a multi-step process).

If the class becomes more than a dumb wrapper, it probably deserves to be its own top-level type.

pcdavid (changes requested):

backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/view/diagram/ViewDiagramDescriptionConverter.java
sbegaudeau: Personnaly, I've never really seen a good reason why we NEED to be absolutely sure that nobody won't ever be able to use a class.

For me it's a mix of trying to predict the future and trying to hide stuff and I don't see why. A proper class which does one thing and well and which could be reused somewhere one day is always better in my mind.

Apart from very specific cases like creating an API which should be manipulated in a very specific way like our builder patterns used to prevent ˋnull` on specific fields or the spring security http configuration pattern (where you can write ˋhttp.crsf().disable().and()...`), I don't see the point.

florianbarbin: I extracted this class with a package visibility.

sbegaudeau: And why do you need this package visibility? Why would you NEED to prevent anyone from seeing this?

I still don't understand what's your reason to try to hide some code from others (and from who?)

pcdavid: > For me it's a mix of trying to predict the future and trying to hide stuff and I don't see why.
> ...
> could be reused somewhere one day

So, you're trying to predict the future too, just with a different default assumption :)

One argument for hiding things unless explicitly needed is to reduce our API surface and minimize future breakage. As soon as we make something public, it becomes API and any change to it will need to be documented and possibly break some users who started depending on it.

sbegaudeau: Since we don't use OSGi, everything is public by default. We can't hide from that. Any class even at the very bottom of our code stack which has no reason to be manipulated by anyone else than us has to be public because we need it in another part of our code.

Should we make `EditingContextEventProcessor` a private class? If so should we move it's code inside of `EditingContextEventProcessorFactory`? Or should we switch it to a package visibility? Should we have all of our code in a single package to hide everything from the outer world?

This is a plain old Java project, everything is public by default. That's the default assumption in Java.

pcdavid (approved):

backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/view/diagram/ViewDiagramDescriptionConverterContext.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public class ViewDiagramDescriptionConverterContext {
+
+    private final Map<org.eclipse.sirius.components.view.NodeDescription, NodeDescription> convertedNodes;
+
+    private final Map<org.eclipse.sirius.components.view.EdgeDescription, EdgeDescription> convertedEdges;
+
+    private final AQLInterpreter interpreter;
+
+    ViewDiagramDescriptionConverterContext(AQLInterpreter interpreter) {
```
sbegaudeau, line 36: Why does the constructor has a package visibility? I'll fix it by myself so we can merge this fix this week.

sbegaudeau (changes requested):

backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/view/diagram/ViewDiagramDescriptionConverterContext.java
florianbarbin: my mistake, sorry

PR #1340, opened 2022-09-02
[fix] Add missing dependency declaration

pcdavid:
`sirius-components-formdescriptioneditors` depends on types defined in `sirius-components-forms` but did not declare the dependency.


sbegaudeau (changes requested):
This should be added to the `peerDependencies` too and a `npm i` should be performed after that to update the `package-lock.json`. I'll handle it.

sbegaudeau (approved):

PR #1370, opened 2022-09-26
[doc] Add an ADR about hidden elements

theogiraudet:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,6 +12,7 @@
 - [ADR-080] Add support for styles preview in FormDescriptionEditor
 - [ADR-081] Add support for multi-groups in Form/FormDescriptionEditor
 - [ADR-082]Add support for optional text completion on textfield widgets
+- [ADR-083] Add the possibility to hide or fade a diagram element
 
 === Breaking changes
 
--- a/doc/adrs/083_add_the_possibility_to_hide_or_fade_a_diagram_element.adoc
+++ b/doc/adrs/083_add_the_possibility_to_hide_or_fade_a_diagram_element.adoc
@@ -1,0 +1,58 @@
+= ADR-083 - Add the possibility to hide or fade a diagram element
+
+== Context 
+
+Sirius Web displays all elements of a diagram but in large diagram, the user may want to focus only on certain elements that are more relevant to its use.
+
+== Decision
+
+We will add to diagram elements a new property `state: ViewModifier`.
+The state can either be `Normal`, `Faded` or `Hidden`.
+The value `Faded` is used to keep the element visible but more attenuated (transparency or shades of gray). 
+The value `Hidden` is used to totally hide the diagram element and make the layout algorithm ignore it. 
+
+For the propagation of the visibility state, if a diagram element is hidden, all contained nodes and connected edges have to be hidden too.
+If a diagram element is faded, we will not implement the propagation of this property to contained or connected diagram elements.
+This may come in a future iteration.
+
+From the user's point of view, they can hide or fade a diagram element through the UI. 
+If they want to reveal this element, the user has to reveal all hidden/faded elements.
+
+=== Backend
+
+==== Auto-layout
+
+In Sirius Web, the auto-layout is managed by ELK. 
+Sirius gives to ELK the graph and get from ELK a layouted graph. 
+This is this layouted graph that is saved in database. 
+When an element is hidden, we need to keep it in memory to allow the user to display it again. 
+
+For that, we need to make these hidden elements persist after the auto-layout was applied. 
+We can use the https://www.eclipse.org/elk/reference/options/org-eclipse-elk-noLayout.html[`noLayout`] on graphical elements (node, border nodes, edges) to tell ELK to ignore these elements when calculating the layout but to keep them in the diagram.
+
+Concerning faded graphic elements, they are still visible (from the user point of view) so the auto-layout will be applied.
+
+==== Data Structure
+
+The diagram data structure needs to be evolved to represent these two new states. 
+
+The reason a node should not be displayed may be that the user has chosen to hide or fade it, or it may be caused by the state propagation of neighboring nodes.
+We need for the frontend to know which modifier(s) the user applied to the node but we also need the state of the node according to the user modifiers as well as the propagation.
+
+Diagram elements, `Node` and `Edge` will have two new attributes.
+The first one, `Set<ViewModifier> modifiers`, will store all modifiers applied by the user.
+The second one, `ViewModifier state`, will store the computed state of the element according to user modifiers and state propagation.
+
+
+=== Frontend
+
+The palette will show a drop-down button allowing to hide or to fade the selected element.
+The toolbar will propose two buttons: one to reveal all hidden elements and the other to reveal all faded elements.
+The code of the palette and of the toolbar must be modified to add these new buttons.
+
+If an element has for state `Hidden` then it is not sent to Sprotty.
+This means that it will not appear in the diagram displayed to the user.
+
+== Status
+
+Accepted


doc/adrs/070_add_the_possibility_to_hide_or_fade_a_graphic_element.adoc
```diff
@@ -0,0 +1,67 @@
+= ADR-070 - Add the possibility to hide or fade a graphic element
```
sbegaudeau: Using ASCIIDoc, you are supposed to use one line per sentence. It would make it easier to comment such PRs.

doc/adrs/070_add_the_possibility_to_hide_or_fade_a_graphic_element.adoc
```diff
@@ -0,0 +1,67 @@
...
...sibility` that is an enumeration defining `NORMAL`, `FADED` and `HIDDEN`. In this last case, some graphic information like the `position` and `size` doesn't have meaning. Indeed, if the auto-layout is enabled, ELK decides the value of these parameters but it ignores hidden elements (because of the `noLayout`). This can create runtime bugs in the existing code: if a function iterates over graphic elements and uses the position/size for computation, the result can be wrong due to hidden elements. To resolve that, we can make these graphic information attributes optional. The `Optional` is empty if the state is `HIDDEN` and wraps the information otherwise. All code sections requesting these values will throw a compilation error. This makes it easier to identify the code to be adapted and thus avoids hidden execution bugs as described. An other solution is to wrap all graphic information in a common class `GraphicNodeInformation` and `GraphicEdgeInformation` and change existing attributes to a single attribute of these type that is optional. This avoids testing the visibility (or to call `Optional#get`) multiple time if we need to access to several graphic information at the same time.
```
sbegaudeau: - We could probably go with something like `ViewVisibility` instead of `GraphicalElementVisibility` since IIRC all the elements that can appear in a diagram are considered "views" (which explain the name of some variables like `sourceView`, `targetView`, `containerView` etc)
- Switching the types of `Node#size: Size`, `Node#position: Position` and `Edge#routingPoints: List<Position>` to `Node#size: Optional<Size>`, `Node#position: Optional<Position>` and `Edge#routingPoints: Optional<List<Position>>` would actually help save other issues. Right now, when we don't have any position for a node, it is initialized with `[-1, -1]` but the problem is that those coordinates are completely valid in a diagram. Using optional would make the lack of size/position/routingPoints explicit. It would move some existing bugs from being runtime bugs to compile time bugs which is good.

doc/adrs/070_add_the_possibility_to_hide_or_fade_a_graphic_element.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+==== Impacted code
```
sbegaudeau: This is fine but you don't have to try to create an explicit list of all the pieces of code impacted. You could have said, the ELK layout, the incremental layout and this and that.

CHANGELOG.adoc
```diff
@@ -5,6 +5,7 @@
 === Architectural decision records
 
 - [ADR-069] Add support for children layout strategy
+- [ADR-070] Add the possibility to hide or fade a graphic element
```
sbegaudeau: It could be better to by a tiny bit more precise since we have countless of "stuff" which can be considered as "graphical elements" with a title like "Add support to hide or fade diagram's views"

@pcdavid In Sirius Desktop all the graphical stuff in a diagram is called "view" right? like nodes, edges, decorators etc?

doc/adrs/070_add_the_possibility_to_hide_or_fade_a_graphic_element.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+=== Backend
+
+==== Auto-layout
+
+In Sirius Web, the auto-layout is managed by ELK. Sirius gives to ELK the graph and get from ELK a layouted graph. This is this layouted graph that is saved in database. When an element is hidden, we need to keep it in memory to allow the user to display it again. 
+
+The first solution is to totally remove a hidden element from the saved graph and to store it separately. But the element is just hidden, not removed and if a part of the application iterates over all elements of the graph, this part can not access to hidden elements. 
+
+The second solution is to persist these hidden elements after the auto-layout was applied. A way to do that is to indicate to ELK to not consider the element in the layout computation. For this, ELK proposes the https://www.eclipse.org/elk/reference/options/org-eclipse-elk-noLayout.html[`noLayout`] property applies to nodes, border nodes, edges and labels. With this solution hidden elements are still present in the diagram, to invisibilize them we can use CSS.
```
gcoutable: Just in case: According to the documentation, the `noLayout` option does not say that such elements will be hidden. They will still be visible in the frontend but with unknown size and position.

sbegaudeau: So we may need to ignore on the frontend those elements and the edges related to them

theogiraudet: Maybe my text isn't very explicit: I propose to use CSS properties to hide elements such as edges or nodes. Indeed, the `noLayout` is just to avoid the "presence" of these hidden elements on the layout of the diagram.
An other solution is to not send these graphic elements to the front (better for performance).

CHANGELOG.adoc
pcdavid: We use the term, and variables which expose these elements are indeed called `view*`, but strictly speaking there is no type called "View" (there is `DView` but it's the top-level element of an aird).
In Sirius Desktop we have:
* `DDiagramElement extends DRepresentationElement extends DSemanticDecorator` as the most general common type for all elements on a diagram.
* Then `DNode`, `DNodeContainer` and `DEdge` are subtypes of `DDiagramElement`

`DDiagramElement` is the one which has a `boolean isVisible()` attribute, but this one is "derived" (not in the common EMF sense, but via a pre-commit listener, `VisibilityUpdater`) from another attribute: `EList<GraphicalFilter> getGraphicalFilters()`.

Each diagram element can have one or more `GraphicalFilter` attached, which are used to compute the element's visibility (as in `isVisible`), and importantly, "adjacent" elements visiblity too. For example an edge is visible only if both its source *and* target are visible (see https://github.com/pcdavid/org.eclipse.sirius/blob/master/plugins/org.eclipse.sirius.diagram/src/org/eclipse/sirius/diagram/business/internal/helper/display/DisplayServiceImpl.java#L221); a label can be hidden either because it was hidden itself explicitly or because its parent node is invisible, etc.

`GraphicalFilter`s act like "modifiers". The main ones are:
* `HideFilter`: "basic" hide/reveal feature as expected in this ADR;
* `FoldingFilter`: can be applied to some edges (think an edge from a parent node to its children) to "fold" the edge and hide the whole sub-tree of elements "below" the edge;
* `CollapseFilter`: can be applied to a node or container to reduce its size and hide all its content.

This may seem over-complicated, but we need to store and remember each of these flags to restore the proper state when an element is revealed.

Basically an element can become "invisible" on the diagram for different, cumulative reasons, and removing *one* of these reasons is not enough to make it visible again, so a `boolean` is not enough.

doc/adrs/070_add_the_possibility_to_hide_or_fade_a_graphic_element.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+Concerning faded graphic elements, they is still visible (from the user point of view) so the auto-layout will be applied.
```
pcdavid: "they is" => "they are"

pcdavid:
As hinted in my comment about how it currently works in Sirius Desktop, I think the document is missing some explicit rules about how setting an element as FADED or HIDDEN can propagate (or not) to other adjacent elements (children, border nodes, labels, source/target elements for edges). In other words, the actual visibility and "display mode" (opaque or semi-transparent) of a given element can depend on the state of other related elements even if the element has no such attributes. Rules like:
* If either the source of target nodes of an edge are HIDDEN, the edge is considered implicitly HIDDEN;
* If the source of an edge is FADED, does it make the edge FADED? if the source *and* target are FADED, does this change anything? (it could be strange visually to have a fully visible edge between two semi-transparent elements, but that is debatable).
* If a container is HIDDEN, obviously all its children (recursively, incl. border nodes) should become invisible even if they have not been HIDDEN themselves explictly. Is it the same for FADED?
* etc.

pcdavid:
Also, what kind of UI is planned to reveal the hidden/faded elements?
The document only mentions:

> the representation will propose a way to open a dialog displaying hidden elements with the possibility to make them visible again.

* But what is displayed exactly in this dialog? A flat list of the labels of the hidden/faded elements? If so, how are represented elements without labels? How to distinguish elements which have the same label?
* Is it the same UI/dialog to reveal/"unfade" elements?
* Can it be used to reveal/"unfade" multiple elements at the same time?
* Is it a modal dialog, which when open would hide most of the diagram?
* It could be a side-panel, which can be opened while keeping the diagram visible, so that the effect is immediatly visible.
* When revealing/unfading an element, is it selected afterwards?

CHANGELOG.adoc
theogiraudet: Do I use `diagram elements` or `diagram's views`?
In this ADR, I propose to use an enum type to describe the state of the diagram element. This can be compared to a boolean because different possible states are incompatible with each other. Thus, an element is invisible (or faded) or not regardless of the reason. Do we have to implement a more flexible architecture to be able to handle the type of situation you mentioned in the future or should we keep it simple as I propose for the moment?

mbats:
> Also, what kind of UI is planned to reveal the hidden/faded elements?

We prepared 3 different versions, the idea is to first implement the v1, then the v2 and later the v3. For 2022-11 we focus only on v1.

Please find attached videos demonstrating the expected behaviours:
* version 1 : 2 actions in the palette to hide/fade an element and 2 actions in the toolbar to show all the hidden/faded elements
https://user-images.githubusercontent.com/1185863/193817834-3cf17e7e-9c6e-417d-8e82-32257530a4dd.mp4
* version 2 : same as v1 but we are able to maintain how many elements are hidden/faded and show this information in the toolbar
https://user-images.githubusercontent.com/1185863/193818803-f1f7a55d-44f2-484a-b483-24df3916e0ed.mp4
* version 3 : a dialog is available in the toolbar to select elements which must be hidden/faded
https://user-images.githubusercontent.com/1185863/193819079-4be71176-f1f9-4fc4-8410-6caf91b4b7e1.mp4

CHANGELOG.adoc
sbegaudeau: I think that there are two types of information we need to have for two different needs:

- On the frontend, we need to know if an element should be displayed on not. For that, we can use several solutions such as "not returning the element in question from the backend", "use a boolean to indicate that its hidden" or even "use an enum to indicate that its hidden"
- On the backend, we need to know why an element should be displayed. For that, we will need the details involved for example "some edge is hidden because it is between two nodes which are in two hidden containers because they are collapsed" (even if collapsing is out of scope for now)

So we may need to keep the modifiers placed on nodes and edges like in Sirius Desktop `Node#modifiers: List<Modifier>` with :

```
enum Modifier {
  HIDDEN,
  FADED,
  (maybe COLLAPSED one day)
}
```

But we may also need, for the frontend, to compute (and store?) the impact of those modifiers. This would prevent us from having to compute this impact every time we need to process a diagram. This information could be something like that: `Node#state: Modifier` and `Edge#state: Modifier`.

For example, let's consider the diagram in this screenshot: 

Capture d’écran 2022-10-14 à 15 26 13

If `Central_Unit` has for `modifiers` the values `[HIDDEN, FADED]`, then the following elements would be impacted with the following `state`:

- `DSP`: `HIDDEN`
- `Motion_Engine`: `HIDDEN`
- The fan with the label `100`: `HIDDEN`
- The edge with the label `10` between `DSP` and `Motion_Engine`: `HIDDEN`
- The edge with the label `6` between `Radar_Capture` and `Motion_Engine`: `HIDDEN`
- The edge with the label `10` between `GPU` and `Motion_Engine` should have the state `HIDDEN`

If on top of that, the `GPU` has for `modifiers` the values `[HIDDEN, FADED]`, then:

- The edge with the label `6` between `Back_Camera` and `GPU` should also have the state `HIDDEN`

As such, we need two pieces of information, the `modifiers` applied on the elements by the user and their impact on the `state` of a diagram element.

What do you think? @theogiraudet @pcdavid

theogiraudet: That sounds good to me, I don't see any case where this proposal doesn't work.

I see that you use a `List` to store the different modifiers. Is there a situation where we can have the same modifier more than once, e.g. `[HIDDEN, HIDDEN]` or is this an impossible case?

sbegaudeau: I don't think we have any reason to have a `List` instead of a `Set` indeed

doc/adrs/070_add_the_possibility_to_hide_or_fade_a_graphic_element.adoc
sbegaudeau: I don't think that we should use CSS for that, we can simply not tell Sprotty about the hidden elements. We should just ignore those hidden elements in the `convertDiagram` function in the frontend

sbegaudeau (approved):

PR #1371, opened 2022-09-27
[releng] Swith to Spring Boot 2.7.5

pcdavid:
See:
* https://github.com/spring-projects/spring-boot/releases/tag/v2.7.3
* https://github.com/spring-projects/spring-boot/releases/tag/v2.7.4

Spring Boot 2.7.3 includes a move to [GraphQL Java 18.3](https://github.com/graphql-java/graphql-java/releases/tag/v18.3) which includes a security fix.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -5,6 +5,7 @@
 === Dependency update
 
 - [backenđ] Upgrade to Apache Batik 1.16.0 (from 1.14.0)
+- [backend] Switch to https://github.com/spring-projects/spring-boot/releases/tag/v2.7.5[Spring Boot 2.7.5]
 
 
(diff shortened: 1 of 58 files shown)


pcdavid:
https://spring.io/blog/2022/10/20/spring-boot-2-7-5-available-now

sbegaudeau:
Everything looks good but there is either a new warning or a new issue (I think it's a new warning) regarding some threads which are not properly stopped when the server is closing. Given the name of the threads, I think that this warning may be quite relevant.

We should try to force a naming convention for the threads created using reactor (maybe in the multicast flux subscription or something like that) to detect where the issue is coming from.

sbegaudeau (approved):

PR #1430, opened 2022-10-19
[1426] Add support for optional text completion on textfield widgets

### LINKED ISSUE #1426: Add support for optional text completion on textfield widgets
When editing text fields which are not "free form" but correspond to some constrained DSL, it would be nice to have support for completion proposals (computed on the backend). For example: AQL expressions or domain types in view definitions.

This should be optional, only enabled on _Textfield_ widgets which explicitly provide a completion proposal provider. Ideally, there should be some visible affordance on such widgets to tell the end-user that completion is available on this widget.

The user-experience should be globally similar to what is customary in editors and IDEs like VS Code or Eclipse, but without all the bells and whistles:

* triggered on some key combination (Ctrl+Space if possible), which sends a request to the backend;
* upon receiving the possible completion proposals from the backend, open them in a list at the cursor's position;
* if the user selects one of the proposals (using the mouse or, ideally, using the arrow keys), insert the proposal's text.

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/1426

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/forms/backend/sirius-components-collaborative-forms/src/main/resources/schema/form.graphqls
+++ b/packages/forms/backend/sirius-components-collaborative-forms/src/main/resources/schema/form.graphqls
@@ -191,6 +191,7 @@ type Textarea implements Widget {
   iconURL: String
   value: String!
   style: TextareaStyle
+  supportsCompletion: Boolean
 }
 
 type TextareaStyle {
@@ -210,6 +211,7 @@ type Textfield implements Widget {
   iconURL: String
   value: String!
   style: TextfieldStyle
+  supportsCompletion: Boolean
 }
 
 type TextfieldStyle {
@@ -347,6 +349,13 @@ type RichText implements Widget {
 type FormDescription implements RepresentationDescription {
   id: ID!
   label: String!
+  completionProposals(widgetId: ID!, currentText: String!, cursorPosition: Int!): [CompletionProposal!]!
+}
+
+type CompletionProposal {
+  description: String!
+  textToInsert: String!
+  charsToReplace: Int!
 }
 
 extend type Mutation {
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,74 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug All sirius-components-core Tests",
+      "autoAttachChildProcesses": true,
+      "skipFiles": ["<node_internals>/**", "**/node_modules/**"],
+      "program": "${workspaceRoot}/node_modules/vitest/vitest.mjs",
+      "args": [
+        "--config",
+        "${workspaceFolder}/packages/core/frontend/sirius-components-core/vite.config.js",
+        "run",
+        "--dir",
+        "${workspaceFolder}/packages/core/frontend/sirius-components-core/src"
+      ],
+      "smartStep": true,
+      "console": "integratedTerminal"
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug All sirius-components-forms Tests",
+      "autoAttachChildProcesses": true,
+      "skipFiles": ["<node_internals>/**", "**/node_modules/**"],
+      "program": "${workspaceRoot}/node_modules/vitest/vitest.mjs",
+      "args": [
+        "--config",
+        "${workspaceFolder}/packages/forms/frontend/sirius-components-forms/vite.config.js",
+        "run",
+        "--dir",
+        "${workspaceFolder}/packages/forms/frontend/sirius-components-forms/src"
+      ],
+      "smartStep": true,
+      "console": "integratedTerminal"
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug All sirius-components-formdescriptioneditors Tests",
+      "autoAttachChildProcesses": true,
+      "skipFiles": ["<node_internals>/**", "**/node_modules/**"],
+      "program": "${workspaceRoot}/node_modules/vitest/vitest.mjs",
+      "args": [
+        "--config",
+        "${workspaceFolder}/packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/vite.config.js",
+        "run",
+        "--dir",
+        "${workspaceFolder}/packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src"
+      ],
+      "smartStep": true,
+      "console": "integratedTerminal"
+    },
+
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug All sirius-components-trees Tests",
+      "autoAttachChildProcesses": true,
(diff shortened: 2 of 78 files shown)


packages/forms/backend/sirius-components-collaborative-forms/src/main/resources/schema/form.graphqls
```diff
@@ -498,3 +500,29 @@ type PushButtonSuccessPayload {
...
+type CompletionRequestSuccessPayload {
+  id: ID!
+  proposals: [CompletionProposal!]!
+}
+
+type CompletionProposal {
+  description: String!
+  textToInsert: String!
+  charsToReplace: Int!
+}
```
sbegaudeau: I wouldn't use this approach to plug the completion proposals in the GraphQL schema for a couple of reasons:

- The input / payload pattern is used for mutations, you don't need it here since you are not mutating anything on the backend. You do not need an error payload here either. We don't have any for our queries and we have a lot of queries.
- You may need an internal Input / Payload to compute the data in the editing context but that's an internal concern that does not need to appear for consumers of the API (for example by requiring them to provide an UUID id)
- You are adding a very specific field on the editing context while we have made sure instead to keep dedicated fiels like that on the representation descriptions instead. Have a look at the way we compute, the palette, the initial label or the context of the connector tool.
- Since you have this new field all the one on the top of the editing context and since it's an input, you have to give it some parameters that are not required such as id, editingContextId and the representation id. Speaking about the editingContextId, with your approach you need to give it twice (once to find the editing context object and another time as the input). We solved the issue for both the editing context id and the representation id already to compute those kinds of fields for diagrams (for example for the palette). For example with the following query:

```
  query getToolSections($editingContextId: ID!, $diagramId: ID!, $diagramElementId: ID!) {
    viewer {
      editingContext(editingContextId: $editingContextId) {
        representation(representationId: $diagramId) {
          description {
            ... on DiagramDescription {
              toolSections(diagramElementId: $diagramElementId) {
                label
                // Additional fields to compute the palette
              }
            }
          }
        }
      }
    }
  }
```

We have thus setup everything to propagate both the editing context id and the representation id to all child data fetchers recursively, see the usage of `LocalContextConstants` on the backend. Your query should instead be:


```
  query completionProposals($editingContextId: ID!, $formId: ID!, $widgetId: ID!, $currentText: String!, $cursorPosition: Int!) {
    viewer {
      editingContext(editingContextId: $editingContextId) {
        representation(representationId: $formId) {
          description {
            ... on FormDescription {
              completionProposals(widgetId: $widgetId, currentText: $currentText, cursorPosition: $cursorPosition) {
                // Specific fields
              }
            }
          }
        }
      }
    }
  }
```

We have successfully applied this pattern on ContextMenu.tsx, ContextualPalette.tsx and EditLabelUIWithInitialContent.ts for example.

sbegaudeau (changes requested):

sbegaudeau:
I still a have a thing or two to check but those are very minor details and from what I have tested it works very nicely (I don't take into account how relevant the proposals are).

packages/forms/backend/sirius-components-collaborative-forms/src/main/resources/schema/form.graphqls
pcdavid: Ths proposed approach works for diagrams, but can not work (at least not as-is) for forms (or more precisely: for properties forms).

The query shape

```
    viewer {
      editingContext(editingContextId: $editingContextId) {
        representation(representationId: $formId) {
          description {
```

invokes `EditingContextRepresentationDataFetcher` to find the representation, but this data fetcher uses `this.representationService.getRepresentationDescriptorForProjectId(editingContextId, representationId)` whose implementation is to look into the `IRepresentationRepository`.
Forms used for the details view are not persisted, so not visible from the `IRepresentationRepository`.

I guess `RepresentationService` could be extended to also look into the active, non-persisted representations, from the `editingContextEventProcessorRegistry`. I'll try that to check if it works, but this seems a significant and possibly risky change.
WDYT?

pcdavid: Hmm, injecting `IEditingContextEventProcessorRegistry` into `RepresentationService` create a dependency cycle.

```
The dependencies of some of the beans in the application context form a cycle:

   editingContextActionsDataFetcher defined in file [sirius-components/packages/sirius-web/backend/sirius-web-graphql/target/classes/org/eclipse/sirius/web/graphql/datafetchers/editingcontext/EditingContextActionsDataFetcher.class]
┌─────┐
|  editingContextEventProcessorRegistry defined in file [sirius-components/packages/core/backend/sirius-components-collaborative/target/classes/org/eclipse/sirius/components/collaborative/editingcontext/EditingContextEventProcessorRegistry.class]
↑     ↓
|  editingContextEventProcessorFactory defined in file [sirius-components/packages/core/backend/sirius-components-collaborative/target/classes/org/eclipse/sirius/components/collaborative/editingcontext/EditingContextEventProcessorFactory.class]
↑     ↓
|  deleteRepresentationEventHandler defined in file [sirius-components/packages/sirius-web/backend/sirius-web-services/target/classes/org/eclipse/sirius/web/services/representations/DeleteRepresentationEventHandler.class]
↑     ↓
|  representationService defined in file [sirius-components/packages/sirius-web/backend/sirius-web-services/target/classes/org/eclipse/sirius/web/services/representations/RepresentationService.class]
└─────┘
```

sbegaudeau: Not being able to handle in memory representations would indeed be an issue. We could modify the support of the deletion of representations by simply making DeleteRepresentationEventHandler depend on IRepresentationRepository directly because RepresentationService#delete is completely useless. It's only used at one location and it only hides one line of code.

sbegaudeau: Now that I think about it since tou have the id of the representation can't we tell that this representation is one of the detail view?

pcdavid: I don't think so, `PropertiesConfiguration` computes the id like this:
```javadoc
this.formId = UUID.nameUUIDFromBytes((PROPERTIES_PREFIX + objectIds).getBytes()).toString();
```

pcdavid:
I've just pushed a solution that kinda works, but has to work around several issues and feels wrong/ugly.
I've kept the changes in separate patches. This first one is the only one that should be needed ideally, the others are needed to make it work (at least the best I have found for now).

sbegaudeau (approved):

sbegaudeau:
I'll wait for the build to merge this

PR #1436, opened 2022-10-24
[doc] Add ADR about improving the support for direct edit

gcoutable:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -6,6 +6,7 @@
 
 - [ADR-075] Add support for a basic image widget
 - [ADR-076] Allow users to upload their own custom images
+- [ADR-077] Improve support for direct edit in diagram
 
 === Dependency update
 
--- a/doc/adrs/077_improve_support_for_direct_edit.adoc
+++ b/doc/adrs/077_improve_support_for_direct_edit.adoc
@@ -1,0 +1,68 @@
+= ADR-077 - Improve support for direct edit in diagram
+
+== Context
+
+The current direct edit behavior may make the user sends data that is not meant to be sent to the backend, like the feature multiplicity in the Domain diagram. 
+We want to be able to define an initial value that will be displayed when the direct edit is triggered.
+
+== Decision
+
+A request using the graphql api will be made to ask the backend the value to put in the direct edit text field.
+This request will be made only if the user hit 'F2' or clicked on the direct edit tool from the palette.
+If the user start to type, the direct edit text field will be initialized with what the user has typed.
+
+=== Backend
+
+In the graphql schema, add `initialDirectEditElementLabel(labelId: ID!): String!` field on `DiagramDescription` type.
+This field will be resolved thanks to the `DiagramDescriptionInitialDirectEditElementLabelDatafetcher` which will leverage the `InitialDirectEditElementLabelEventHandler` to call either the compatibility layer or the View DSL to calculate the value to put in the direct edit text field thanks to `IInitialDirectEditElementLabelProvider`.
+
+```java
+public interface IInitialDirectEditElementLabelProvider {
+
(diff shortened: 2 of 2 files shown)


doc/adrs/075_improve_support_for_direct_edit.adoc
```diff
@@ -0,0 +1,37 @@
...
+The current direct edit behavior may make the user sends data that is not meant to be sent to the backend, like the feature multiplicity in the Domain diagram. 
+We want to be able to define an initial value that will be displayed when the direct edit is triggered.
+
+== Decision
+
+The description of a label should be able to provide the value used to initialize the direct edit textfield.
+
+=== Backend
+
+Add the optional `initialEditTextProvider` to `LabelDescription`.
```
sbegaudeau: If I understand everything here, you want to compute the label which should be used if the user trigger a direct edit on any element of the diagram even if it’s not used for every refresh?

We would end up computing something we know is not useful 99% of the time hundreds of time every refresh.

I don’t see how it’s a viable option compared to asking what we need from the backend when we need it. When the user want to edit a label, we just ask for the initial value.

sbegaudeau (changes requested):

doc/adrs/075_improve_support_for_direct_edit.adoc
pcdavid: That was the first approach we envisioned, but it's more complex (needs a new GraphQL query, needs to handle an asynchronous request to the backend on an operation which is partially handled by Sprotty and not under our complete control, and that the user expects to be immediate).

Yes it would be better, but given the time constraints we have, the reasoning was that the proposed design (equivalent to compute a second/alternative label text):
* is much much simpler/faster to implement;
* will have zero performance drawback on existing diagrams which do not specify this optional `initialEditLabelExpression`;
* even on diagrams which will use this, the overhead will only be payed for the subset of visible labels which want to make use of this;
* it can be seen as a first implementation, and *if* this causes actual measurable performance issues, we can always change the implementation to be more lazy afterwards, without changing the View DSL part.

sbegaudeau: We are creating new queries or mutations in the GraphQL API all the time, I don't think that creating a new one is really an issue. For the behavior, we are already extending the `EditLabelUI`, by changing the behavior of `applyTextContents` we could change the way the initial value is computed.

It could be prototyped in a couple of minutes with a simple `setTimeout()` to figure out potential issues with asynchronous initialization. From there, triggering a HTTP request should not be difficult.

Unless, we have a specific reason to say that it's difficult to do, I don't see the issue in doing things right.

sbegaudeau: I have discussed this with @gcoutable and we think that the solution that I propose should be viable. He will test it to see if there's something fishy with this approach and if everything is ok, he will start implementing it fully.

doc/adrs/075_improve_support_for_direct_edit.adoc
```diff
@@ -0,0 +1,37 @@
...
+Add the optional `initialEditText` to `Label`.
+Update `LabelComponent` accordingly
+Update the graphql schema to add `initialEditText` as an optional value (allow `null` value)
```
sbegaudeau: The approach chosen has changed, please update this ADR accordingly

sbegaudeau (changes requested):

sbegaudeau (approved):

PR #1532, opened 2022-12-14
Improve AQL completion

pcdavid:
- [enh] Use the domain type to provide more precise AQL completion proposals
- [enh] Sort feature proposals before service calls

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/emf/backend/sirius-components-interpreter/src/main/java/org/eclipse/sirius/components/interpreter/AQLInterpreter.java
+++ b/packages/emf/backend/sirius-components-interpreter/src/main/java/org/eclipse/sirius/components/interpreter/AQLInterpreter.java
@@ -38,6 +38,7 @@ import org.eclipse.acceleo.query.validation.type.EClassifierType;
 import org.eclipse.acceleo.query.validation.type.IType;
 import org.eclipse.emf.common.util.BasicDiagnostic;
 import org.eclipse.emf.common.util.Diagnostic;
+import org.eclipse.emf.ecore.EClass;
 import org.eclipse.emf.ecore.EPackage;
 import org.eclipse.emf.ecore.EcorePackage;
 import org.eclipse.emf.ecore.impl.EStringToStringMapEntryImpl;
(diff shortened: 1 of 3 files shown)


sbegaudeau (changes requested):
While I don't have any issues with the code, before the contribution, I had in the completion on `all:self.` some useful methods like `eContents()`, `eClass()` and tons of useless proposals but I didn't have access to my attributes or references. Now I don't have the useful or useless methods anymore, I don't have anything at all.

There seems to be an issue with the recognition of the type by AQL since the use of the `selfType` can lead to no proposal at all.

Capture d’écran 2022-12-16 à 09 50 46

Here I would have expected something like `name` and `containers` and maybe the useful methods like `eClass()` or `eContents()`. I suspect that domain based metamodels never worked properly but now by considering the real type of the object, you can easily end up with no proposals at all...

pcdavid:
It seems there are at least two issues:

1. The domain being defined inside the current project, it is not visible from inside the project itself. `AQLTextfieldCustomizer.getAccessibleEPackages(IEditingContext)` (like all other implementations of this) only looks for `EPackage`s inside the current resource set, which was created before the `discover` domain existed. This is a long known limitation that you can not see/use the definition of a domain from the project it was created in (at least not in a reliable way: you can see an earlier version which existed before you opened the project, which may or may not match what you consider the current state), but before the introduction of this completion feature, well View only dealt with plain strings which did not get interpreted until actually instanciated in a representation, so this was less of an issue. In practice this means domain type completion does not work for domain types defined in the same project...
2. Even if I create my View definition in a separate project, where the domain can be seen/resolved, for some reason I do not understand yet, the AQL interpreter we create, even if given the proper `EClass` (with attributes and all), does not provide any result for completion on `self.`. Not the EClass's features, not even the generic EObject-based services.

PR #1573, opened 2023-01-11
Implement default/canonical behaviors as AQL-invocable services

### LINKED ISSUE #1562: Make the default/canonical tools' behavior callable from standard AQL expressions and/or operations
The behaviors of the "canonical" tools which are applied on a View-based diagram when no tool is specified is hard-coded inside an opaque/internal class.

It should be possible to invoke them explicitly from inside a user-defined tools, using either a plain AQL expressions (through Java services) or if that is not possible through reified `org.eclipse.sirius.components.view.Operation`s.

pcdavid:
- [1569] Only delegate semantic deletion to the element's Delete Tool
- [1562] Implement default/canonical behaviors as AQL-invocable services

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/ViewDiagramDescriptionConverter.java
+++ b/packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/ViewDiagramDescriptionConverter.java
@@ -85,7 +85,6 @@ import org.eclipse.sirius.components.view.RepresentationDescription;
 import org.eclipse.sirius.components.view.SourceEdgeEndReconnectionTool;
 import org.eclipse.sirius.components.view.TargetEdgeEndReconnectionTool;
 import org.eclipse.sirius.components.view.ViewPackage;
-import org.eclipse.sirius.components.view.emf.CanonicalBehaviors;
 import org.eclipse.sirius.components.view.emf.IRepresentationDescriptionConverter;
 import org.eclipse.sirius.components.view.emf.diagram.providers.api.IViewToolImageProvider;
 import org.springframework.stereotype.Service;
@@ -112,8 +111,6 @@ public class ViewDiagramDescriptionConverter implements IRepresentationDescripti
 
     private final StylesFactory stylesFactory;
 
-    private final CanonicalBehaviors canonicalBehaviors;
-
     private final Function<VariableManager, String> semanticTargetIdProvider;
 
     private final Function<VariableManager, String> semanticTargetKindProvider;
@@ -134,7 +131,6 @@ public class ViewDiagramDescriptionConverter implements IRepresentationDescripti
         this.semanticTargetIdProvider = variableManager -> this.self(variableManager).map(this.objectService::getId).orElse(null);
         this.semanticTargetKindProvider = variableManager -> this.self(variableManager).map(this.objectService::getKind).orElse(null);
         this.semanticTargetLabelProvider = variableManager -> this.self(variableManager).map(this.objectService::getLabel).orElse(null);
-        this.canonicalBehaviors = new CanonicalBehaviors(objectService, editService);
         this.viewToolImageProvider = Objects.requireNonNull(viewToolImageProvider);
     }
 
@@ -342,6 +338,7 @@ public class ViewDiagramDescriptionConverter implements IRepresentationDescripti
                         .handler(variableManager -> {
                             VariableManager child = variableManager.createChild();
                             child.put(CONVERTED_NODES_VARIABLE, capturedConvertedNodes);
+                            child.put("nodeDescription", nodeDescription);
                             return new DiagramOperationInterpreter(converterContext.getInterpreter(), this.objectService, this.editService, this.getDiagramContext(child), capturedConvertedNodes).executeTool(nodeTool, child);
                         })
                         .targetDescriptions(allTargetDescriptions)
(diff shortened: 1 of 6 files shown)


packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/ViewDiagramDescriptionConverter.java
```diff
@@ -680,18 +695,26 @@ private void deleteFromDiagram(VariableManager variableManager) {
...
-            };
-        } else {
-            return this.canonicalBehaviors::editLabel;
-        }
+            } else {
+                Result aqlResult = converterContext.getInterpreter().evaluateExpression(childVariableManager.getVariables(), "aql:self.defaultEditLabel(newLabel)");
+                if (aqlResult.getStatus() == Status.OK) {
+                    result = new Success();
+                } else {
+                    result = new Failure("Error while deleting element");
```
sbegaudeau: Not `deleting`

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/ViewDiagramDescriptionConverter.java
```diff
@@ -359,7 +356,13 @@ private List<ITool> getNodeTools(ViewDiagramDescriptionConverterContext converte
                         .handler(variableManager ->  {
                             VariableManager child = variableManager.createChild();
                             child.put(CONVERTED_NODES_VARIABLE, capturedConvertedNodes);
-                            return this.canonicalBehaviors.createNewNode(nodeDescription, variableManager);
+                            child.put("nodeDescription", nodeDescription);
+                            Result result = converterContext.getInterpreter().evaluateExpression(child.getVariables(), "aql:self.defaultCreateNode(nodeDescription)");
```
sbegaudeau, line 360: I thought that we were soon moving away from having custom variables in the view DSL by removing `CONVERTED_NODES_VARIABLES` with the work on representation description identifier.

Why use a new variable to simply create a node? I thought you were trying to create model operations as Java services. Something like `all:self.createInstance('feature', 'domain::Type')` would be more easy to manipulate. You don't really use the node description in the service and if the service to create a node is so tightly connected to a node description it may not be easy to reuse for anything beyond the most basic use cases.

sbegaudeau: These new variables "nodeDescription" and "edgeDescription" could be a nice way to get rid of "convertedVariables" but after thinking about it a bit more I think that they still create a sizable issue. If I understand things correctly, the "nodeDescription" variable would not, as of today, represent the description of the object on which we have clicked nor what we may want to create but the description of where the tool is being defined in the View DSL.

Let’s say that we open the palette on the background of a papaya diagram, now we would have X tools with each a different "nodeDescription" variables because we imagine that each one of those is a tool used to create the semantic element defined on the node description holding them.

Since we are trying to move into a direction with "less magic and assumptions on what the user may do", the issue we encounter is that to support "creating default tools that do what we think the user may want", we need those kinds of magic variables. The service "createDefaultNode" versus a "createInstance" is not an issue because that’s not where the magic happens. The core of the magic is "how can we find out the semantic type of what we think the user want to create".

The only clean but more expensive and very limited approach that I can think of would be to have a wizard to let the user type the expected domain type and create both the node description and a tool based on create instance. Yet such tool would stop working if the user started to edit its description to change the domain type for example.

If we still move to a system where defining a tool on a node means that said tool will appear when we click on the node, how would such a variable continue to work? How could we give a relevant "nodeDescription" for top level tool defined on the diagram description? Should we give up this goal and keep instead that tools defined on a NodeDescription are used to create this kind of nodes? You may have thought about that more than me, do you have an idea of how would that work?

I’m more and more convinced that we can’t have a fully dynamic system which both "create something which works magically" and not "create some stuff which will get in the way of anything non trivial". I don't really think there is another way of doing what you are trying to do here, I'm just having serious doubts that we can fulfill all our requirements.

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/ViewDiagramDescriptionConverter.java
pcdavid: Bad copy/paste, I'll fix it.

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/ViewDiagramDescriptionConverter.java
pcdavid: > Something like `all:self.createInstance('feature', 'domain::Type')` would be more easy to manipulate.

Yes, but that would mean that if the studio maker edits the description's domain type the default expression created here would become obsolete.

pcdavid: > If I understand things correctly, the "nodeDescription" variable would not, as of today, represent the description of the object on which we have clicked nor what we may want to create but the description of where the tool is being defined in the View DSL.

True at the moment, but this is supposed to change with https://github.com/eclipse-sirius/sirius-components/issues/1592. The ADR in  https://github.com/eclipse-sirius/sirius-components/pull/1595 describes the overall design I have in mind for that, but not all issues are completely handled, especialy this one.

Anyway, the scope of *this* particular ticket and PR does not include these changes; it's supposed to "work as before" (functionally), just more explicitly. Maybe the new variables introduced will end up being just temporary workarounds if we find a new overall design which works and does not need them in time for 2023.3.0.

sbegaudeau (approved):

PR #1587, opened 2023-01-17
[1401] The setSelection is called even if the selection did not changed

### LINKED ISSUE #1401: The setSelection is called even if the selection did not changed
* [x] **I have checked that this bug has not yet been reported by someone else**
* [x] **I have checked that this bug appears on Chrome**
* [x] **I have specified the version** :
- latest 2022.9.3

* [x] **I have specified my environment** : 
- All

### Screenshots

...

### Steps to reproduce

scenario 1
* open any model 
* open the dev tools to see the subscription messages
* in the tree explorer, click on any element -> the Form subscription is initialized
* re-click on the same element 
KO -> the Form subscription is stopped and another one is initialized

scenario 2
* same beghavior when clicking twice on a diagram element

### Expected behavior

If the selection does not change, nothing should happen

### Actual behavior

...

mcharfadi:
Bug: #1401
Signed-off-by: Michaël Charfadi <michael.charfadi@obeosoft.com>

### DIFF
--- a/packages/core/frontend/sirius-components-core/src/workbench/WorkbenchMachine.ts
+++ b/packages/core/frontend/sirius-components-core/src/workbench/WorkbenchMachine.ts
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021, 2022 Obeo.
+ * Copyright (c) 2021, 2023 Obeo.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
@@ -117,6 +117,7 @@ export const workbenchMachine = Machine<WorkbenchContext, WorkbenchStateSchema, 
               UPDATE_SELECTION: {
                 target: 'initial',
                 actions: 'updateSelection',
+                cond: 'isNewSelection',
               },
               HIDE_REPRESENTATION: {
                 target: 'initial',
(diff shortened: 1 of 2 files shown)


packages/core/frontend/sirius-components-core/src/workbench/WorkbenchMachine.ts
```diff
@@ -117,6 +120,7 @@ export const workbenchMachine = Machine<WorkbenchContext, WorkbenchStateSchema,
               UPDATE_SELECTION: {
                 target: 'initial',
                 actions: 'updateSelection',
+                cond: (context, event) => !equals(context.selection, event.selection),
```
sbegaudeau: I'll move this code to the guard section of the state machine, it's easier to read than just inline like that.

sbegaudeau (changes requested):

packages/core/frontend/sirius-components-core/src/workbench/WorkbenchMachine.ts
```diff
@@ -77,6 +77,9 @@ const isRepresentationRenamedEventPayload = (
   payload: GQLEditingContextEventPayload
 ): payload is GQLRepresentationRenamedEventPayload => payload.__typename === 'RepresentationRenamedEventPayload';
 
+const equals = (a: Selection, b: Selection) =>
+  a.entries.length === b.entries.length && a.entries.every((v, i) => v.id === b.entries[i].id);
+
```
sbegaudeau: Please make sure to use meaningful variable names instead of variables with one character.

sbegaudeau (changes requested):
In your commit message, you should have the full link of the issue, not just `Bug: #1401`

sbegaudeau (approved):

PR #1666, opened 2023-02-02
[1584] Improve node size handling

### LINKED ISSUE #1584: Keep custom node size on Arrange All when possible
When the end-users resizes a node, we mark it with `CustomizableProperties.Size`.
This is supposed to mean "the size of this node has been set explicitly by the end-user and should be kept if possible".

The incremental layout does take this flag in consideration, but not ELK, which will revert these nodes to their default size.

For container nodes that is expected, as this is the point of the "Arrange All" to re-compute these, and trying to force a size set by the user could lead to invalid layouts.

For image nodes the issue is less clear-cut, as is depends on different factors, notably if we want to enforce the image's aspect-ratio.

For rectangle leaf nodes (i.e. nodes with no chilren), we should try to conserve the size set by the end-user.

pcdavid:
- Add support for non-resizable nodes
- For resizable rectangular nodes (and only them), keep the user-set size during layout if possible (compatible for other constraints like minimum size to hold all its content).

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -90,6 +90,9 @@ A node can be collapsed or expanded if its view description is collapsible or if
 - https://github.com/eclipse-sirius/sirius-components/issues/1624[#1624] [diagram] The `collapsingState` of a node is now available as a variable to compute its style.
 As a result, it is possible to create a conditional style which will use the collapsing state quite easily.
 Using the view DSL, one can use the expression `aql:collapsingState.toString() = 'COLLAPSED'` as a condition to create a style which will be used when the node is collapsed.
+- https://github.com/eclipse-sirius/sirius-components/issues/1584[#1584] [diagram] Studio makers can indicate whether or not a node type can be resized by the end-user or not.
+Nodes which can not be resized (or have not been resized yet) haver their actual size computed from the NodeDescrition's sizeProvider in a uniform way for all node types.
+Nodes which can and *have* been resized keep their user-chosen size across both incremental and full layout (as long as the requested size is compatible with other constraints).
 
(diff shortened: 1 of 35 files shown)


pcdavid:
Robot Topography with the PR: nodes with computed size are updated when the model changes *until* the end-user overrides with an explicitly set size. Sizes are stable with incremental and full (ELK) layout.

https://user-images.githubusercontent.com/10608/216373270-ccc6ce99-c144-4b5c-92f9-7782f5ef750c.mp4

pcdavid:
View based modelers (example studio attached):
* nodes can (optionally) be marked as "non resizable" (by default they are)
* if non-resizable, the end-user can not resize them, and their size is *always* the one returned by their sizeProvider (widthExpression and heightExpression)
* if resizable, this is still the default, but as with the VSM-based Topography example above, as soon as the end-user has chosen an explicit size it is kept by incremental and full layouts.


https://user-images.githubusercontent.com/10608/216374533-37b428ad-c1e9-413d-b5c9-c5ad42871bc7.mp4

[size-studio.zip](https://github.com/eclipse-sirius/sirius-components/files/10570773/size-studio.zip)

sbegaudeau:
Is that use case suppose to be supported or not? Is there something I do wrong?

https://www.youtube.com/watch?v=iv2HEMSKbds

sbegaudeau:
> Is that use case suppose to be supported or not? Is there something I do wrong?
> 
> https://www.youtube.com/watch?v=iv2HEMSKbds

This was caused by the fact that 50x50 is way below the minimum size allowed for a node in our ELK configuration.

sbegaudeau (approved):

PR #1931, opened 2023-04-17
[1918] Add shape about the filtering of tree based representations

AxelRICHARD:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/1918

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,86 @@
...
+Each character added/removed in the search bar updates the highlighted tree items.
+User close the search bar with the close button.
+All highlighted items are not highlighted anymore.
+
+=== Scenario 2 - Filter
+
+User want to filter tree items containing a specific value as their label or part of their label.
+User type the shortcut to enable the search bar.
+User type the value to search in the search bar.
+The tree items containing the searched value are highlighted (only the part of the items containing the value).
```
mbats: How do we represent that there are some matches available under a collapsed item?

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,86 @@
...
+User push the filter button (to enable the filter) in the search bar.
+All visible leaf tree items that do not match the search are hidden.
+It will not concern collapsed root/intermediate tree items.
```
mbats: Are we ok about the fact that we keep the hierarchy of the tree and so we will still show the intermediate items in the hierarchy of a match ?

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,86 @@
...
+3) Filter tree items
+4) Navigate through highlighted tree items
+5) Display the number of highlighted tree items somewhere in the search bar.
+
+== Rabbit holes
+
+The navigation through up/down arrow keys could lead to some misunderstanding for the users.
+The keys are already used to navigate through the tree items in the Explorer View.
+When the search will be enabled, the up/down arrow keys won't have the same behavior than when the search is disabled.
+An alternate solution could be to have two buttons in the search bar, allowing to navigate through the highlighted tree items.
```
mbats: Both should exist buttons in the search bar to navigate and shortcuts, it is normal from my point of view to no have the same beahviout for the shortcut if the search is active or not

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+Some results may not be highlighted because they are collapsed. With the proposed solution, there is no simple way to access to those tree items. This could not be the behavior wanted by some users.
```
mbats: In this case the user must knwo that there is a match available under a parent item which is collapsed by showing in a tag the number of matches for instance.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,86 @@
...
+The documentation will have to explain that.
+
+The filtering could also be confusing for the users. Indeed, the filtering will not concern collapsed root/intermediate tree items. Only leaf tree items are concerned. This could not be the behavior wanted by some users.
```
mbats: As said earlier the end user needs to know that there is a match under a collpase item else this feature is useless.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,86 @@
+= (M) Shape
```
mbats: Replace "Shape" by the title of the shape : "Filtering tree based representations"

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+The solution should not slow-down the user interface. The user interface must remains as smooth as it was before the developments of this filter.
+The results displayed by the filter should be visible immediately, i.e. without any latency.
+Otherwise, if the solution can only be implemented with a latency, then during the computation time, a "busy" indicator should be displayed, informing the user his request is processing.
+
+== Solution
+
+Add a search capacity in the Explorer View and other tree based representations, by providing a textual filtering of the tree just like what is being done in VSCode using a simple textfield.
+
+=== Breadboarding
```
mbats: See the following breadboarding for the serach : Filter_Explorer_2

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
AxelRICHARD: Done

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
AxelRICHARD: Done

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
AxelRICHARD: It is not handled by this shape. The demand in https://github.com/eclipse-sirius/sirius-components/issues/1918 is to mimic the behavior of VSCode and the case you describe is not handled by VSCode.
Should we do something about it?

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
AxelRICHARD: yes we are

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
AxelRICHARD: Ok

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
AxelRICHARD: Ok, I will update the shape with the proposal.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
AxelRICHARD: Ok, I will update the shape with the proposal.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
mbats: Yes, if it is possible it should be part of the shape.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,92 @@
...
+The up/down arrow keys allow to jump from an highlighted item to the next one/previous one.
+
+=== Scenario 1 - Search
+
+User want to search tree items containing a specific value as their label or part of their label.
+User type the shortcut to enable the search bar.
+User type the value to search in the search bar.
+The tree items containing the searched value are highlighted (only the part of the items containing the value).
+Each character added/removed in the search bar updates the highlighted tree items.
+User close the search bar with the close button.
```
pcdavid: As an end-user, I expect that:
- once I've found the item I searched for, hitting "Return" finishes the search (closes the search bar) and selects the element in question.
- if I don't find what I'm looking for, hitting "Esc" closes the search bar and returns the explorer to the previous state.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+== Key Result
+
+The solution should not slow-down the user interface. The user interface must remains as smooth as it was before the developments of this filter.
+The results displayed by the filter should be visible immediately, i.e. without any latency.
+Otherwise, if the solution can only be implemented with a latency, then during the computation time, a "busy" indicator should be displayed, informing the user his request is processing.
+
+== Solution
+
+Add a search capacity in the Explorer View and other tree based representations, by providing a textual filtering of the tree just like what is being done in VSCode using a simple textfield.
```
pcdavid: It's not clear to me if the search/filtering only considers the tree items actually visible on the front (and thus can be fully implemented frontend-side), or if if (the search in particular) will also potentially reveal new elements (by asking the backend).

I think the proposal is for the first option, but unless I missed something this is never mentioned explicitly.

"like what is being done in VSCode" in particular can be confusing. In VS Code, even if all my projects are collapsed in the Explorer, it will look for my search term as deep as needed and expand new folders to reveal the matching files.

A purely frontend-based "search" feature look like it could confuse users. If I search for "foo" in my project, why would it "find" it only when the corresponding item was already expanded/visible, and then a few minutes later, just because I have collapsed the part of the tree that contains it the same "search" does not find it.
As a user, it means that if I search for something in my project and Sirius Web does not give me any match, I can not assume the element does not exist. Maybe it's present somewhere deep inside a hierarchy I have not expanded yet? But is it not the whole purpose of a search to find that for me?

pcdavid: Actually it is mentioned in the very last sentence of the "No go" section that all of this only considers visible tree items.
I think the term "search" should not be used as it gives wrong expectations.

pcdavid (dismissed):

doc/iterations/2023.6/scopes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,16 @@
+= Scope
```
sbegaudeau: No need to contribute the scope right now, it should be contributed when the work is done

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,92 @@
+:imagesdir: images/
+= (M) Filtering tree based representations
+
+== Problem
+
+It is quite difficult to search for/identify specific tree item(s) in a tree when this tree contains lots of tree items.
+
+== Key Result
+
+The solution should not slow-down the user interface. The user interface must remains as smooth as it was before the developments of this filter.
```
sbegaudeau: One sentence per line

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
sbegaudeau: +1 we are planning to add search support later this year probably, so there should not be anything here mentioning search. This is about filtering. In VS Code, there are both search and filtering support, we are only considering filtering for now.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,92 @@
...
+== Solution
+
+Add a search capacity in the Explorer View and other tree based representations, by providing a textual filtering of the tree just like what is being done in VSCode using a simple textfield.
+
+=== Breadboarding
+
+See the following breadboarding for the search:
+image:filterTreeBasedRepresentations-01.png[Filter widget]
+ 
+A shortcut (most likely ctrl+f) will be available in the Explorer View, allowing to display a search bar under the toolbar at the top of the Explorer View.
```
sbegaudeau: It should be noted that this may require having the ability to focus the explorer itself (this is what VS Code does). We don't support that today.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
sbegaudeau: While it may look nice since it's quite similar to VS Code, this floating filter area is more complex than what was imagined. Do we really want a persistent non-modal dialog there? A simple textfield with buttons under the toolbar wouldn't be enough?

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
sbegaudeau: I agree on Esc but since it's not searching, I don't see what hitting return should do when multiple tree items are matching the filter. Select them all? I don't see any behavior on return in VS Code

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
sbegaudeau: I don't see how the frontend could know that "there are some matches available under a collapsed item" so I don't see how that could be done. VS Code does not do it either because it's a complex question even for them.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,92 @@
...
+User type the up arrow key to select the previous highlighted tree item.
+When the last highlighted tree item is selected, the down arrow key has no effect.
+When the first highlighted tree item is selected, the up arrow key has no effect.
+
+=== Cutting backs
+
+Sort the things todo, define the nice-to-have. What could be removed if finally the feature does not fit the appetite.
+1) Search bar open/close
+2) Highlight tree items
+3) Filter tree items
```
sbegaudeau: Could we really remove any of those three and still consider that we have delivered the feature?

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,92 @@
...
+4) Navigate through highlighted tree items
+5) Display the number of highlighted tree items somewhere in the search bar.
```
sbegaudeau: Is that new? I don't see this number being mentioned before

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
sbegaudeau: One sentence per line.

This would not be possible while keeping the goal of the shape similar (frontend only). Even VS Code does not do it because it would require them to ask a lot of things to the extensions. Either we reconsider the whole feature with some backend support (which would have to pretty much perform a resolve all after each refresh to resolve all proxies in order to be sure that we don't miss a match).

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
sbegaudeau: One sentence per line.

We either reconsider the whole feature to work both frontend and backend or like in VS Code, the user will have to understand.

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
```diff
@@ -0,0 +1,92 @@
...
+When the search will be enabled, the up/down arrow keys won't have the same behavior than when the search is disabled.
+An alternate solution could be to have two buttons in the search bar, allowing to navigate through the highlighted tree items.
+
+Some results may not be highlighted because they are collapsed. With the proposed solution, there is no simple way to access to those tree items. This could not be the behavior wanted by some users.
+The documentation will have to explain that.
+
+The filtering could also be confusing for the users. Indeed, the filtering will not concern collapsed root/intermediate tree items. Only leaf tree items are concerned. This could not be the behavior wanted by some users.
+The documentation will have to explain that.
+
+For those two concerns, a solution could be to let users know that there is a match available under a parent item which is collapsed by showing in a tag the number of matches for instance.
```
sbegaudeau: No, this cannot be done with the constraint (frontend only) that have defined this task. This is a major no-go.

sbegaudeau (changes requested):

doc/iterations/2023.6/shapes/filterTreeBasedRepresentations.adoc
pcdavid: Yes, the "Return" part of my comment was when I thought this was about an actual "search" and not "filtering".

sbegaudeau (approved):

PR #1938, opened 2023-04-19
[1919] Shape Odesign converter

mcharfadi:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/1919

### DIFF
--- a/doc/iterations/2023.6/scopes/scope Odesign converter.adoc
+++ b/doc/iterations/2023.6/scopes/scope Odesign converter.adoc
@@ -1,0 +1,16 @@
+= Scope
+
+== First task
+
+- [] Shape written, reviewed and merged
+
+== Second task
+
+- [] First prototype of the interfaces
+- [] First prototype of the implementation
+- [] Improvement of the first prototype after the discovery of performance issues
+- [] Second prototype of the backend merged
+
+== Third task
+
+- [ ] Integrated tests updated
--- a/doc/iterations/2023.6/shapes/shape Odesign converter.adoc
+++ b/doc/iterations/2023.6/shapes/shape Odesign converter.adoc
@@ -1,0 +1,46 @@
+= (M) ODesign reader
+
+== Problem
+
+We can create representations descriptions programmatically, with the view model and with the compatibility layer that converts odesign files in representation descriptions.
+It will be more efficient to concentrate our efforts in the creation of representation description programmatically and with the view model.
+As such the odesign will be converted in a view model. We will be able to support more concepts of the odesign model in the view model in a near future (see https://github.com/eclipse-sirius/sirius-components/issues/1920)
+
+Using plain old Java code, we should be able to transform an odesign model into a view model while converting the diagram and form descriptions. 
+
+== Key Result
+
+We will use the Flow model to validate the scenario below (https://github.com/eclipse-sirius/sirius-components/issues/1922).
+
+== Solution
+
+This code should be provided in a dedicated project named sirius-components-odesign-converter
+
+package org.eclipse.sirius.components.odesign.converter;
+
+import org.eclipse.sirius.viewpoint.description.Group;
+import org.eclipse.sirius.components.view.View;
+
+public interface IOdesignConverter {
+    Optional<View> convert(Group group);
+}
+
+The implementation will use the the soon to be introduced view creation API (https://github.com/eclipse-sirius/sirius-components/issues/1921)
+
+=== Scenario
+
+A user will be able to use the new interface IOdesignConverter to convert an odesign file to a view model.
+
+=== Breadboarding
+
+no ui changes
+
+=== Cutting backs
+
+
+== Rabbit holes
+
+The implementation will use an iterative transformation approach meaning that the API should be decomposed in smaller APIS that convert specifics entities and that it should be possible to add more entities to be converted in the future.
+
+== No-gos
+


doc/iterations/YYYY-MM/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,45 @@
+= (appetite) Shape
```
mbats: Give a name to the shape, the appetite should be a M

doc/iterations/YYYY-MM/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+== Problem
+
+We can create representations descriptions programmatically, with the view model and with the compatibility package that converts odesign files in representation descriptions.
```
mbats: compatibility package -> compatibility layer

doc/iterations/YYYY-MM/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,45 @@
...
+It will be more efficient to concentrate our efforts in the creation of representation description programmatically and with the view model.
+As such the odesign will be converted in a view model. We will be able to support more concepts of the odesign model in the view model in a near future (see https://github.com/eclipse-sirius/sirius-components/issues/1920)
+
+Using plain old Java code, we should be able to transform an odesign model into a view model while converting the diagram and form descriptions. 
+
+== Key Result
+
+We will use the migration flow package https://github.com/eclipse-sirius/sirius-components/issues/1922 to validate the scenatio below.
```
mbats: We will use the Flow model to validate the scenario below.

doc/iterations/YYYY-MM/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,45 @@
...
+import org.eclipse.sirius.viewpoint.description.Group;
+import org.eclipse.sirius.components.view.View;
+
+public interface IOdesignConverter {
+    Optional<View> convert(Group group);
+}
+
+=== Scenario
+
+An user will be able to use the new interface IOdesignConverter to convert an odesign file to a view model.
```
mbats: An -> A

doc/iterations/YYYY-MM/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+=== Breadboarding
+
+no ui changes
+
+=== Cutting backs
+
+I would be nice to be able to upload an odesign document to instanciate a view in Sirius Web.
```
mbats: No, this is totally out of the scope.

doc/iterations/YYYY-MM/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+== Rabbit holes
+
+The Api should be decomposed in smaller apis that convert specifics enities and it should be possible de add more entities to be converted in the future.
```
mbats: enities -> entities

doc/iterations/YYYY-MM/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+== No-gos
+
```
mbats: Iterative transformation.

doc/iterations/YYYY-MM/2023.6/shapes/shape Odesign converter.adoc
mbats: Be more detailed about the generated code which APIs will it use ? the EMF one from the View DSL or the new APIs that will be created to simplify the creation of View DSL programatically ?

pcdavid:
The `iterations/YYYY-MM` folder is here as a template, the new files should be added directly in an iteration-specific folder, i.e. `iterations/2023.6`. See for example https://github.com/eclipse-sirius/sirius-components/pull/1931.

doc/iterations/2023.6/scopes/scope Odesign converter.adoc
```diff
@@ -0,0 +1,16 @@
...
+== Second task
+
+- [] First prototype of the interfaces
+- [] First prototype of the implementation
+- [] Improvement of the first prototype after the discovery of performance issues
+- [] Second prototype of the backend merged
+
+== Third task
+
+- [ ] Integrated tests updated
```
sbegaudeau, line 16: - Use a proper commit message, something like `[doc] Contribute the shape for the odesign to view converter` would be better. No need to reference the issue in your commit message since it would require you to add that to the CHANGELOG etc and you may even create multiple issues in the end for your task.
- Remove the scope file, it is intended to be maintained on your branch and be committed when the work is done. We will not merge this file beforehand and will not update it each time something is modified. You do not have to throw it away but this file will contain your notes while you are working on the feature. You will present how you are moving forward with your work thanks to this file but no need to start committing it right now.
- You can contribute all your shapes in one pull request, no need to create one pull request per task

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,46 @@
+= (M) ODesign reader
```
sbegaudeau, line 1: - Name the file something like `odesign_to_view_converter.adoc`, no need to mention that it's a shape once again and this file may be aggregated in a website so let's not use too many spaces because it creates unreadable URLs
- The title of the shape has nothing to do with an ODesign reader but instead something like `Convert odesign models to view models`

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+== Problem
+
+We can create representations descriptions programmatically, with the view model and with the compatibility layer that converts odesign files in representation descriptions.
```
sbegaudeau, line 5: convert to

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,46 @@
...
+It will be more efficient to concentrate our efforts in the creation of representation description programmatically and with the view model.
+As such the odesign will be converted in a view model. We will be able to support more concepts of the odesign model in the view model in a near future (see https://github.com/eclipse-sirius/sirius-components/issues/1920)
```
sbegaudeau, line 7: One sentence per line

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,46 @@
...
+This code should be provided in a dedicated project named sirius-components-odesign-converter
+
+package org.eclipse.sirius.components.odesign.converter;
+
+import org.eclipse.sirius.viewpoint.description.Group;
+import org.eclipse.sirius.components.view.View;
+
+public interface IOdesignConverter {
+    Optional<View> convert(Group group);
+}
```
sbegaudeau, line 26: Use backticks `` to mention technical names and triple backticks for code blocks ``````

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+The implementation will use the the soon to be introduced view creation API (https://github.com/eclipse-sirius/sirius-components/issues/1921)
```
sbegaudeau, line 28: So I'll need the shape of this part too

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+We can create representations descriptions programmatically, with the view model and with the compatibility layer that converts odesign files in representation descriptions.
+It will be more efficient to concentrate our efforts in the creation of representation description programmatically and with the view model.
+As such the odesign will be converted in a view model. We will be able to support more concepts of the odesign model in the view model in a near future (see https://github.com/eclipse-sirius/sirius-components/issues/1920)
+
+Using plain old Java code, we should be able to transform an odesign model into a view model while converting the diagram and form descriptions. 
+
+== Key Result
+
+We will use the Flow model to validate the scenario below (https://github.com/eclipse-sirius/sirius-components/issues/1922).
```
sbegaudeau, line 13: While I agree in theory, it looks like you are saying that during this iteration, you have four tasks:

- convert an odesign to a view model
- simplify the programmatic creation of view models
- support additional features in view to match what's available in the odesign compatibility layer
- migrate flow to prove that everything is working and we can get rid of our odesign compatibility layer

Here it feels that none of the steps 1 to 3 can be done without 4.
My gut feeling is that this would end up with all the tasks being started at the same time and I'm not even sure that you could create a relevant commit for each task specifically.

So if some task is required to be done before another, it needs to be explicit first (which comes bask to the fact that you should pretty much create one PR with all the shapes at once).

Now, back to the question, how can we ensure that the commit that you will deliver to provide the converter works?

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,46 @@
...
+=== Breadboarding
+
+no ui changes
+
+=== Cutting backs
+
+
+== Rabbit holes
+
+The implementation will use an iterative transformation approach meaning that the API should be decomposed in smaller APIS that convert specifics entities and that it should be possible to add more entities to be converted in the future.
```
sbegaudeau, line 43: I'm not sure that it's a rabbit hole, it's an interesting detail to consider something to be extendable in the future but that's not a risk. Otherwise, specify why it's a risk.

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
```diff
@@ -0,0 +1,46 @@
...
+
+The implementation will use the the soon to be introduced view creation API (https://github.com/eclipse-sirius/sirius-components/issues/1921)
+
+=== Scenario
+
+A user will be able to use the new interface IOdesignConverter to convert an odesign file to a view model.
+
+=== Breadboarding
+
+no ui changes
```
sbegaudeau, line 36: Uppercase at the beginning of sentences

sbegaudeau (changes requested):

doc/iterations/2023.6/scopes/scope Odesign converter.adoc
sbegaudeau: And use the new title of the commit message as the title of the pull request

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
sbegaudeau: Additional question that needs to be answered:

- Are you planning to only deliver a converter when all the features of the odesign compatibility layer have been migrated to the view DSL?

doc/iterations/2023.6/shapes/shape Odesign converter.adoc
sbegaudeau: Since all four shapes are related, the order of the tasks should clearly appear as one of the first thing discussed.

PR #1948, opened 2023-04-27
[1914] Add shape for custom widgets support

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-components/issues/1914

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,6 +12,7 @@
 - Add support for the features of the compatibility layer not supported in the view DSL
 - Add an example showing how to contribute to the details view with the View DSL
 - Improve the layout of simple nodes
+- Support custom widgets in Form representations
 
 === Architectural decision records
 
--- a/doc/iterations/2023.6/shapes/support_custom_widgets_in_form_representations.adoc
+++ b/doc/iterations/2023.6/shapes/support_custom_widgets_in_form_representations.adoc
@@ -1,0 +1,147 @@
+= (M) Support custom widgets in Form representations
+
+== Problem
+
+The _Form_ representation currently supports a fixed set of generic widgets.
+Even if we enrich this set, it will never account for all the specific needs of concrete applications.
+The Sirius Components framework should be open for custom applications to provide their own widgets.
+
+If an application provides such a custom widget, its use both by the studio maker and by the end users should be indisguishable from the core widgets.
+
+== Key Result
+
+An application based on Sirius Component should be able to extend the _Form_ representation with additional, custom widget definitions.
+This should be possible without forking/modifying Sirius Components, i.e. stricly on the application side.
+
+When running such an application, studio makers should be able to use the additional widgets inside their modelers, whether the modelers are defined:
+* programmatically using the _Form_ API (`org.eclipse.sirius.components.forms.description.FormDescription`)
+* programmatically by creating View-based _Form Description_
+* manually by configuring their View-based _Form Description_ using the Explorer and _Details_ view
+* manually by configuring their View-based _Form Description_ using the visual _FormDescriptionEditor_
+
+It should be possible for a developer to provide a (set of) custom widgets as a reusable library (with backend and frontend parts) and for a studio maker (different person) to use these custom widgets in his own application by adding the appropriate dependencies.
+
+Sirius Web should provide at least one complete and documented example of such a custom widget to serve as a reference for implementers.
+
+== Solution
+
+We will need to make the Form representation implementation extensible at several levels to support widget types which the core implementation does not know about:
+
+1. The core Form representation itself
+2. The GraphQL Schema
+3. The View DSL
+4. The frontend
+
+We will provide a simple "Slider" widget as an example and to validate the feature.
+The slider widget will only support displaying/adjusting an integer value inside a range (min/max).
+The backend part will be provided in new, separate Java projects/Maven modules that only `sirius-web-sample-application` will depend on.
+The frontend part of the Slider widget will be entirely contained in `sirius-components/packages/sirius-web/frontend/sirius-web`.
+It will use [the Slider component from MUI](https://v4.mui.com/components/slider/) for the frontend UI.
+
+[#core]
+=== Form Representation Core
+
+It is already possible to provide custom sub-types of `AbstractWidgetDescription` (e.g. `SliderDescription`), and to add this to a `GroupDescription.controlDescription`.
+However the `FormRenderer` (more specifically, its helpers `FormElementFactory`, `FormInstancePropsValidator` and `FormComponentPropsValidator`) does not know what to do with `AbstractWidgetDescription`s which are not in its hard-coded list.
+
+We will introduce a new Java API named `WidgetDescriptor` which represents a new kind of widget the renderer should know about.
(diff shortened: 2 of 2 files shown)


doc/iterations/2023.6/shapes/custom-widgets.adoc
```diff
@@ -0,0 +1,147 @@
...
+An application based on Sirius Component should be able to extend the _Form_ representation with additional, custom widget definitions.
+This should be possible without forking/modifying Sirius Components, i.e. stricly on the application side.
+
+When running such an application, studio makers should be able to use the additional widgets inside their modelers, whether the modelers are defined:
+* programmatically using the _Form_ API (`org.eclipse.sirius.components.forms.description.FormDescription`)
+* programmatically by creating View-based _Form Description_
+* manually by configuring their View-based _Form Description_ using the Explorer and _Details_ view
+* manually by configuring their View-based _Form Description_ using the visual _FormDescriptionEditor_
+
+It should be possible for a developer to provide a (set of) custom widgets as a reusable library (with backend and frontend parts) and for a studio maker (different person) to use these custom widget in his own application by adding the appropriate dependencies.
```
sbegaudeau: these custom widgets

doc/iterations/2023.6/shapes/custom-widgets.adoc
```diff
@@ -0,0 +1,147 @@
...
+
+Sirius Web should provide at least one complete and documented example of such a custom widget to serve as a reference for implementers.
+
+== Solution
+
+We will need to make the Form representation implementation extensible at several levels support widget types which the core implementation does not know about:
```
sbegaudeau: to support

doc/iterations/2023.6/shapes/custom-widgets.adoc
```diff
@@ -0,0 +1,147 @@
...
+=== Form Representation Core
+
+It is already possible to provide custom sub-types of `AbstractWidgetDescription` (e.g. `SliderDescription`), and to add this to a `GroupDescription.controlDescription`.
+However the `FormRenderer` (more specifically, its helpers `FormElementFactory`, `FormInstancePropsValidator` and `FormComponentPropsValidator`) does not know what to do with `AbstractWidgetDescription`s which are not in its hard-coded list.
+
+We will introduce a new Java API which represents a new kind of widget component.
+`FormRenderer` will take a list of such "widget descriptors" as a parameter, and its helpers will use that to render the new components.
+
+This list of widget descriptors will be passed by `FormEventProcessor`, which will itself get it from `FormEventProcessorFactory`.
+`FormEventProcessorFactory` itself will get these using usual Spring dependency mecahnisms.
```
sbegaudeau: mechanisms

doc/iterations/2023.6/shapes/custom-widgets.adoc
```diff
@@ -0,0 +1,147 @@
...
+The frontend part of the Slider widget will be entirely contained in `sirius-components/packages/sirius-web/frontend/sirius-web`.
+It will use [the Slider component from MUI](https://v4.mui.com/components/slider/) for the frontend UI.
+
+[#core]
+=== Form Representation Core
+
+It is already possible to provide custom sub-types of `AbstractWidgetDescription` (e.g. `SliderDescription`), and to add this to a `GroupDescription.controlDescription`.
+However the `FormRenderer` (more specifically, its helpers `FormElementFactory`, `FormInstancePropsValidator` and `FormComponentPropsValidator`) does not know what to do with `AbstractWidgetDescription`s which are not in its hard-coded list.
+
+We will introduce a new Java API which represents a new kind of widget component.
```
sbegaudeau: the term descriptor may be more relevant than component... since we will need to provide components too.

The following comment may probably appear in an ADR instead of the shape but since your shape is very detailed I'm thinking about this:

I think that, we would also need to change the behavior of the form components. Right now, we have something like this (I'm using JSX on purpose):

```
new FormRenderer().render(<FormComponent />);

const FormComponent = (props) => {
  return (
    <form>
      {props.getFormDescription().getPageDescriptions().map(pageDescription => {
        return <PageComponent pageDescription={pageDescription} />;
      })}
    </form>
  );
}

const PageComponent = (props) => {
  return (
    <page>
      {props.getPageDescription().getGroupDescriptions().map(groupDescription => {
        return <GroupComponent groupDescription={groupDescription} />
      })}
    </page>
  );
}

const GroupComponent = (props) => {
  return (
    <group>
      {props.getGroupDescriptions().getWidgetDescriptions().map(widgetDescription => {
        const element = null;
        if (isTextfield(widgetDescription)) {
          element = <TextfieldComponent />
        } else if (isCheckbox(widgetDescription)) {
          element = <CheckboxComponent />
        }
        // etc for all widgets
        return element;
      })}
    <group>
  );
}
```

In order to change this hardcoded list of components here, we should either:

#### Provide a list of custom component and use it deep down:

```
new FormRenderer().render(
  <FormComponent
    widgetComponents={
       SliderComponent,
       DatePickerComponent,
       ColorPickerComponent
    }
  />
);
```

And then give this list of hardcoded component from `FormComponent` to `PageComponent` and to `GroupComponent` and to `FlexboxContainerComponent` at least.

#### Introduce a concept of `context`

```
new FormRenderer().render(
  <CustomWidgetContext.Provider widgetComponents={
       SliderComponent,
       DatePickerComponent,
       ColorPickerComponent
    }
  >
    <FormComponent />
  <CustomWidgetContext />
);
```

And find the custom widgets in `GroupComponent` and `FlexboxContainerComponent`

There is another option but it's needlessly complicated :)

sbegaudeau: I think the first option is the most efficient for now since context would be a bit overkill even in JSX for such small need but maybe one day... :)

doc/iterations/2023.6/scopes/custom-widgets.adoc
```diff
@@ -0,0 +1,32 @@
+= Scope: Support custom widgets in Form representations
```
sbegaudeau: From my understanding, we were supposed to only work on shapes for the moment and not scopes since they should only live in the PR with the code being developed and merged when the work is done.

doc/iterations/2023.6/shapes/custom-widgets.adoc
sbegaudeau: I don't think that any of this needs to be mentioned here today :)

sbegaudeau (changes requested):

doc/iterations/2023.6/shapes/custom-widgets.adoc
pcdavid: I'll replace the use of "component". The rest of the doc (next sentence) actually talks about "list of widget descriptors" already.

For the rest of your comment, it seems we are roughly on the same page, we'll iron out the details when I start actually implementing this (which should happen relatively soon).

doc/iterations/2023.6/scopes/custom-widgets.adoc
pcdavid: OK

sbegaudeau (approved):
I'll wait for the build to merge it

PR #1956, opened 2023-05-03
 [doc] WIP Contribute the adr of the view creation API

mcharfadi:
[doc] WIP Contribute the adr of the view creation API

### DIFF
--- a/doc/adrs/098_simplify_the_programmatic_creation_of_view_models.adoc
+++ b/doc/adrs/098_simplify_the_programmatic_creation_of_view_models.adoc
@@ -1,0 +1,256 @@
+= ADR-098 - WIP API view creation
+
+== Context
+
+We can create representations programmatically, with the view model and with the compatibility layer that converts odesign files in representation descriptions.
+
+Another use case is to create a view model programmatically but we don't have an API to create complex models easily.
+
+Such an API would allow us to generate view diagram description, edges, nodes and tools based on a metamodel description.
+
+This API would also offer default behaviors for the created view concepts.
+
+These default behaviors can be altered to generate a more specific view dsl.
+
+== Decision
+
+=== Organization of the code creating and linking elements
+
+==== Builder pattern : Creating the elements
+
+The builder pattern allows us to configure a builder with some specific behaviors or values and reuse this configured builder to create several descriptions.
+
+This implies that the API should offer the methods to set default behaviors during the instantiation of the builder.
+
+The builder pattern allows us to chain the setters, implement behaviors other than the default one provided during the builder instantiation, and is easily readable.
+
+We will implement builders for creating : 
+
+* diagramDescription
+* nodeDescription
+* edgeDescription
+* styleDescription
+* toolDecription
+
+==== Provider interfaces : Organization of the code creating and linking elements
+
+The provider interface makes the implementation more explicit and allow us to organize the code.
+
+We will expose these interfaces :
+
+* IDiagramElementDescriptionProvider
+* IEdgeDescriptionProvider
+* INodeDescriptionProvider
+* INodeToolProvider
+* IStyleProvider
+
+The implementation of these interfaces will be in charge of creating elements (preferably with the builders but not necessarily) and linking them together.
+
+For linking elements together we will offer a cache mechanism (see section) (change wording 'cache').
+
+==== Based on String, EMF model, Domain model ?
+
+The current API in papaya retrieves the name of an Entity to help set the DomainType attribute of EdgeDescription and NodeDescription.
+
+Using EClass to instantiate EdgeDescription and NodeDescription will allow the implementation of complex behaviors.
+
+But if we use EClass some model created with Domain would not match the specification and would need to be converted to EMF model.
+
+Since there are many different scenarios, we would probably end up with an API asking for strings (the lowest common denominator) but some utilities could be provided to create those strings from EObjects (see AQL section).
+
+==== Example
+
+```
+public class ClassDiagramAttributeNodeDescriptionProvider implements INodeDescriptionProvider {
+    @Override
+    public NodeDescription create() {
+        // Create a builder with some default settings, we could retrieve such builder from anywhere else
+        var builder = NodeDescriptionBuilder.withDefaults()
+            .withDeleteTool(true)
+            .withDefaultStyle(style)
+            .initialize();
+
+        // Use it to create a node description while changing some values for this specific use case
+        var nodeDescription = builder.name("CD papaya::Attribute")
+            .labelExpression("aql:self.name")
+            .domainType("papaya::Attribute")
+            .synchronizationPolicy(SynchronizationPolicy.UNSYNCHRONIZED)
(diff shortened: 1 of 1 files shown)


doc/adrs/098_simplify_the_programmatic_creation_of_view_models.adoc
```diff
@@ -0,0 +1,208 @@
+= ADR-098 - WIP API view creation
+
+== Context
+
+We can create representations programmatically, with the view model and with the compatibility layer that converts odesign files in representation descriptions.
+
+An other use case is to create view model programmatically but we don't have an API to create complexes models easily.
```
sbegaudeau: complexes -> complex

doc/adrs/098_simplify_the_programmatic_creation_of_view_models.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+=== Based on String, EMF model, Domain model ?
+
+The current API in papaya retrieve the name of an Entity to help set the DomainType attribute of EdgeDescription and NodeDescription.
+
+Using EClass to instantiate EdgeDescription and NodeDescription will allow the implementation of complex behaviors.
+
+But if we use EClass some model created with Domain would not match the specification and would need to be converted to EMF model.
+
+Since there are many different scenarios, we should let the user of the API the choose to implement methods that use EMF or Domain models but not impose this choice.
```
sbegaudeau: So it would probably end up with an API asking for strings (the lowest common denominator) but some utilities could be provided to create those string from EObjects.

doc/adrs/098_simplify_the_programmatic_creation_of_view_models.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+The builder pattern allows us to chain the setters, implement behaviors other than the default one provided during the builder instantiation, and is easily readable.
+```
+NodeDescription b = nodeBuilder.create(EClass eclass)
+  .name('efg')
+  .nameExpression("aql:self.title")
+  .conditionalStyles(conditionalStyles)
+```
+
+The provider pattern makes the implementation more explicit.
```
sbegaudeau: I'm not sure that there is a provider pattern per se and I don't think that they are opposed one to the other. The two pieces of code are solving different issues.

## First issue : Organization of the code creating and linking elements

The first major issue that I have is that I have several pieces of code used to create various parts of the view model. Those pieces of code can be organized in various categories such as:

- code creating diagram descriptions
- code creating node descriptions
- code creating edge descriptions

How this code is doing its job is irrelevant here. All these pieces of code have some common concerns for example, for all those parts I need to consider to major concerns:

- creating elements
- linking elements together

As a result, I need to define the list of the pieces of code responsible for the creation of node descriptions, edge descriptions and diagram descriptions and for each of those pieces of code I need to define what will be created and how things will be linked together.

This is the role of the various interface such as `INodeDescriptionProvider` and `IEdgeDescriptionProvider`.

Once you have define what will be created and how it will be linked, you can move on to the next major topic.

## Second issue : Creating the elements

Here you can argue that a builder based API would be useful to simplify the creation of some elements. You could thus have something like this:

```
public class ClassDiagramAttributeNodeDescriptionProvider implements INodeDescriptionProvider {
    @Override
    public NodeDescription create() {
        // Create a builder with some default settings, we could retrieve such builder from anywhere else
        var builder = NodeDescriptionBuilder.withDefaults()
            .withDeleteTool(true)
            .withDefaultStyle(style)
            .initialize();

        // Use it to create a node description while changing some values for this specific use case
        var nodeDescription = builder.name("CD papaya::Attribute")
            .labelExpression("aql:self.name")
            .domainType("papaya::Attribute")
            .synchronizationPolicy(SynchronizationPolicy.UNSYNCHRONIZED)
            .childrenLayoutStrategy(ViewFactory.eINSTANCE.createListLayoutStrategyDescription())
            .build();

        return nodeDescription;
    }
}
```


## Third issue : Maintaining the consistency of the code

In the example above, I had to mention that my domain type is `EPACKAGE_NAME::ECLASS_NAME` so it would be helpful to have access to something to simplify that for example something like that:

```
        EClass papayaAttributeEClass = ...
        IEcoreViewBuilderHelper helper = ...
        String domainType = helper.domainType(papayaAttributeEClass);

        var nodeDescription = builder.name("CD papaya::Attribute")
            .domainType(domainType)
            .build();
```

In a similar manner, it is important to be able to define your naming convention and apply it easily. Something like this could help:


As you can see, you have here multiple levels of concerns and various steps to consider to move forward.
```
        INodeDescriptionNamingConvention namingConvention = domainType -> {
            return "CD " + domainType;
        };

        var builder = NodeDescriptionBuilder.withDefaults()
            .withNamingConvention(namingConvention)
            .withDeleteTool(true)
            .withDefaultStyle(style)
            .initialize();
```

doc/adrs/098_simplify_the_programmatic_creation_of_view_models.adoc
```diff
@@ -0,0 +1,208 @@
...
+
+We can also instantiate a list of tools and associate it to the palette of the diagramDescription.
+
+In the create() method implementation, we also create the cache used to store all futurs created elements references.
+
+=== DiagramElementDescriptionProvider Provider
+
+```
+public interface IDiagramElementDescriptionProvider<T extends DiagramElementDescription> {
+    EObject create();
```
sbegaudeau, line 201: The result is not parameterized here?

doc/adrs/098_simplify_the_programmatic_creation_of_view_models.adoc
```diff
@@ -0,0 +1,208 @@
...
+```
+
+==== ReconnectionTools
+
+```
+public void addDefaultEdgeCreationTools(EdgeDescription edgeDescription) {}
+public void addDefaultReconnectionTools(EdgeDescription edgeDescription) {}
+```
+
+=== Cache
```
sbegaudeau: Minor detail: this will be a very important topic in order to find stuff while creating the view model. Caching is not really the main concern so the name may not be really relevant.

doc/adrs/098_simplify_the_programmatic_creation_of_view_models.adoc
```diff
@@ -0,0 +1,208 @@
...
+public class PapayaViewCache {
+    private final Map<String, List<EObject>> data = new HashMap<>();
+
+    public void put(EObject eObject) {
+        var eObjects = this.data.getOrDefault(eObject.eClass().getName(), new ArrayList<>());
+        eObjects.add(eObject);
+        this.data.put(eObject.eClass().getName(), eObjects);
+    }
+
+    public NodeDescription getNodeDescription(String name) {
```
sbegaudeau: Sidenote: As you can see here, we need to retrieve elements using some detail. The most usable one being the name of the element (this is why there is a name in the first place). You need to define a naming convention somewhere (because otherwise things cannot be organized properly) and you need to be able to find stuff by their name afterward.

PR #2011, opened 2023-05-31
[1988] Add support for optional help text on widgets

### LINKED ISSUE #1988: Add support for optional help text on widgets
In Sirius Desktop's EEF-based property views, it is possible to associate a help text on widgets that can display additional text in a tooltip.
We should also support this kind of mechanism in our Forms (including in View-based forms).

## Tasks

- [x] [Shape written](https://github.com/eclipse-sirius/sirius-components/pull/1987)
- [x] ADR written
- [x] Core support for `AbstractWidget[Description].helpTextProvider` and `hasHelpText` flag.
- [x] Implement the GraphQL Query to invoke a widget's `helpTextProvider`.
- [x] Frontend: display the "?" hint for widgets with `hasHelpText` (need more discussion on the icon and layout)
- [x] Frontend: invoke the GraphQL Query and display the result (as a tooltip)
- [x] View: add `WidgetDescription.helpTextExpression`
- [x] View: create a proper `helpTextProvider` from the expression during the View conversion.
- [x] Frontend: display the "?" hint in the FormDescriptionEditor version of the widgets (if the `helpTextExpression` is defined)
- [x] Add tests
- [x] Update documentation and CHANGELOG with a screenshot/animation.

pcdavid:
# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


packages/forms/backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/renderer/FormElementFactory.java
```diff
@@ -403,7 +415,12 @@ private Textfield instantiateTextfield(TextfieldElementProps props, List<Object>
         if (props.getStyle() != null) {
             textfieldBuilder.style(props.getStyle());
         }
-
+        if (props.getHelpTextProvider() != null) {
+            textfieldBuilder.helpTextProvider(props.getHelpTextProvider());
+        }
+        if (props.getHelpTextProvider() != null) {
+            textfieldBuilder.helpTextProvider(props.getHelpTextProvider());
+        }
```
frouene: duplicated line

packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/handlers/HelpTextRequestEventHandler.java
```diff
@@ -0,0 +1,92 @@
+/*******************************************************************************
+ * Copyright (c) 2022, 2023 Obeo.
```
frouene: shouldn't it just be 2023

packages/forms/backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/Select.java
```diff
@@ -84,11 +85,12 @@ public static final class Builder {
         private String value;
 
         private Function<String, IStatus> newValueHandler;
-
         private SelectStyle style;
```
frouene: There should be a blank line

frouene:
The helpExpression doesn't work for slider, I think the `helpTextProvider` is missing from the builder of `Slider.java`.

frouene:
The helpExpression is not display for `toolbarAction` widget, but it seems to be the same behavior as for the label.

frouene (changes requested):

packages/forms/backend/sirius-components-forms-graphql/src/main/java/org/eclipse/sirius/components/forms/graphql/datafetchers/form/FormDescriptionHelpTextDataFetcher.java
```diff
@@ -0,0 +1,74 @@
...
+ * }
+ * </pre>
+ *
+ * @author pcdavid
+ */
+@QueryDataFetcher(type = "FormDescription", field = "helpText")
+public class FormDescriptionHelpTextDataFetcher implements IDataFetcherWithFieldCoordinates<CompletableFuture<String>> {
+    private static final String WIDGET_ID_ARGUMENT = "widgetId";
+
+    private final IEditingContextEventProcessorRegistry editingContextEventProcessorRegistry;
```
sbegaudeau: Prefer the `IEditingContextDispatcher`, it helps the integration of third party applications

git all
```diff
@@ -0,0 +1,368 @@
+[1m[0m[38;5;11m──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────[0m
```
sbegaudeau: huh?

packages/forms/backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/Button.java
```diff
@@ -96,6 +96,8 @@ public static final class Builder {
 
         private List<Diagnostic> diagnostics;
 
+        private Supplier<String> helpTextProvider;
```
sbegaudeau, line 99: I suspect this is what you were talking about with the modification of dozens of files with widgets, components, descriptions to set that up. In the diagram representation, we went the other way by removing the description as the sole mean to carry information for a service and instead try to use its content only for the rendering.

You could have introduced something like that to ask for the help text and use it in the handler:

```
public interface IWidgetHelpTextProvider {
    String getHelpText(...)
}
```

When the frontend would ask for the help text, you would just ask this service which could then leverage the view model to find the relevant information. Here, the converters needs to setup everything first. Which is why you have to make the help text provider optional because it needs to carry two pieces of informations at once:

- is there some help text? (for the rendering)
- what is the value of the help text?

You could have simply used a boolean to carry the information relevant for the rendering.

We could also introduce a common `AbstractWidgetBuilder` which would be inherited by all the Builders of the widgets. It could encapsulate things like the id, presence of help text, diagnostics, label. I don't see any issues with having some simple inheritance for builders like that since builders are 99% data structure.

sbegaudeau (dismissed):
I have not tested it but what is that `git all` file exactly?

git all
pcdavid: > I have not tested it but what is that `git all` file exactly?

Just a cli glitch/mistake. Removed.

packages/forms/backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/Button.java
pcdavid: > You could have simply used a boolean to carry the information relevant for the rendering.

Yes, but I would still have needed about the same amount of changes to carry the boolean (instead of the `Supplier<String>`), wouldn't I?

> I don't see any issues with having some simple inheritance for builders like that since builders are 99% data structure.

AFAICT, it's forbidden by the rules you setup: `immutableClassesShouldHaveANestedBuilder`, and `HaveAValidBuilderCondition` requires the builder class to be `final`.

packages/forms/backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/renderer/FormElementFactory.java
pcdavid: done

packages/forms/backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/Button.java
pcdavid: > When the frontend would ask for the help text, you would just ask this service which could then leverage the view model

What about widgets created programatically which do not come from a View model?

packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/handlers/HelpTextRequestEventHandler.java
pcdavid: done

packages/forms/backend/sirius-components-forms/src/main/java/org/eclipse/sirius/components/forms/Select.java
pcdavid: done.

packages/forms/backend/sirius-components-forms-graphql/src/main/java/org/eclipse/sirius/components/forms/graphql/datafetchers/form/FormDescriptionHelpTextDataFetcher.java
pcdavid: Done. I'll push a separate PR to do the same cleanup in the CompletionProposals data fetcher, but there may be others.

pcdavid:
> The helpExpression doesn't work for slider, I think the `helpTextProvider` is missing from the builder of `Slider.java`.

There was a little more missing, but done.

pcdavid:
> The helpExpression is not display for `toolbarAction` widget, but it seems to be the same behavior as for the label.

I've added help support for toolbarActions. The tooltip is associated to the button itself instead of a separate "?" icon (there is not really any room for such an icon in this case).

packages/forms/frontend/sirius-components-forms/src/propertysections/HelpTooltip.types.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+export interface HelpTooltipProps {
+  editingContextId: string;
+  formId: string;
+  widgetId: string;
+  children?: any;
```
sbegaudeau: React.ReactElement would work here

packages/forms/frontend/sirius-components-forms/src/propertysections/HelpTooltip.tsx
```diff
@@ -0,0 +1,96 @@
...
+          }
+        }
+      }
+    }
+  }
+`;
+
+export function HelpTooltip(props: HelpTooltipProps) {
+  const [open, setOpen] = useState(false);
+  const [content, setContent] = useState<JSX.Element>(null);
```
sbegaudeau: You could consolidate both state into one.

packages/forms/frontend/sirius-components-forms/src/propertysections/HelpTooltip.tsx
```diff
@@ -0,0 +1,96 @@
...
+        } else {
+          const lines = text.split('\n');
+          setContent(
+            <div>
+              {lines.map((line) => (
+                <>
+                  {line}
+                  <br />
+                </>
+              ))}
```
sbegaudeau: Use `React.Fragment` to have access to the `key` props to remove the warning

packages/forms/frontend/sirius-components-forms/src/propertysections/HelpTooltip.tsx
```diff
@@ -0,0 +1,96 @@
...
+    <Tooltip
+      open={open && content !== null}
+      onClose={handleClose}
+      onOpen={handleOpen}
+      title={content || ''}
+      arrow
+      placement={'top'}>
+      {props.children || <HelpOutlineOutlined color="secondary" style={{ marginLeft: 8, fontSize: 16 }} />}
+    </Tooltip>
+  );
```
sbegaudeau: I'm still not sure why but moving the cursor a tiny bit can trigger a lot of HTTP request to reload the tooltip. I wonder if there is some options in MUI which would prevent so many open/close events

sbegaudeau (approved):

PR #2127, opened 2023-06-29
[doc] Contribute a shape for the filtering/ordering pages API

frouene:
Shape for https://github.com/eclipse-sirius/sirius-components/issues/2125.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -8,6 +8,7 @@
 - Add new custom widget to edit single and multi-valued references
 - Add support for Help Expressions in Form widgets
 - Add border style for containers
+- Add an API for filtering and ordering form pages
 
 === Architectural decision records
 
@@ -79,7 +80,8 @@ The help text can include multiple lines (separated by `\n`), but no text format
 +
 image:doc/images/Widget_Help_Tooltip.png[Example of a help tooltip on a widget]
 +
-- https://github.com/eclipse-sirius/sirius-components/issues/2048[#2048] [diagram] Add a basic support for the resize. It is possible to reduce the size of a node less than the space needed to display all children.
+- https://github.com/eclipse-sirius/sirius-components/issues/2048[#2048] [diagram] Add a basic support for the resize.
+It is possible to reduce the size of a node less than the space needed to display all children.
 - https://github.com/eclipse-sirius/sirius-components/issues/2064[#2064] [forms] Make the `IWidgetDescriptor` API more flexible.
 - https://github.com/eclipse-sirius/sirius-components/issues/1618[#1618] [view] Split the view metamodel into dedicated subpackages.
 - https://github.com/eclipse-sirius/sirius-components/issues/2083[#2083] [diagram] Add support for edge markers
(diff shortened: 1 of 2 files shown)


doc/iterations/2023.8/shapes/order_and_filter_from_pages.adoc
```diff
@@ -0,0 +1,40 @@
+= (S) Add an API for filtering and ordering view properties pages
+
+== Problem
+
+A form in view properties can be composed of several pages from various contributions.
+A specifier should be able to order and filter these pages to control the form representation.
+
+== Key Result
+
+Offer the specifier the ability to programmatically order and filter the pages of a form.
```
AxelRICHARD: Isn't it already possible with the current version of the tool?
Pages have domain type and precondition expression, so it is already possible to filter the pages.
Pages can be moved with the FormDescriptionEditor so it is already possible to order the pages.

pcdavid: I think the idea is for *applications* (which did not necessarily define the Forms definitions and this can not change their preconditions) to control "after the fact" which pages they want to see and in which order.

pcdavid: If my interpretation is correct, "A specifier should be able to order and filter these pages to control the form representation." should probably be reformulated. Indeed, if we assume this will be used by the specifier (of the form), it is redundant with the existing mechanism.
The proposed API makes sense however for applications which reuse existing Forms but want some control (maybe exposed to their end-users) on the Pages to display.

doc/iterations/2023.8/shapes/order_and_filter_from_pages.adoc
```diff
@@ -0,0 +1,40 @@
...
+== Solution
+
+Add a new API with in entry the list of the pageDescription which can be applied on an item.
+This API should return the list of the pages to display in the form.
+
+[source,java]
+----
+public interface IPropertiesPageDescriptionPostProcessFiltering {
+    List<String> postProcessFiltering(List<PageDescription> pageDescriptions, PropertiesConfiguration propertiesConfiguration, IEditingContext editingContext);
+}
```
sbegaudeau: Why use the `PropertiesConfiguration`? It's a bit odd to rely on this object since it represents one subscription to a representation. While technically today, it may be the same object for all those who subscribe to the Details view, it's still conceptually very odd.

Something like this would make more sense:

```
public interface IPageDescriptionProcessor {
    List<PageDescription> process(List<PageDescription> pageDescriptions, VariableManager variableManager);
}
```

or even more powerful :

```
public interface IFormProcessor {
    Form process(Form form, VariableManager variableManager);
}
```

This way, you have a simple API which let application maintainer do anything they want with forms such as:

- Change the appearance of the form using external constraints (authentication, authorizations, other criteria)
- Filter / hide anything in the form anywhere

I don't want an API to only let maintainers change the order of page descriptions or remove page descriptions to later have to create another dedicated API to let them change another tiny detail. One API to change what they want after we are done with our job is better and from years of experience with EEF on desktop, they will want to customize EVERYTHING.

CHANGELOG.adoc
```diff
@@ -7,6 +7,7 @@
 - Add disable / enable concept for widgets
 - Add new custom widget to edit single and multi-valued references
 - Add support for Help Expressions in Form widgets
+- Add an API for filtering and ordering view properties pages
```
sbegaudeau: I'm not sold on an API that would only work for the details view. I know that this exact requirement will arrive later for the form representation or any other form based view and we definitely won't maintain two different APIs for that

sbegaudeau (dismissed):

doc/iterations/2023.8/shapes/order_and_filter_from_pages.adoc
frouene: I changed the shape for the second option with the `IFormProcessor`

CHANGELOG.adoc
frouene: I made it more generic, to all form pages.

doc/iterations/2023.8/shapes/order_and_filter_from_pages.adoc
```diff
@@ -0,0 +1,38 @@
+= (S) Add an API for filtering and ordering form pages
+
+== Problem
+
+A form can be composed of several pages from various contributions.
+A specifier should be able to order and filter these pages to control the form representation.
```
pcdavid: I still believe this should not talk about *specifiers*, but about developers and/or application integrators, who do not define/control the way the various `FormDescriptions` are defined, but want to control how the result appears to their users in their specific application.

pcdavid (changes requested):

doc/iterations/2023.8/shapes/order_and_filter_from_pages.adoc
frouene: Yes, you're right, I changed all the specifier references.

pcdavid (approved):

PR #2148, opened 2023-07-03
[2126] Add support for multiple TreeDescription

### LINKED ISSUE #2126: Add support form multiple different "explorer-like" tree representations
In the `sirius-components-collaborative-trees` most of the code is generic, but `TreeEventProcessorFactory` is hard-coded to support a single `TreeDescription` source, namely `IExplorerDescriptionProvider`.

For the various modals we will need for the reference widget (see #2116, #2117, #2118) we will need tree representations which use a different `TreeDescription` definition.

`TreeEventProcessorFactory` and the related code (both back and front) should be made more configurable to support for multiple `TreeDescription` definitions, so that the modals can use/subscribe to tree definitions adapted to their case.

pcdavid:


pcdavid:
This reuses `TreeDescription.getCanCreatePredicate()`, to find the correct `TreeDescription` implementation to use for a given subscription, based on the treeId.
It's a little different from how `getCanCreatePredicate()` is used for more "normal" representations like diagrams, but it was already the case before that this "normal meaning" did not fit well with the explorer or for the `SelectionDescription`, which are not meant to be created by the end-user but programatically.

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/explorer/ExplorerDescriptionProvider.java
```diff
@@ -81,12 +83,19 @@ public ExplorerDescriptionProvider(IObjectService objectService, IURLParser urlP
...
-    public TreeDescription getDescription() {
-        // This predicate will NOT be used while creating the explorer but we don't want to see the description of the
-        // explorer in the list of representations that can be created. Thus, we will return false all the time.
-        Predicate<VariableManager> canCreatePredicate = variableManager -> false;
+    public void addRepresentationDescriptions(IRepresentationDescriptionRegistry registry) {
+        registry.add(this.getExplorerDescription());
+    }
+
+    public TreeDescription getExplorerDescription() {
+        // This predicate will be called when the used tries to create a new representation, but in this case the
```
AxelRICHARD: used -> user

packages/trees/frontend/sirius-components-trees/src/views/ExplorerView.tsx
```diff
@@ -84,7 +85,14 @@ const useExplorerViewStyles = makeStyles((theme) => ({
   },
 }));
 
-export const ExplorerView = ({ editingContextId, selection, setSelection, readOnly }: WorkbenchViewComponentProps) => {
+export const ExplorerView = ({
```
AxelRICHARD: Shouldn't it be split in two different components, `TreeView` and `ExplorerView`?

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/explorer/ExplorerDescriptionProvider.java
pcdavid: done

packages/trees/frontend/sirius-components-trees/src/views/ExplorerView.tsx
pcdavid: done.

AxelRICHARD (approved):

PR #2206, opened 2023-07-25
[doc] Shape for color management in View

AxelRICHARD:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/iterations/2023.10/shapes/color_management_in_view.adoc
+++ b/doc/iterations/2023.10/shapes/color_management_in_view.adoc
@@ -1,0 +1,82 @@
+= (M) Color Management in View
+
+== Problems
+
+It is possible to set `None` for color properties of a Node/Edge Description.
+Depending on the case, `None` means `black`, `white` or `transparent`.
+
+It is possible to set `None` for color properties of a Form Widget Description.
+Depending on the case, `None` means `theme.primary`, `theme.secondary` or `transparent`.
+
+Some color properties of a Node/Edge/Widget Description are mandatory, which means when `None` is selected, an error message is displayed.
+That's ok but we should review which color property is really mandatory or not.
+
+== Key results
+
+* Specifier always know which color will be displayed at the end, corresponding exactly to what is set for each color property.
+* A set of predefined colors is accessible in color properties.
+
+== Solutions
+
+`None` doesn't exist anymore as a possible value for color properties.
+
+All color properties are mandatory.
+
+Theme colors are provided by default.
+They are accessible in the Reference widget of color properties by default.
+Color palettes are provided by default (along with the creation of a View, or directly provided in the ResourceSet).
+Those color palettes and their colors cannot be modified/deleted.
+
+
+Those color palettes will contain the following colors:
+
+* special colors like "transparent", "inherit", "none" and use some of them as default values when we are creating new elements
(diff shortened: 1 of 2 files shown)


doc/iterations/2023.10/shapes/color_management_in_view.adoc
```diff
@@ -0,0 +1,106 @@
...
+* Specifier always know which color will be displayed at the end, corresponding exactly to what is set for each color property.
+* A set of predefined colors is accessible in color properties.
+
+== Solutions
+
+1) Preferred solution
+
+`None` doesn't exist anymore as a possible value for color properties.
+
+All color properties are mandatory.
```
sbegaudeau, line 23: I would go for this solution over any other for one simple fact: It stops us from setting magic values everywhere. The specifier has to give us a proper value.

doc/iterations/2023.10/shapes/color_management_in_view.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+Theme colors are provided by default (need to choose between):
+
+* A) the theme colors are accessible in the Select widget of color properties by default.
+No need to declare them in a Color Palette, they are accessible everywhere.
+* B) the theme colors are accessible in the Select widget of color properties by default, and they come from a color palette created along with the View and this color palette cannot be modified/deleted.
```
sbegaudeau: I would go for the second option since it would be way easier. Otherwise, we would have to create some additional model with default values and then handle it as a special use case at tons of different locations. For example, we wouldn't want it to be visible or modifiable in the explorer, the details view, or dialogs, etc. This would create tons of special use cases to handle today even if I believe that some default library will one day be the best way to provide several key user experience improvements.

I would move to this direction in two steps.

- We would also need some special colors like "transparent", "inherit", "none" and use some of them as default values when we are creating new elements.
- We should probably provide every single one of the raw MUI colors like [here](https://mui.com/material-ui/customization/color/#color-palette) in various palette (red, pink, blue, etc) with their proper name
- We may also need some [special colors](https://mui.com/material-ui/customization/palette/) like `palette.primary.light`, `palette.secondary.dark` etc to create theme independent representations (for example a for with buttons matching the `palette.error.main` color to perform destructive actions)

This way, we would have a couple of default palette with special CSS Colors (transparent, none, inherit), raw MUI colors (blue 400, grey 800), special theme colors (palette.primary.light). We would need some special treatment for those special theme colors in order to interpret them as JS values instead of CSS colors but that's quite easy.

doc/iterations/2023.10/shapes/color_management_in_view.adoc
mbats: +1

doc/iterations/2023.10/shapes/color_management_in_view.adoc
mbats: +1

sbegaudeau (approved):

PR #2227, opened 2023-08-04
[fix] Fix potential NPE on missing DiagramDescription.palette

pcdavid:
It's pretty easy to create a DiagramDescription without a palette:
* from the "View" onboard area action (invokes `EditingContextActionHandler.createEmptyViewResource()`)
* from the "New model" explorer action (invokes `StereotypeDescriptionRegistryConfigurer.getEmptyViewContent()`)

Both locations could be fixed/improved, but it will always be possible to end up in such a state, so the View conversion should not crash in this case.


gcoutable (approved):
Can be merged after the build has finished

PR #2324, opened 2023-08-31
[enh] Preview colors when configuring a node style

pcdavid:
https://github.com/eclipse-sirius/sirius-web/assets/10608/5d31b4d8-69b6-4872-9a94-7c095bede5af

What's missing:
- a proper ticket and changelog entry;
- this only works for the very specific case of Node styles, where we override the Details definition with `NodeStylePropertiesConfigurer`. Colors on edge styles or in forms are not impacted;
- the actual color *definition* in the explorer is not impacted either;
- there is zero validation/sanitation on the raw color value we insert into the generated SVG. could defining a color as `blue'/> INJECT ANYTHING HERE <rect style='fill:'` has bad side effects?


pcdavid:
Updated version to work with the reference widget:

Capture d’écran du 2023-09-06 12-01-51

PR #2337, opened 2023-09-04
[2254] Add precondition on diagram tools

### LINKED ISSUE #2254: Add support for preconditions on diagram tools
Preconditions are supported on the odesign compatibility layer but not for the view DSL

frouene:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2254

### DIFF
--- a/packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/configuration/ViewPaletteToolsConfiguration.java
+++ b/packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/configuration/ViewPaletteToolsConfiguration.java
@@ -26,7 +26,7 @@ import org.springframework.stereotype.Service;
  * @author frouene
  */
 @Service
-public class ViewReconnectionToolsExecutorParameters {
+public class ViewPaletteToolsConfiguration {
 
     private final IURLParser urlParser;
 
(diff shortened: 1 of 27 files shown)


packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/configuration/ViewPaletteToolsConfiguration.java
```diff
@@ -18,15 +18,15 @@
...
+import org.springframework.context.annotation.Configuration;
 
 /**
  * Bundles the common dependencies that view tool services need into a single object for convenience.
  *
  * @author frouene
  */
-@Service
-public class ViewReconnectionToolsExecutorParameters {
+@Configuration
```
pcdavid: Not sure why this change.
`@Configuration`'s Javadoc says:

> Indicates that a class declares one or more `@Bean` methods and may be processed by the Spring container to generate bean definitions and service requests for those beans at runtime

But this does not define any `@Bean`, it just bundles existing ones into a single _parameter object_.

pcdavid (approved):

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/configuration/ViewPaletteToolsConfiguration.java
frouene: I thought the other classes that bundles parameters were declared as Configuration, but no, they're actually Services, and it's in their names that we talk about configuration.
I changed it

frouene: Done

PR #2394, opened 2023-09-18
[2386] Reference widget candidates lazy loading

### LINKED ISSUE #2386: Reference widget slow down form renderer when the candidate lists is too big
With large model, the reference widget candidate list can be very large. 
If the form contains several reference widget, each renderer may apply a vast amount of data and slow down significantly the application.

The solution will be to change the reference widget behavior to only retrieve candidats when needed (list opening and selection modal opening)

frouene:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2386

# Pull request template

## General purpose
- [x] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


pcdavid:
There's a typo in the commit message: "laoding" => "loading".

packages/forms/backend/sirius-components-widget-reference/src/main/java/org/eclipse/sirius/components/widget/reference/ReferenceWidgetComponent.java
```diff
@@ -114,7 +114,7 @@ public Element render() {
                 .label(label)
                 .iconURL(iconURL)
                 .values(items)
-                .options(options)
+                .optionsProvider(() -> options)
```
pcdavid: `options` is still computed on render, we are just deferring the transfer of the values to the front on demand. Are we sure this is enough to fix the performance issues?

pcdavid: If we just do it this way, I think we could simply keep the ReferenceWidgetComponent & widget as before, just not declare the `referenceOptions: [ReferenceValue!]!` in the GraphQL schema so that the front is required to use the separate query if/when it needs them.

pcdavid (changes requested):

packages/forms/backend/sirius-components-widget-reference/src/main/java/org/eclipse/sirius/components/widget/reference/ReferenceWidgetComponent.java
frouene: Yes, you're right, I tested it with a big model and there is some latency on each render.
I'll change it to compute the options inside the supplier.

frouene: Done

pcdavid (approved):
Looks OK, but I have not actually tested.

packages/forms/backend/sirius-components-widget-reference/src/main/resources/schema/reference.graphqls
```diff
@@ -1,3 +1,7 @@
+extend type EditingContext {
+  referenceValueOptions(representationId: ID!, referenceWidgetId: ID!): [ReferenceValue!]!
```
sbegaudeau: Do not place your fields like that directly on EditingContext because everyone will start using this global object as the go to place to put everything. Place that on the description like the palette of the diagram, this way it stays within the domain of the form representation.

It would also remove the need to ask for a representation id while there is already an API to go this way. Look at how the diagram tools are retrieved in the GraphQL API.

packages/forms/frontend/sirius-components-widget-reference/src/components/ValuedReferenceAutocomplete.tsx
```diff
@@ -72,6 +98,45 @@ export const ValuedReferenceAutocomplete = ({
   };
   const classes = useStyles(props);
 
+  const { addErrorMessage } = useMultiToast();
+  const [open, setOpen] = useState<boolean>(false);
+  const [options, setOptions] = useState<GQLReferenceValue[] | null>(null);
```
sbegaudeau: Use one state to help preventing invalid combined states

packages/forms/frontend/sirius-components-widget-reference/src/components/ValuedReferenceAutocomplete.tsx
```diff
@@ -115,7 +186,11 @@ export const ValuedReferenceAutocomplete = ({
       multiple
       filterSelectedOptions
       disabled={readOnly || widget.readOnly}
-      options={widget.referenceOptions}
+      open={open}
+      onOpen={() => setOpen(true)}
```
sbegaudeau: Why aren"t you calling "getReferenceValueOptions" in this function? You are setting an "open" state which will trigger a new rendering which will change the value of loading which will be detected by useEffect to perform the call to the function.

sbegaudeau (changes requested):

packages/forms/backend/sirius-components-widget-reference/src/main/resources/schema/reference.graphqls
frouene: Done

packages/forms/frontend/sirius-components-widget-reference/src/components/ValuedReferenceAutocomplete.tsx
frouene: Done

packages/forms/frontend/sirius-components-widget-reference/src/components/ValuedReferenceAutocomplete.tsx
frouene: I need a state `open` in the case where there is no option. I need to know that the component is opened and I can't only check the result of `getReferenceValueOptions`.

sbegaudeau (approved):

PR #2411, opened 2023-09-20
[releng] Switch to lexical 0.15.0

pcdavid:
Just did a quick manual tests, it seems to work as before.
But more verifications are needed to actually merge this.


vscode-extension/package.json
```diff
@@ -140,7 +140,7 @@
     "@material-ui/icons": "4.11.3",
     "@material-ui/lab": "4.0.0-alpha.61",
     "@xstate/react": "1.6.3",
-    "lexical": "0.8.1",
+    "lexical": "0.15.0",
```
sbegaudeau, line 143: Isn't this file missing an explicit dependency to `@lexical/react`? Shouldn't there be an update of the package.json of `sirius-web` and `sirius-web-application`?

sbegaudeau (changes requested):

pcdavid:
Obsolete, see https://github.com/eclipse-sirius/sirius-web/pull/6321 instead.

PR #2474, opened 2023-10-17
[2443] Use dynamic handles

### LINKED ISSUE #2443: Make handle appear only conditionally
By conditionally, I mean something like on selection or on hover, I don't know, it will have to be defined

mcharfadi:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2443

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -94,6 +94,9 @@ An example of use has been added on the domain entity items.
 - https://github.com/eclipse-sirius/sirius-web/issues/2483[#2483] [diagram] Take the header into account in the React Flow layout.
 - https://github.com/eclipse-sirius/sirius-web/issues/2504[#2504] [diagram] Create nodes at the cursor position when using create instance or drop tools.
 - https://github.com/eclipse-sirius/sirius-web/issues/2444[#2444] [form] It is now possible to ignore widgets with invalid descriptions
+- https://github.com/eclipse-sirius/sirius-web/issues/2443[#2443] [diagram] Add dedicated handles to create edges.
+Hide handles between edges and each edges are now placed in differents positions around the node.
+Allow the creation of an edge by dropping the connection line on a node.
 
 == v2023.10.0
 
--- a/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/converter/convertDiagram.ts
+++ b/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/converter/convertDiagram.ts
@@ -23,7 +23,14 @@ import {
   GQLRectangularNodeStyle,
   GQLViewModifier,
 } from '../graphql/subscription/nodeFragment.types';
-import { BorderNodePositon, ConnectionHandle, Diagram, Label, NodeData } from '../renderer/DiagramRenderer.types';
+import {
+  BorderNodePositon,
+  ConnectionHandle,
+  Diagram,
+  EdgeData,
+  Label,
+  NodeData,
+} from '../renderer/DiagramRenderer.types';
 import { MultiLabelEdgeData } from '../renderer/edge/MultiLabelEdge.types';
 import { IconLabelNodeData } from '../renderer/node/IconsLabelNode.types';
 import { ImageNodeData } from '../renderer/node/ImageNode.types';
@@ -53,23 +60,55 @@ const toRectangularNode = (
   } = gqlNode;
 
   const connectionHandles: ConnectionHandle[] = [];
+  const numberSourceHandles = (handles: ConnectionHandle[]): string => {
+    return handles.filter((handle) => handle.type === 'source').length.toString();
+  };
+  const numberTargetHandles = (handles: ConnectionHandle[]): string => {
+    return handles.filter((handle) => handle.type === 'target').length.toString();
+  };
+  console.log(insideLabel?.text);
   gqlEdges.forEach((edge) => {
-    if (edge.sourceId === gqlNode.id)
+    if (edge.sourceId === gqlNode.id) {
+      console.log(numberSourceHandles(connectionHandles));
+      console.log(`handle--source--${gqlNode.id}--${numberSourceHandles(connectionHandles)}`);
       connectionHandles.push({
-        id: `handle--source--${edge.id}`,
+        id: `handle--source--${gqlNode.id}--${numberSourceHandles(connectionHandles)}`,
         edgeId: edge.id,
         nodeId: gqlNode.id,
         position: Position.Right,
         type: 'source',
       });
-    if (edge.targetId === gqlNode.id)
+    }
+
+    if (edge.targetId === gqlNode.id) {
+      console.log(numberTargetHandles(connectionHandles));
+      console.log(`handle--target--${gqlNode.id}--${numberTargetHandles(connectionHandles)}`);
       connectionHandles.push({
-        id: `handle--target--${edge.id}`,
+        id: `handle--target--${gqlNode.id}--${numberTargetHandles(connectionHandles)}`,
         edgeId: edge.id,
         nodeId: gqlNode.id,
         position: Position.Left,
         type: 'target',
       });
+    }
+  });
+  console.log(numberSourceHandles(connectionHandles));
+  console.log(`Extra handle--source--${gqlNode.id}--${numberSourceHandles(connectionHandles)}`);
+  connectionHandles.push({
+    id: `handle--source--${gqlNode.id}--${numberSourceHandles(connectionHandles)}`,
+    edgeId: '',
+    nodeId: gqlNode.id,
+    position: Position.Right,
+    type: 'source',
+  });
+  console.log(numberTargetHandles(connectionHandles));
+  console.log(`Extra handle--target--${gqlNode.id}--${numberTargetHandles(connectionHandles)}`);
+  connectionHandles.push({
+    id: `handle--target--${gqlNode.id}--${numberTargetHandles(connectionHandles)}`,
+    edgeId: '',
+    nodeId: gqlNode.id,
+    position: Position.Left,
+    type: 'target',
   });
   const data: RectangularNodeData = {
     targetObjectId,
@@ -530,10 +569,10 @@ export const convertDiagram = (gqlDiagram: GQLDiagram): Diagram => {
   const nodeId2Depth = new Map<string, number>();
   nodes.forEach((node) => nodeId2Depth.set(node.id, nodeDepth(nodeId2node, node.id)));
 
-  const edges: Edge[] = gqlDiagram.edges.map((gqlEdge) => {
-    const sourceNode = nodeId2node.get(gqlEdge.sourceId);
-    const targetNode = nodeId2node.get(gqlEdge.targetId);
-
+  let usedHandles: string[] = [];
+  const edges: Edge<EdgeData>[] = gqlDiagram.edges.map((gqlEdge) => {
+    const sourceNode: Node<NodeData> | undefined = nodeId2node.get(gqlEdge.sourceId);
+    const targetNode: Node<NodeData> | undefined = nodeId2node.get(gqlEdge.targetId);
     const data: MultiLabelEdgeData = {
       targetObjectId: gqlEdge.targetObjectId,
       targetObjectKind: gqlEdge.targetObjectKind,
@@ -552,6 +591,17 @@ export const convertDiagram = (gqlDiagram: GQLDiagram): Diagram => {
       data.endLabel = convertEdgeLabel(gqlEdge.endLabel);
     }
 
+    const sourceHandle = sourceNode?.data.connectionHandles
+      .filter((handle) => handle.type === 'source')
+      .find((handle) => !usedHandles.find((usedHandles) => usedHandles === handle.id));
+
+    const targetHandle = targetNode?.data.connectionHandles
+      .filter((handle) => handle.type === 'target')
+      .find((handle) => !usedHandles.find((usedHandles) => usedHandles === handle.id));
+    if (sourceHandle?.id && targetHandle?.id) {
+      usedHandles.push(sourceHandle?.id, targetHandle.id);
+    }
+
     return {
       id: gqlEdge.id,
       type: 'multiLabelEdge',
@@ -566,14 +616,14 @@ export const convertDiagram = (gqlDiagram: GQLDiagram): Diagram => {
       },
       data,
       hidden: gqlEdge.state === GQLViewModifier.Hidden,
-      sourceHandle: `handle--source--${gqlEdge.id}`,
-      targetHandle: `handle--target--${gqlEdge.id}`,
+      sourceHandle: sourceHandle?.id,
+      targetHandle: targetHandle?.id,
       sourceNode: sourceNode,
       targetNode: targetNode,
       interactionWidth: 0.25,
     };
   });
-
+  //id: `handle--target--${gqlNode.id}--${index}`,
   return {
     metadata: {
       id: gqlDiagram.id,
--- a/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/converter/convertDiagram.ts
+++ b/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/converter/convertDiagram.ts
@@ -11,8 +11,9 @@
  *     Obeo - initial API and implementation
  *******************************************************************************/
 
-import { Edge, Node, XYPosition } from 'reactflow';
+import { Edge, Node, Position, XYPosition } from 'reactflow';
 import { GQLDiagram } from '../graphql/subscription/diagramFragment.types';
+import { GQLEdge } from '../graphql/subscription/edgeFragment.types';
 import { GQLLabel, GQLLabelStyle } from '../graphql/subscription/labelFragment.types';
 import {
   GQLIconLabelNodeStyle,
@@ -22,7 +23,7 @@ import {
   GQLRectangularNodeStyle,
   GQLViewModifier,
 } from '../graphql/subscription/nodeFragment.types';
-import { BorderNodePositon, Diagram, Label, NodeData } from '../renderer/DiagramRenderer.types';
+import { BorderNodePositon, ConnectionHandle, Diagram, Label, NodeData } from '../renderer/DiagramRenderer.types';
 import { MultiLabelEdgeData } from '../renderer/edge/MultiLabelEdge.types';
 import { IconLabelNodeData } from '../renderer/node/IconsLabelNode.types';
 import { ImageNodeData } from '../renderer/node/ImageNode.types';
@@ -36,7 +37,8 @@ const defaultPosition: XYPosition = { x: 0, y: 0 };
 const toRectangularNode = (
   gqlNode: GQLNode<GQLRectangularNodeStyle>,
   gqlParentNode: GQLNode<GQLNodeStyle> | null,
-  isBorderNode: boolean
+  isBorderNode: boolean,
+  gqlEdges: GQLEdge[]
 ): Node<RectangularNodeData> => {
   const {
     targetObjectId,
@@ -50,6 +52,25 @@ const toRectangularNode = (
     labelEditable,
   } = gqlNode;
 
+  const connectionHandles: ConnectionHandle[] = [];
+  gqlEdges.forEach((edge) => {
+    if (edge.sourceId === gqlNode.id)
+      connectionHandles.push({
+        id: `handle--source--${edge.id}`,
+        edgeId: edge.id,
+        nodeId: gqlNode.id,
+        position: Position.Right,
+        type: 'source',
+      });
+    if (edge.targetId === gqlNode.id)
+      connectionHandles.push({
+        id: `handle--target--${edge.id}`,
+        edgeId: edge.id,
+        nodeId: gqlNode.id,
+        position: Position.Left,
+        type: 'target',
+      });
+  });
   const data: RectangularNodeData = {
     targetObjectId,
     targetObjectLabel,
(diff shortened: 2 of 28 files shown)


CHANGELOG.adoc
```diff
@@ -57,6 +57,7 @@ This coupling will be removed in the near future as a consequence, the dependenc
 - https://github.com/eclipse-sirius/sirius-web/issues/2410[#2410] [diagram] Do a fit to screen on a diagram opening
 - https://github.com/eclipse-sirius/sirius-web/issues/2268[#2268] [diagram] Create child nodes where the user has clicked.
 - https://github.com/eclipse-sirius/sirius-web/issues/2475[#2475] [diagram] Prevent nodes to move each diagram refresh event. Only the size is still updated
+- https://github.com/eclipse-sirius/sirius-web/issues/2443[#2443] [diagram] Use dynamic handles.
```
pcdavid: A little short for such a visible change. I'm not sure anyone outside of the dev team can know what this is about.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/RectangularNode.tsx
```diff
@@ -40,25 +46,35 @@ const rectangularNodeStyle = (
   if (selected) {
     rectangularNodeStyle.outline = `${theme.palette.primary.main} solid 1px`;
   }
+
+  if (hoveredNode && !selected) {
+    rectangularNodeStyle.outline = `${theme.palette.primary.main} dashed 0.5px`;
+  }
+
```
pcdavid: Not sure how this is related to connection handles? And why only for Rectangular node?

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/RectangularNode.tsx
```diff
@@ -40,25 +46,35 @@ const rectangularNodeStyle = (
...
+  const { newConnectionStyleProvider } = useConnector();
+  const { hoveredNode } = useContext<NodeContextValue>(NodeContext);
 
   const handleOnDrop = (event: React.DragEvent) => {
     onDrop(event, id);
   };
 
+  const isHoveredNode = () => {
+    return !!hoveredNode && hoveredNode.id === id;
+  };
```
pcdavid: Any reason to define this as a function instead of simply like this?

```java
const isHoveredNode = hoveredNode?.id === id;
```

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/RectangularNode.tsx
```diff
@@ -67,38 +83,9 @@ export const RectangularNode = memo(({ data, isConnectable, id, selected }: Node
...
-          id={`handle--${id}--bottom`}
-          type="source"
-          position={Position.Bottom}
-          isConnectable={isConnectable}
-          style={newConnectionStyleProvider.getHandleStyle(data.descriptionId)}
-          onMouseDown={onConnectionStartElementClick}
-        />
+        {selected ? <ConnectionSourceHandles nodeId={id}></ConnectionSourceHandles> : null}
+        <ConnectionTargetHandles nodeId={id}></ConnectionTargetHandles>
+        <ConnectionHandles nodeId={id} customHandles={data.customHandles}></ConnectionHandles>
```
pcdavid: Simpler syntax for empty elements:

```
{selected ? <ConnectionSourceHandles nodeId={id} /> : null}
<ConnectionTargetHandles nodeId={id} />
<ConnectionHandles nodeId={id} customHandles={data.customHandles} />
```

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ListNode.tsx
```diff
@@ -70,38 +73,9 @@ export const ListNode = memo(({ data, isConnectable, id, selected }: NodeProps<L
...
-          id={`handle--${id}--bottom`}
-          type="source"
-          position={Position.Bottom}
-          isConnectable={isConnectable}
-          style={newConnectionStyleProvider.getHandleStyle(data.descriptionId)}
-          onMouseDown={onConnectionStartElementClick}
-        />
+        {selected ? <ConnectionSourceHandles nodeId={id}></ConnectionSourceHandles> : null}
+        <ConnectionTargetHandles nodeId={id}></ConnectionTargetHandles>
+        <ConnectionHandles nodeId={id} customHandles={data.customHandles}></ConnectionHandles>
```
pcdavid: Simpler syntax for empty elements:

```
{selected ? <ConnectionSourceHandles nodeId={id} /> : null}
<ConnectionTargetHandles nodeId={id} />
<ConnectionHandles nodeId={id} customHandles={data.customHandles} />
```

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ImageNode.tsx
```diff
@@ -59,38 +62,9 @@ export const ImageNode = memo(({ data, isConnectable, id, selected }: NodeProps<
...
-        id={`handle--${id}--bottom`}
-        type="source"
-        position={Position.Bottom}
-        isConnectable={isConnectable}
-        style={newConnectionStyleProvider.getHandleStyle(data.descriptionId)}
-        onMouseDown={onConnectionStartElementClick}
-      />
+      {selected ? <ConnectionSourceHandles nodeId={id}></ConnectionSourceHandles> : null}
+      <ConnectionTargetHandles nodeId={id}></ConnectionTargetHandles>
+      <ConnectionHandles nodeId={id} customHandles={data.customHandles}></ConnectionHandles>
```
pcdavid: Simpler syntax for empty elements:

```
{selected ? <ConnectionSourceHandles nodeId={id} /> : null}
<ConnectionTargetHandles nodeId={id} />
<ConnectionHandles nodeId={id} customHandles={data.customHandles} />
```

pcdavid: These 3 elements (ConnectionSourceHandles, ConnectionTargetHandles and ConnectionHandles) are always used together and in the same exact way. Couldn't this be a single React component?

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/CustomHandle.tsx
```diff
@@ -0,0 +1,65 @@
...
+        position: 'relative',
+        transform: 'none',
+        left: 'auto',
+        opacity: '0',
+        pointerEvents: 'none',
+      };
+  }
+};
+
+export const CustomHandle = ({ nodeId, id, type, position, handleIndex, ...props }: CustomHandleProps) => {
```
pcdavid: We normally use the terminology "CustomXXX" for concepts that applications can extend by contributing their own (custom) versions. For example custom _widgets_, custom _menu entries_, custom _node types_, etc.
This should probably be renamed to better reflect the component's purpose.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/CustomHandle.tsx
```diff
@@ -0,0 +1,65 @@
...
+        pointerEvents: 'none',
+      };
+    case Position.Bottom:
+      return {
+        position: 'relative',
+        transform: 'none',
+        left: 'auto',
+        opacity: '0',
+        pointerEvents: 'none',
+      };
```
pcdavid: They're all *almost* identical, but the way they are defined it's difficult to see the actual differences.
Proposal:
```js
  const style: React.CSSProperties = {
    position: 'relative',
    transform: 'none',
    opacity: '0',
    pointerEvents: 'none',
  };
  switch (position) {
    case Position.Left:
    case Position.Right:
      style.top = 'auto';
      break;
    case Position.Top:
    case Position.Bottom:
      style.left = 'auto';
      break;
  }
  return style;
```

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/CustomHandle.tsx
pcdavid: `handleIndex` is not used; it can be removed from `CustomHandleProps`.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/CustomHandle.types.ts
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+import { HandleProps, HandleType, Position } from 'reactflow';
+export interface CustomHandleProps extends HandleProps {
+  id: string;
+  nodeId: string;
+  handleIndex: number;
```
pcdavid: `handleIndex` does not seem as needed.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/NodeContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+import React, { useState } from 'react';
+import { Node } from 'reactflow';
+import { NodeData } from '../DiagramRenderer.types';
+import { NodeContextProviderProps, NodeContextProviderState, NodeContextValue } from './NodeContext.types';
+const defaultValue: NodeContextValue = {
```
pcdavid, line 18: This whole `NodeContext` seem only to be used to set a dashed outline on reactangular nodes on hover.
* I'm not sure how this is related to the subject of connection handles.
* Why only on Rectangular nodes?

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ConnectionSourceHandles.tsx
```diff
@@ -0,0 +1,260 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+const customConnectionHandlesStyle = (position: Position, theme: Theme): React.CSSProperties => {
+  switch (position) {
```
pcdavid: Difficult to read what's the same (most properties) and what is different between the 4 cases.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ConnectionSourceHandles.tsx
```diff
@@ -0,0 +1,260 @@
...
+      return {
+        position: 'absolute',
+        borderRadius: '0',
+        right: '-15px',
+        border: 'solid black',
+        borderWidth: '0 2px 2px 0',
+        display: 'inline-block',
+        padding: '2px',
+        transform: 'rotate(-45deg)',
+        WebkitTransform: 'rotate(-45deg)',
```
pcdavid: Is the Webkit-specific property really needed? I don't see any other uses elsewhere in our code. On the contrary, we have several places where we use the standard `transform:` and do not add a Webkit-specific variant.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ConnectionSourceHandles.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+      if (toolSectionsError) {
+        addErrorMessage(toolSectionsError.message);
+      }
+    }
+  }, [toolSectionsLoading, toolSectionsData, toolSectionsError]);
+
+  return (
+    <>
+      {candidates.length > 0 ? (
```
pcdavid: Why 4 separate tests for the same condition?

```jsx
  if (candidates.length > 0) {
    return (
      <>
        <Handle />
        <Handle />
        <Handle />
        <Handle />
      </>
    );
  } else {
    return null;
  }
```
This applies to a few other places.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/converter/convertDiagram.ts
```diff
@@ -122,7 +144,7 @@ const toIconLabelNode = (
   const style = gqlNode.style as GQLIconLabelNodeStyle;
   const { id, label } = gqlNode;
   const labelStyle = label.style;
-
+  const customHandles: CustomHandleProps[] = [];
```
pcdavid: The name `customHandle` (here and elsewhere) should probably be changed. It does not tell what these handles are/do.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ConnectionTargetHandles.tsx
```diff
@@ -0,0 +1,51 @@
...
+    pointerEvents: 'all',
+    cursor: 'crosshair',
+  };
+};
+const connectionNodeIdSelector = (state) => state.connectionNodeId;
+export const ConnectionTargetHandles = ({ nodeId }: ConnectionTargetHandlesProps) => {
+  const connectionNodeId = useStore(connectionNodeIdSelector);
+  const isConnecting = !!connectionNodeId;
+  return (
+    <>
```
pcdavid: Why a fragment? There is at most one component.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ConnectionTargetHandles.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <Handle
+          id={`handle--${nodeId}--target`}
+          type="target"
+          position={Position.Top}
+          style={targetHandleNodeStyle()}
+          isConnectableEnd={true}
+          isConnectableStart={false}
+        />
+      ) : (
+        ''
```
pcdavid: We usually return `null` to render nothing. En empty string will render a text node in the DOM, I'm not sure that's what we want.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ConnectionHandles.tsx
```diff
@@ -0,0 +1,138 @@
...
+                nodeId={nodeId}
+                type={customHandle.type}
+                position={customHandle.position}
+                key={customHandle.id}
+                handleIndex={index}
+                isConnectable={true}
+              />
+            );
+          })}
+      </div>
```
pcdavid: It's the same pattern repeated 4 times.
It's certainly possible to do something like

```js
for (var position of [Position.Left, Position.Top, Position.Bottom, Position.Right]) {
  <div style={customHandlesStule(position)>...</>
}
```

pcdavid (changes requested):
I have not yet reviewed everything in depth, but I prefer publishing my current remarks now so that we can iterate instead of waiting any longer.

It's too late now, but the PR mixes several changes that should probably have been done separately and iteratively. As it is, it's difficult to follow what code changes correspond to which UI/behavior change between:
* distinguish between the "create connection" handles displayed on a node to start the edge creation process from the "edge source/target handle" displayed *on an edge* to trigger a reconnection;
* only display the "create connection" handles (resp. "edge source/target handle") on a node (resp. edge) when it is selected);
* hightlight the compatible nodes (and fade the incompatible ones) during the edge creation process (actually, this does not seem to work as expected);
* distribute the endpoints of multiple edges which share the same side of a node so that they are visually distinguished.

The names chosen (especially "Custom Handle") do not help.

Cypress tests should also be added.

CHANGELOG.adoc
```diff
@@ -57,6 +57,9 @@ This coupling will be removed in the near future as a consequence, the dependenc
 - https://github.com/eclipse-sirius/sirius-web/issues/2410[#2410] [diagram] Do a fit to screen on a diagram opening
 - https://github.com/eclipse-sirius/sirius-web/issues/2268[#2268] [diagram] Create child nodes where the user has clicked.
 - https://github.com/eclipse-sirius/sirius-web/issues/2475[#2475] [diagram] Prevent nodes to move each diagram refresh event. Only the size is still updated
+- https://github.com/eclipse-sirius/sirius-web/issues/2443[#2443] [diagram] Add dedicated handles to create edges.
+Hide handles between edges and each edges are now placed in differents positions around the node.
+Allow the creation of an edge by droping the connection line on a node.
```
pcdavid: Typo: "drop[p]ing"

pcdavid (changes requested):

CHANGELOG.adoc
```diff
@@ -89,6 +89,10 @@ This coupling will be removed in the near future as a consequence, the dependenc
 All icons can now be composed of multiple images, which will be overlayed on top of each other.
 An example of use has been added on the domain entity items.
 - https://github.com/eclipse-sirius/sirius-web/issues/2483[#2483] [diagram] Take the header into account in the React Flow layout.
+- https://github.com/eclipse-sirius/sirius-web/issues/2475[#2475] [diagram] Prevent nodes to move each diagram refresh event. Only the size is still updated
```
pcdavid: The line which references #2475 should not be part of this PR. Probably a mistake during a conflict resolution.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ImageNode.tsx
```diff
@@ -53,45 +67,17 @@ export const ImageNode = memo(({ data, isConnectable, id, selected }: NodeProps<
...
-        type="source"
-        position={Position.Bottom}
-        isConnectable={isConnectable}
-        style={newConnectionStyleProvider.getHandleStyle(data.descriptionId)}
-        onMouseDown={onConnectionStartElementClick}
-      />
+      {selected ? <ConnectionSourceHandles nodeId={id} /> : null}
+      <ConnectionTargetHandles nodeId={id} />
+      <ConnectionHandles connectionHandles={data.connectionHandles} />
+      {data.label ? <Label diagramElementId={id} label={data.label} faded={data.faded} transform="" /> : null}
```
pcdavid: This creates a duplicate label.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/connector/ConnectorContext.tsx
```diff
@@ -58,6 +58,7 @@ export const ConnectorContextProvider = ({ children }: ConnectorContextProviderP
       targetCandidates: [],
       isNewConnection: false,
     }));
+    setState((prevState) => ({ ...prevState, connection: null, isNewConnection: false }));
```
pcdavid: This seems unnecessary give the `setState` call just above already does the same.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ImageNode.tsx
```diff
@@ -40,11 +43,22 @@ const imageNodeStyle = (
...
+
+  const firstUpdate = useRef<boolean>(true);
+
+  useEffect(() => {
+    if (firstUpdate.current && firstUpdate) {
+      firstUpdate.current = false;
+    } else {
+      updateNodeInternals(id);
+    }
+  }, [JSON.stringify(data.connectionHandles)]);
```
pcdavid: Not sure about this, this looks like a potential performance issue.
Same remark for the other kinds of nodes.

pcdavid (changes requested):

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/node/ImageNode.tsx
sbegaudeau: - This code should not be there, we don't want to repeat that code in every single node, this belongs in a dedicated handle hook. Custom shapes will have to reuse it for example.
- Why do you need this ref? Why not a state?

I agree with @pcdavid, on top of being a potential performance issue, you do not have any idea of what would this behavior be in the future. That being said, it works for a large number of use cases it can be a pretty code solutions BUT, and I say this "but" with every single letter in upper case, I think that the unique place where this code would be needs a link to the original explanation which details why we need to be VERY VERY careful with this kind of solutions: https://github.com/facebook/react/issues/14476#issuecomment-471199055

Honestly, given the use case, I would rather have something like this:


```
const connectionHandlesIdentity = data.connectionHandles
                                      .map(handle => `${handle.nodeId}#${handle.edgeId}`)
                                      .join(', ');

useEffect(() => { ... }, [connectionHandlesIdentity])
```

The main difference being that if we change the structure of ConnectionHandle, we can select whether or not this code should be impacted.

mcharfadi: Thanks for the explanation, I'll use a string as dependency of the useffect and put it in a custom hook then.

I used a ref because a state change would trigger a re render (unlike a ref change) and call updateNodeInternals which if called too much can impact the performance a lot. From my testing in this particular use case I get more render by using a state than a ref.

pcdavid (approved):
There are some known issues/possible improvements, but the basics are in place and work. @sbegaudeau unless you see something that jusitfies a veto from you, I say we merge this and continue improving in further PRs.

pcdavid (changes requested):
Just to prevent this being merged before we decide how to proceed.

PR #2496, opened 2023-10-20
[288] Reset dragged node's position if dropped on an invalid target

### LINKED ISSUE #288: Let end users move contained elements
Today, when an element is moved in its container, it is blocked at its borders.
The purpose is to allow the move of an element to another container and manage the semantic consequences.

It should thus be possible to move a child element from one container to another. This should work with free from layouted nodes and list layouted nodes. It should also allow moving border nodes from one node to another.

pcdavid:
https://github.com/eclipse-sirius/sirius-web/assets/10608/ca30d920-8cb1-4fdd-9beb-89c9afb7d094

Bug: https://github.com/eclipse-sirius/sirius-web/issues/288

### DIFF
--- a/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/DiagramRenderer.tsx
+++ b/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/DiagramRenderer.tsx
@@ -21,6 +21,7 @@ import {
   EdgeSelectionChange,
   Node,
   NodeChange,
+  NodePositionChange,
   NodeSelectionChange,
   OnEdgesChange,
   OnNodesChange,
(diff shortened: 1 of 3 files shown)


gcoutable (commented):
Instead of giving the reset position here, is it possible to store the dragging node on start or stop directly in the hook state?

pcdavid:
> Instead of giving the reset position here, is it possible to store the dragging node on start or stop directly in the hook state?

Tried it, and for some reason I don't yet unserstand, moving the `onNodesChange([resetPosition])` (with the same `resetPosition` value) inside the `useDropNode` hook does not work (the position is not reset, even though `onNodesChange` is invoked.

frouene (approved):

PR #2502, opened 2023-10-24
[2460] Separates the management of element creation from the reference widget in specific services

### LINKED ISSUE #2460: Invalid ChildCreationDescription returned while creating new a instance from a Reference widget
While creating a new instance using the "+" button from Reference Widget that represents a containment reference the presented candidates are incorrect. The creation description should be limited to the reference being edited. In the current implementation it return all pair of <ERefence,EClass> that match the criteria <ParentType,ChildType> whereas it should take into account the current containment reference.

Using the following meta model:

image

It easy to reproduce a case where we can create "Entity one in refC1" while editing the reference "Refc2".

image

Causing the weird behavior:

BugREferenceWidgetWrongCandidateCreation

To reproduce:

[InstanceProblem.zip](https://github.com/eclipse-sirius/sirius-web/files/12857767/InstanceProblem.zip)
[Studio.zip](https://github.com/eclipse-sirius/sirius-web/files/12857768/Studio.zip)

frouene:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2460

# Pull request template

## General purpose
- [x] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/forms/backend/sirius-components-collaborative-widget-reference/src/main/java/org/eclipse/sirius/components/collaborative/widget/reference/handlers/CreateElementEventHandler.java
+++ b/packages/forms/backend/sirius-components-collaborative-widget-reference/src/main/java/org/eclipse/sirius/components/collaborative/widget/reference/handlers/CreateElementEventHandler.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package org.eclipse.sirius.components.collaborative.widget.reference.handlers;
 
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
@@ -23,6 +24,7 @@ import org.eclipse.sirius.components.collaborative.api.Monitoring;
 import org.eclipse.sirius.components.collaborative.forms.api.IFormEventHandler;
 import org.eclipse.sirius.components.collaborative.forms.api.IFormInput;
 import org.eclipse.sirius.components.collaborative.forms.api.IFormQueryService;
+import org.eclipse.sirius.components.collaborative.widget.reference.ReferenceWidgetDefaultCreateElementHandler;
 import org.eclipse.sirius.components.collaborative.widget.reference.dto.CreateElementInReferenceSuccessPayload;
 import org.eclipse.sirius.components.collaborative.widget.reference.dto.CreateElementInput;
 import org.eclipse.sirius.components.collaborative.widget.reference.messages.IReferenceMessageService;
@@ -32,8 +34,8 @@ import org.eclipse.sirius.components.core.api.IFeedbackMessageService;
 import org.eclipse.sirius.components.core.api.IObjectService;
 import org.eclipse.sirius.components.core.api.IPayload;
 import org.eclipse.sirius.components.forms.Form;
+import org.eclipse.sirius.components.widget.reference.IReferenceWidgetCreateElementHandler;
 import org.eclipse.sirius.components.widget.reference.ReferenceWidget;
-import org.eclipse.sirius.components.widget.reference.dto.CreateElementInReferenceHandlerParameters;
 import org.springframework.stereotype.Service;
 
 import io.micrometer.core.instrument.Counter;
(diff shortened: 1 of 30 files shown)


frouene:
This PR doesn't correct the issue #2460 but it's a first step to facilitate the widget reference configuration.

packages/forms/backend/sirius-components-collaborative-widget-reference/src/main/java/org/eclipse/sirius/components/collaborative/widget/reference/handlers/CreateElementEventHandler.java
```diff
@@ -87,8 +95,14 @@ public void handle(One<IPayload> payloadSink, Many<ChangeDescription> changeDesc
             if (optionalWidget.isPresent() && optionalWidget.get().isReadOnly()) {
                 payload = new ErrorPayload(input.id(), this.messageService.unableToEditReadOnlyWidget());
             } else {
-                var handlerInput = this.getHandlerInput(editingContext, input);
-                Optional<Object> optionalObject = optionalWidget.map(ReferenceWidget::getCreateElementHandler).map(handler -> handler.apply(handlerInput));
+                Optional<Object> optionalObject = this.referenceWidgetCreateElementHandlers.stream()
+                        .filter(provider -> provider.canHandle(input.descriptionId()))
+                        .findFirst()
```
pcdavid, line 100: The default `ReferenceWidgetDefaultCreateElementHandler` will always return `true` to `canHandle`. How can we be sure it's not the one we get here instead of another, application-provided one?

packages/forms/backend/sirius-components-collaborative-widget-reference/src/main/java/org/eclipse/sirius/components/collaborative/widget/reference/handlers/ReferenceWidgetChildCreationDescriptionsEventHandler.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public void handle(Sinks.One<IPayload> payloadSink, Sinks.Many<ChangeDescription> changeDescriptionSink, IEditingContext editingContext, IInput input) {
+        this.counter.increment();
+
+        List<ChildCreationDescription> childCreationDescriptions = List.of();
+        if (input instanceof ReferenceWidgetChildCreationDescriptionsInput castInput) {
+
+            childCreationDescriptions = this.referenceWidgetCreateElementHandlers.stream()
+                    .filter(provider -> provider.canHandle(castInput.descriptionId()))
+                    .findFirst()
```
pcdavid, line 69: Same remark.

packages/forms/backend/sirius-components-collaborative-widget-reference/src/main/java/org/eclipse/sirius/components/collaborative/widget/reference/handlers/ReferenceWidgetRootCreationDescriptionsEventHandler.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public void handle(Sinks.One<IPayload> payloadSink, Sinks.Many<ChangeDescription> changeDescriptionSink, IEditingContext editingContext, IInput input) {
+        this.counter.increment();
+
+        List<ChildCreationDescription> childCreationDescriptions = List.of();
+        if (input instanceof ReferenceWidgetRootCreationDescriptionsInput castInput) {
+
+            childCreationDescriptions = this.referenceWidgetCreateElementHandlers.stream()
+                    .filter(provider -> provider.canHandle(castInput.descriptionId()))
+                    .findFirst()
```
pcdavid, line 68: Same remark.

pcdavid (changes requested):

packages/forms/backend/sirius-components-collaborative-widget-reference/src/main/java/org/eclipse/sirius/components/collaborative/widget/reference/handlers/CreateElementEventHandler.java
frouene: I changed the approch, now I explicitly call this service when no other canHandle is true.
This way, an application that wants to provide its own handler, doesn't need to play with spring annotation `@Order`

pcdavid (approved):
I find it strange to keep both `ReferenceWidgetDefaultCreateElementHandler` and `ReferenceWidgetDefaultCandidateSearchProvider` as Spring Boot `@Services`.

This means they will always appear in e.g. `List<IReferenceWidgetRootCandidateSearchProvider>` so they need to return `false` on `canHandle` (which seems _wrong_ as they actually can handle all `descriptionId`), and *then* we also inject them through they concrete type to use them as fallbacks/actual defaults.

IMO they should not be `@Services`, declare they `canHandle` all cases (which is the case), and be created explicitly where needed (`ReferenceWidgetDefaultCandidateSearchProvider` has no constructor dependency, `ReferenceWidgetDefaultCreateElementHandler` only needs the `IEditService`, which we have or can get everywhere we explicitly need to build a `ReferenceWidgetDefaultCreateElementHandler`).

Still approving so that this can be merged, as the current solution works in practice, but I'll try to propose a separate PR with the approach outlined above.

pcdavid:
See https://github.com/eclipse-sirius/sirius-web/pull/2518 for a followup corresponding to my previous comment.

PR #2518, opened 2023-10-31
[2460] Reference widget default handler/provider do not need to be @Service

pcdavid:
In https://github.com/eclipse-sirius/sirius-web/pull/2502 both `ReferenceWidgetDefaultCreateElementHandler` and `ReferenceWidgetDefaultCandidateSearchProvider` are defined as Spring Boot `@Services`.

This means they will always appear in e.g. `List<IReferenceWidgetRootCandidateSearchProvider>` along with the "normal" (application-provided) ones. They need to return `false` on `canHandle()` to be ignored when iterating on these lists, which seems wrong as they actually *can* handle all descriptionId. Then we also inject them (a second time) through their concrete type to use them as fallbacks/actual defaults.

Do not mark them as `@Services`.

They can (correctly) return true on `canHandle`.

In the few places where we explictly want these default implementations, create them explicitly.
- `ReferenceWidgetDefaultCandidateSearchProvider` has no constructor dependency
- `ReferenceWidgetDefaultCreateElementHandler` only needs the `IEditService`, which we can get easily.

Bug: https://github.com/eclipse-sirius/sirius-web/issues/2460

### DIFF
--- a/packages/forms/backend/sirius-components-collaborative-widget-reference/src/main/java/org/eclipse/sirius/components/collaborative/widget/reference/ReferenceWidgetDefaultCreateElementHandler.java
+++ b/packages/forms/backend/sirius-components-collaborative-widget-reference/src/main/java/org/eclipse/sirius/components/collaborative/widget/reference/ReferenceWidgetDefaultCreateElementHandler.java
@@ -21,7 +21,6 @@ import org.eclipse.sirius.components.core.api.ChildCreationDescription;
 import org.eclipse.sirius.components.core.api.IEditService;
 import org.eclipse.sirius.components.core.api.IEditingContext;
 import org.eclipse.sirius.components.widget.reference.IReferenceWidgetCreateElementHandler;
-import org.springframework.stereotype.Service;
 
 /**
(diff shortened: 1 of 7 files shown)


packages/forms/backend/sirius-components-collaborative-widget-reference/src/main/java/org/eclipse/sirius/components/collaborative/widget/reference/ReferenceWidgetDefaultCreateElementHandler.java
```diff
@@ -21,15 +21,13 @@
...
-import org.springframework.stereotype.Service;
 
 /**
  * * Default implementation of {@link IReferenceWidgetCreateElementHandler}.
  * * This implementation use {@link IEditService}.
  *
  * @author frouene
  */
-@Service
 public class ReferenceWidgetDefaultCreateElementHandler implements IReferenceWidgetCreateElementHandler {
```
frouene, line 31: The `canHandle` should also return true to be consistent with `ReferenceWidgetDefaultCandidateSearchProvider`

frouene (approved):

PR #2529, opened 2023-11-06
[2528] Add support for multiple IObjectService

### LINKED ISSUE #2528: Add support for multiple IObjectService

### Expected behavior

With 2023.10.5, only one _IObjectService_ can be provided.
Add a mechanism allowing several _IObjectService_ to live together would be an improvement.

### Cuttings backs

...

### Potential side effects

...

No alternatives considered.

AxelRICHARD:
# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/adrs/113_add_support_for_multiple_IObjectService.adoc
+++ b/doc/adrs/113_add_support_for_multiple_IObjectService.adoc
@@ -1,0 +1,101 @@
+= ADR-113 - Add support for multiple IObjectService
+
+== Context
+
+Only one `IObjectService` can be provided in a Sirius Web based application.
+The default implementation (`org.eclipse.sirius.components.emf.services.ObjectService`) is a good default implementation but sometimes we need to customize it based on a specific predicate.
+
+== Decision
+
+* Add new `IObjectServiceDelegate`.
+All Sirius Web applications may provide as many `IObjectServiceDelegate` as they want.
+
+[source,java]
+----
+public interface IObjectServiceDelegate {
+    
+    boolean canHandle(Object object);
+    boolean canHandle(IEditingContext editingContext);
+
+    String getId(Object object);
+    String getLabel(Object object);
+    String getKind(Object object);
+    String getFullLabel(Object object);
+    List<String> getImagePath(Object object);
+    Optional<Object> getObject(IEditingContext editingContext, String objectId);
+    List<Object> getContents(IEditingContext editingContext, String objectId);
+    Optional<String> getLabelField(Object object);
+    boolean isLabelEditable(Object object);
+}
+----
+
+* Add new `IDefaultObjectService` and its implementation @Service `DefaultObjectService`.
+The default implementation `DefaultObjectService` is the same than the existing one `ObjectService`.
+
+[source,java]
+----
+public interface IDefaultObjectService {
+    
+    String getId(Object object);
+    String getLabel(Object object);
+    String getKind(Object object);
+    String getFullLabel(Object object);
+    List<String> getImagePath(Object object);
+    Optional<Object> getObject(IEditingContext editingContext, String objectId);
+    List<Object> getContents(IEditingContext editingContext, String objectId);
+    Optional<String> getLabelField(Object object);
+    boolean isLabelEditable(Object object);
+}
+----
+
+* Change @Service `ObjectService` to `ComposedObjectService`.
+`ComposedObjectService` can be moved from `org.eclipse.sirius.components.emf.services` to `org.eclipse.sirius.components.core.services`.
+
+
+[source,java]
+----
+@Service
+public class ComposedObjectService implements IObjectService {
+      private final List<IObjectServiceDelegate> objectServiceDelegates;
+
+      private final IDefaultObjectService defaultObjectService;
+...
+
+      @Override
+      public String getId(Object object) {
+          // Search for a specific provider
+          Optional<IObjectServiceDelegate> optionalObjectServiceDelegate = this.objectServiceDelegates.stream().filter(objectServiceDelegate -> objectServiceDelegate.canHandle(object)).findFirst();
+          if (optionalObjectServiceDelegate.isPresent()) {
+            IObjectServiceDelegate objectServiceDelegate = optionalObjectServiceDelegate.get();
+            return objectServiceDelegate.getId(object);
+          } 
+          // Fallback to default behavior
+          return this.defaultObjectService.getId(object);
+      }
+...
+      @Override
+      public Optional<Object> getObject(IEditingContext editingContext, String objectId) {
+          // Search for a specific provider
+          Optional<IObjectServiceDelegate> optionalObjectServiceDelegate = this.objectServiceDelegates.stream().filter(objectServiceDelegate -> objectServiceDelegate.canHandle(editingContext)).findFirst();
+          if (optionalObjectServiceDelegate.isPresent()) {
+            IObjectServiceDelegate objectServiceDelegate = optionalObjectServiceDelegate.get();
+            return objectServiceDelegate.getObject(editingContext, objectId);
+          } 
+          // Fallback to default behavior
+          return this.defaultObjectService.getObject(editingContext, objectId);
+      }
+}
+----
+
+
+== Status
+
+To be reviewed
+
+== Consequences
+
+A minor breaking change: if some existing applications may be referencing `ObjectService` by excluding it and extending it.
+They may now have to use `IObjectServiceDelegate` instead.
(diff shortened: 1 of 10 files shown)


doc/adrs/112_add_support_for_multiple_IObjectService.adoc
```diff
@@ -0,0 +1,118 @@
+= ADR-112 - Add support for multiple IObjectService
```
sbegaudeau: - We will definitely move forward with something similar to the second option. There no need for everybody to manage this complexity everywhere.
- I'm not fan of `IObjectServiceProvider` because it does not provide instances of `IObjectService`. I would rather go forward with renaming `ObjectService` to `ComposedObjectService` and introduce an  `IObjectServiceDelegate` which would have to be implemented by those integrating Sirius Components.
- It would be nice to have the default behavior implement some sort of `IDefaultObjectService` (which would be similar to IObjectService). This way, it would allow an application integrating multiple metamodels to define a custom default behavior for every metamodels and small customization where it is necessary in `IObjectServiceDelegate`.
- It should also be noted that the `IObjectService` should/will be separated into smaller interfaces in order to simplify all that (extracting an `ILabelService` and an `IImageService` would be a no brainer). It would also greatly simplify the manipulation of multiple metamodels in a single application and the reusability of metamodels in various Sirius Web applications. No need to copy/paste every time how to retrieve an object from its id if you just want to customize a label.

To sum things up, we need to have an API which would be more similar to the EMF Edit ones without the dependency to EMF.

It should also be noted that this issue was blocking the use of styled strings in the explorer for example.

AxelRICHARD: Done, except for the separation into smaller interfaces in order to simplify all that. This will be done in a second step.

packages/core/backend/sirius-components-core/src/main/java/org/eclipse/sirius/components/core/api/IObjectServiceDelegate.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Optional;
+
+/**
+ * Interface of the delegation service interacting with domain objects.
+ *
+ * @author arichard
+ */
+public interface IObjectServiceDelegate {
+
+    boolean canHandle(IEditingContext editingContext, Object object);
```
sbegaudeau: I would not give access to the `IEditingContext` in this method because most methods here do not give us the ability to leverage the editing context and if some application needs to find out the editing context from the object it can be done in application-specific code.

sbegaudeau: After a look at the rest of the code, I'm even more convinced since you have to give `null` most of the time since you can't do anything else. We will never be able to give the editing context in most use cases and applications can retrieve it by themselves if they want so no need for such variable.

packages/core/backend/sirius-components-core/src/main/java/org/eclipse/sirius/components/core/services/ComposedObjectService.java
```diff
@@ -0,0 +1,123 @@
...
+        this.defaultObjectService = Objects.requireNonNull(defaultObjectService);
+    }
+
+    @Override
+    public String getId(Object object) {
+        return this.objectServiceDelegates.stream()
+                .filter(delegate -> delegate.canHandle(null, object))
+                .findFirst()
+                .map(delegate -> delegate.getId(object))
+                .orElse(this.defaultObjectService.getId(object));
```
sbegaudeau: This is not what we agreed on. In this case, if someone creates a bad delegate returning `null` then the default value would be returned instead and that's exactly what I don't want. If a delegates indicates that it can handle something, it has to handle it properly, no magic fallback. So the code should look like this:


```
var optionalDelegate = this.objectServiceDelegates.stream()
    .filter(delegate -> delegate.canHandle(object))
    .findFirst();
if (optionalDelegate.isPresent()) {
  return optionalDelegate.get().getId(object);
)
return this.defaultObjectService.getId(object);
```

When we will split the `IObjectService` interface into multiple interfaces like `ILabelService` etc, we could simplify the code a bit but it would keep the same philosophy. I fully expect the code to crash and burn as fast as possible if someone tells us that they can handle such an important responsibility and fail to fulfill it.

sbegaudeau: Of course, this comment is valid for all the other parts of this code

doc/adrs/113_add_support_for_multiple_IObjectService.adoc
```diff
@@ -0,0 +1,84 @@
...
+== Decision
+
+* Add new `IObjectServiceDelegate`.
+All Sirius Web applications may provide as many `IObjectServiceDelegate` as they want.
+
+[source,java]
+----
+public interface IObjectServiceDelegate {
+    
+    boolean canHandle(IEditingContext editingContext, Object object);
```
sbegaudeau: You can also remove the editingContext here (see my comment below) since I didn't catch it at first.

doc/adrs/113_add_support_for_multiple_IObjectService.adoc
```diff
@@ -0,0 +1,84 @@
...
+----
+
+
+== Status
+
+To be reviewed
+
+== Consequences
+
+No consequences for existing Sirius Web applications.
```
sbegaudeau: A minor breaking change is that some existing applications may be referencing ObjectService by excluding it and extending it or things like that.

sbegaudeau: It would be a good place to mention that in the near future, we will split the `IObjectService` interface into smaller interfaces like `IIdentityService`, `ILabelService`, `IContentService` etc. There will be API breaks around this part of the code later.

doc/adrs/113_add_support_for_multiple_IObjectService.adoc
```diff
@@ -0,0 +1,84 @@
...
+      @Override
+      public String getId(Object object) {
+          // Search for a specific provider
+          Optional<IObjectServiceDelegate> optionalObjectServiceProvider = this.objectServiceDelegates.stream().filter(objectServiceProvider -> objectServiceProvider.canHandle(object)).findFirst();
+          if (optionalObjectServiceProvider.isPresent()) {
+            IObjectServiceProvider  objectServiceProvider = optionalObjectServiceProvider.get();
+            return objectServiceProvider.getId(object);
+          } 
+          // Fallback to default behavior
+          return this.defaultObjectService.getId(object);
```
sbegaudeau: - provider -> delegates
- You are not following this algorithm later :)

sbegaudeau (changes requested):

doc/adrs/113_add_support_for_multiple_IObjectService.adoc
AxelRICHARD: Done

packages/core/backend/sirius-components-core/src/main/java/org/eclipse/sirius/components/core/api/IObjectServiceDelegate.java
AxelRICHARD: Done

packages/core/backend/sirius-components-core/src/main/java/org/eclipse/sirius/components/core/services/ComposedObjectService.java
AxelRICHARD: Done

doc/adrs/113_add_support_for_multiple_IObjectService.adoc
AxelRICHARD: Done

doc/adrs/113_add_support_for_multiple_IObjectService.adoc
AxelRICHARD: Done

sbegaudeau (approved):
The code looks good, I have not executed it myself but I believe that everything would crash and burn in case of an error in this part of the code so I'll merge it :)

PR #2540, opened 2023-11-07
[2359] Ignore hidden nodes when computing DnD targets & overlaps

### LINKED ISSUE #2539: DnD target detection does not ignore hidden nodes
**Steps to reproduce**

1. Create a (React Flow) diagram with (at least) two nodes _N1_ and _N2_ (which can not be dropped on each other);
2. Select _N1_ and hide it (using the "hide" tool in its palette). Remember its position/footprint before.
3. Select _N2_ and try to move it to a position that overlaps the footprint occupied by _N1_ before it was hidden.

**Expected result**: it should be possible to move _N2_ to said position, as it is visually "free".

**Actual result**: _N2_ is reset to its initial position.

https://github.com/eclipse-sirius/sirius-web/assets/10608/3fcf9f84-1a7a-4869-8237-993b5fbb0990

This is because the node DnD code considers that the new request position overlaps a node which is not a compatible drop target, so it cancels the position change. Hidden nodes should not be considered when computing DnD targets & node overlaps.

pcdavid:
See https://github.com/eclipse-sirius/sirius-web/issues/2539 for the steps to reproduce.


gcoutable (approved):
Can be merged after the build has finished

gcoutable:
Failing e2e tests are not related to changes made in this PR, I will merge it

PR #2558, opened 2023-11-09
[2555] Add distinct edge creation handles

### LINKED ISSUE #2555: Add separate edge creation handles
This is a first subset of #2443 which will only add the _edge creation handles_.
These will be different from the handles (black dots) used as connection points between edges and nodes.
They will only be visible when a node is selected (if the node can have edges), and will only serve to initiate the edge creation process.

The implementation will be based on the part of PR https://github.com/eclipse-sirius/sirius-web/pull/2474 which corresponds to this subset.

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2555

https://github.com/eclipse-sirius/sirius-web/assets/10608/48b57e46-eb43-4519-83c9-3792eab1152c

### DIFF
--- a/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/index.ts
+++ b/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/index.ts
@@ -11,28 +11,28 @@
  *     Obeo - initial API and implementation
  *******************************************************************************/
 
-export { DiagramRepresentation } from './representation/DiagramRepresentation';
-export { DiagramPaletteToolContext } from './renderer/palette/DiagramPaletteToolContext';
-export type { DiagramPaletteToolContextValue } from './renderer/palette/DiagramPalette.types';
-export { DiagramPaletteToolContribution } from './renderer/palette/DiagramPaletteToolContribution';
-export type { DiagramPaletteToolContributionComponentProps } from './renderer/palette/DiagramPaletteToolContribution.types';
-export type { Diagram, NodeData } from './renderer/DiagramRenderer.types';
-export type { GraphQLNodeStyleFragment } from './graphql/subscription/nodeFragment.types';
-export type { NodeTypeContextValue } from './contexts/NodeContext.types';
 export { NodeTypeContext } from './contexts/NodeContext';
-export type { ILayoutEngine, INodeLayoutHandler } from './renderer/layout/LayoutEngine.types';
-export type { DiagramNodeType } from './renderer/node/NodeTypes.types';
-export { GQLViewModifier } from './graphql/subscription/nodeFragment.types';
-export type { GQLNode, GQLNodeStyle } from './graphql/subscription/nodeFragment.types';
-export { BorderNodePositon } from './renderer/DiagramRenderer.types';
+export type { NodeTypeContextValue } from './contexts/NodeContext.types';
+export type { IConvertEngine, INodeConverterHandler } from './converter/ConvertEngine.types';
 export { convertLabelStyle } from './converter/convertDiagram';
 export { AlignmentMap } from './converter/convertDiagram.types';
-export type { IConvertEngine, INodeConverterHandler } from './converter/ConvertEngine.types';
-export { NodeTypeContribution } from './renderer/node/NodeTypeContribution';
-export { useDrop } from './renderer/drop/useDrop';
-export { useConnector } from './renderer/connector/useConnector';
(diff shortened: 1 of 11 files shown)


packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/index.ts
```diff
@@ -11,28 +11,28 @@
...
+export { BorderNodePositon } from './renderer/DiagramRenderer.types';
+export type { Diagram, NodeData } from './renderer/DiagramRenderer.types';
+export { Label } from './renderer/Label';
 export { useConnector } from './renderer/connector/useConnector';
+export { useDrop } from './renderer/drop/useDrop';
 export { useDropNode } from './renderer/dropNode/useDropNode';
-export { Label } from './renderer/Label';
-export { DiagramElementPalette } from './renderer/palette/DiagramElementPalette';
-export * from './renderer/layout/layoutNode';
+export { ConnectionCreationHandles } from './renderer/handles/ConnectionCreationHandles';
```
pcdavid, line 27: `export { ConnectionCreationHandles }` is the only actual change. It is needed for the custom EllipseNode.

sbegaudeau: Thanks for mentioning this :)

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/handles/ConnectionCreationHandles.tsx
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  const candidates = useConnectionCandidatesQuery(editingContextId, diagramId, nodeId);
+  useEffect(() => {
+    const shouldRender = candidates !== null && candidates.length > 0;
+    setState((prevState) => ({ ...prevState, shouldRender }));
+    if (candidates !== null) {
+      setCandidates(candidates);
+    }
+  }, [candidates]);
```
sbegaudeau: Just nitpicking a little here but you don't need to have `shouldRender` in the state and you don't need an useEffect to update it, it can simply be derived from the `candidates` variable. Here you still need the useEffect to update the context anyway.

```
  const candidates = useConnectionCandidatesQuery(editingContextId, diagramId, nodeId);
  const shouldRender = candidates !== null && candidates.length > 0;
  useEffect(() => {
    if (candidates !== null) {
      setCandidates(candidates);
    }
  }, [candidates]);
```

The handle would thus be rendered a tiny bit quicker since you would not need a new rendering to see the updated value of `shouldRender`. I don't really care about such minor improvement in the context of this review but in general we are way overusing `useEffect` in order to compute derived values.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/handles/useConnectionCandidatesQuery.ts
```diff
@@ -0,0 +1,116 @@
...
+          });
+        });
+      }
+      setCandidates(nodesCandidates);
+
+      if (error) {
+        addErrorMessage(error.message);
+      }
+    }
+  }, [loading, data, error]);
```
sbegaudeau: Unless I'm mistaken, here you have the same small issue, everything here could be derived without any state or useEffect (you would need one for the error message only):


```
      const diagramDescription: GQLRepresentationDescription | null = data?.viewer.editingContext.representation.description ?? null;
      const candidates: GQLNodeDescription[] = [];
      if (isDiagramDescription(diagramDescription)) {
        diagramDescription.palette.tools.filter(isSingleClickOnTwoDiagramElementsTool).forEach((tool) => {
          tool.candidates.forEach((candidate) => nodesCandidates.push(...candidate.targets));
        });
        diagramDescription.palette.toolSections.forEach((toolSection) => {
          toolSection.tools.filter(isSingleClickOnTwoDiagramElementsTool).forEach((tool) => {
            tool.candidates.forEach((candidate) => nodesCandidates.push(...candidate.targets));
          });
        });
      }

      useEffect(() => {
        if (error) {
          addErrorMessage(error.message);
        }
      }, [error]);
```

We have been doing this kind of error pretty much everywhere :)

sbegaudeau (commented):
I have not executed the code, I'll let @mcharfadi test it but the code looks ok.

mcharfadi (approved):
I executed the code, and it worked well

PR #2561, opened 2023-11-09
[2549] Use our algorithm instead of elk.js to do the arrange-all

### LINKED ISSUE #2549: The arrangeAll action in reactflow breaks the list layout

gcoutable:
# Pull request template

## General purpose
- [x] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/layout/useLayout.ts
+++ b/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/layout/useLayout.ts
@@ -13,13 +13,15 @@
 
 import { ServerContext, ServerContextValue } from '@eclipse-sirius/sirius-components-core';
 import { useCallback, useContext, useEffect, useState } from 'react';
-import { useReactFlow, XYPosition } from 'reactflow';
+import { Node, XYPosition, useReactFlow } from 'reactflow';
 import { NodeTypeContext } from '../../contexts/NodeContext';
 import { NodeTypeContextValue } from '../../contexts/NodeContext.types';
-import { Diagram, EdgeData, NodeData } from '../DiagramRenderer.types';
-import { cleanLayoutArea, layout, performDefaultAutoLayout, prepareLayoutArea } from './layout';
+import { EdgeData, NodeData } from '../DiagramRenderer.types';
+import { DiagramNodeType } from '../node/NodeTypes.types';
 import { LayoutContext } from './LayoutContext';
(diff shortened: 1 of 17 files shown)


packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/layout/useLayout.ts
```diff
@@ -109,9 +111,17 @@ export const useLayout = (): UseLayoutValue => {
     });
   }, []);
 
+  const arrangeAll = useCallback((afterLayoutCallback: (laidoutDiagram: RawDiagram) => void): void => {
```
sbegaudeau: Why do you need `useCallback` here?

gcoutable: I would have said to prevent the arrangeAll function from being created repeatedly over renders.

gcoutable: It may also but used as useEffect dependency (like all functions in hook)

sbegaudeau: - Creating the function over and over is not an issue per-se, we are creating tons of functions all the time everywhere
- Technically you are right, it would improve the performance of some code using it as a dependency in an `useEffect` (it may even prevent infinite loops in some use case) but here, you are definitely not using the `arrangeAll` function in an `useEffect`

Some people, mostly those writing frameworks, are using `useMemo` and `useCallback` on everything they return to prevent issues in every possible use cases that their consumers will create. It has a memory impact and it makes the code more complex. I would say that this `useCallback` is not necessary of today so we don't need it because others will see that kind of code and try to put it everywhere.

The React Forget compiler is supposed to remove this need soon (in React 19 I think) so I don't want to put more `useMemo` / `useCallback` then necessary in our code base.

gcoutable: I will remove the `useCallback`

sbegaudeau (approved):

PR #2564, opened 2023-11-10
[2559] Allow to target the whole node when creating or reconnecting an edge

### LINKED ISSUE #2559: Allow to target the whole node when creating or reconnecting an edge
Currently when creating a new edge or reconnecting an existing one, the user is required to point to one of the 4 relatively small handles on the target node.
It would improve the ease of use if the use could point anywhere inside the target node (except of course inside subnodes).

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2559


mcharfadi:
I'm not seeing the ConnectionLine when starting a new connection. One fix is to call `updateNodeInternals(params.nodeId);` in `onConnectStart`.
Also, I don't see the hovered compatible nodes highlighted when I'm connecting. Do you think we should add it later ?

pcdavid:
For the missing `ConnectionLine`, it was caused by an additional commit that was not directly related to this PR. I've reverted it.

All the feedback/highlight are known to be missing/incorrect at the moment, it's my next task after this one.

mcharfadi (approved):

PR #2585, opened 2023-11-16
[2584] Add feedback during edge creation

### LINKED ISSUE #2584: Add feedback during edge creation
As mentioned in [the corresponding shape](https://github.com/eclipse-sirius/sirius-web/pull/2554):

During the edge creation or reconnection process (the user is dragging a [connection line](https://reactflow.dev/examples/edges/custom-connectionline) from its source to the wanted target):
* nodes which are known to be incompatible with any kind of edge creation tool starting from the source are temporarily faded;
* when the mouse cursor hovers on a compatible target node, that node is highlighted (using the same style as uses for node DnD);
* nodes (compatible but not hovered) which where not faded by the user keep their style
* nodes (compatible but not hovered) which where faded by the user are temporarily unfaded to indicate they are compatible.

pcdavid:


pcdavid:
Still a work in progress, this version does not handle edge reconnections.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/connector/useConnector.tsx
```diff
@@ -78,6 +94,7 @@ export const useConnector = (): UseConnectorValue => {
     if (params.nodeId) {
       updateNodeInternals(params.nodeId);
     }
+    setIsNewConnection(true);
```
mcharfadi: When we start a new connection or we start to reconnect an edge, onConnectStart is called.
We added `setIsNewConnection and isNewConnection` to know if we are starting a new connection or if we are reconnecting an existing edge.
The reason was that when we implemented [the feedback for edge creation](https://github.com/eclipse-sirius/sirius-web/pull/2331), the feedback for edge reconnexion was not supported.
So this line needs to be removed until we implement feedback for edge reconnection.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/connector/useConnector.tsx
```diff
@@ -41,17 +43,31 @@ export const useConnector = (): UseConnectorValue => {
...
+          if (isSelectedNode) {
+            // Hightlight the selected target
+            boxShadow = `0px 0px 2px 2px ${theme.palette.primary.main}`;
+          }
+          // Make sure all compatible nodes, even normally faded ones, are fully visible
+          opacity = '1';
+        } else {
+          // Force fade all incompatible nodes
+          opacity = '0.4';
+        }
```
mcharfadi, line 66: Just curious since I don't see any issue with the result. But any reason why you did not keep the previous way of setting the values of the returned style ? It seems strange to me to set undefined values to filter them out later.

pcdavid: You're right, the current version is halfway between the old version, which caused issues when combined with the style customizations from `useDropNode`, and what I think is a cleaner/easier to read *and* more reliable way.
I'll rework that.

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/connector/useConnector.tsx
pcdavid: But this version of the PR *does* implement feedback for reconnection.
I commented yesterday that this was not yet supported, but that was for the version from yesterday.

pcdavid: I was wrong: reconnection feedback only seemed to work "by chance" on the cases I tested. We actually do not have (and never have had) the backend information required for proper reconnection feedback.
I'll remove the half-baked and misleading attempt. Reconnection feedback will be handled later.

mcharfadi (approved):

PR #2664, opened 2023-11-28
[2579] Fix invalid cursor position when typing a new label directly

### LINKED ISSUE #2579: Invalid text cursor position when triggering direct edit on React Flow diagram
Steps to reproduce:

1. Create/open a diagram which uses React Flow and with at least one node with an editable label.
2. Select the node.
3. Start typing a new text for the label (e.g. "New Label").

The first character you type is placed at the end of the input, but the cursor is at the beginning of the input (so before the initial character). The resulting text is "ew LabelN" with the "N" at the end instead of the beginning.

### LINKED ISSUE #2798: Direct Edit display a feedback too specific
Direct Edit feedback is not correct for some node. For example, when user activates direct edit on a comment, the feedback proposes "Enter the new name" but a comment has no "name", just a "body".

* [X] **I have checked that this bug has not yet been reported by someone else**
* [X] **I have checked that this bug appears on Chrome**
* [X] **I have specified the version** : 2023.10.9
* [X] **I have specified my environment** : Window 11

### Screenshots

directEdit

### Steps to reproduce

Activate direct edit on a node.

### Expected behavior

When user activates direct edit, the feedback should be more general such as "Enter the new value" or "Enter the new {featureName}" (last proposition is the prefer one from some client). An other solution could be to remove the feedback (maybe a bad idea because the color of the field to enter the new text is the same than the node so it could be difficult to identify that direct edit is activate).

### Actual behavior

Direct Edit feedback is not correct for some node. For example, when user activate dire edit on a comment, the feedback propose "Enter the new name" but a comment has no "name", just a "body".

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2579

**Before**

[Capture vidéo du 2023-11-28 15-26-11.webm](https://github.com/eclipse-sirius/sirius-web/assets/10608/f2044677-2d15-4102-8446-4369da9bcfcd)

**After**

[Capture vidéo du 2023-11-28 15-26-53.webm](https://github.com/eclipse-sirius/sirius-web/assets/10608/69f88fd8-a09a-4270-8b48-6e9ed605d406)


packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/direct-edit/DiagramDirectEditInput.tsx
```diff
@@ -140,6 +140,13 @@ export const DiagramDirectEditInput = ({ labelId, editingKey, onClose, transform
     return cleanup;
   }, [initialLabelItemError, initialLabelItemData]);
 
+  useEffect(() => {
+    if (textInput.current) {
+      const text = textInput.current.value;
+      textInput.current.setSelectionRange(text.length, text.length);
```
gcoutable, line 146: Why setting the selection range to the size of the new value (which should be 0 if direct the direct edit has been triggered thanks to the direct edit tool or 'f2', or 1 when the user start to type)?

frouene:
I've upadted the cypress test to match the new way of doing things

packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/direct-edit/DiagramDirectEditInput.tsx
pcdavid: From what I found, using `setSelectionRange` is the only way change the cursor (caret) position.

The whole point of the PR is that when triggering direct edit by starting to type, we end up with an initial state of the input like this: "$A", where "A" is the letter typed by the user and "$" is the position of the caret (0).
calling `setSelectionRange(text.length, text.length)` ensures that when we display the input, we always put the caret at the end, i.e. "A$".

frouene (approved):

PR #2677, opened 2023-11-29
[2580] Cancel direct edit on 'Escape'

### LINKED ISSUE #2580: Hitting "Esc" while editing a label in a React Flow diagram validates the edition instead of cancelling it
Steps to reproduce:

1. Create/open a diagram which uses React Flow and with at least one node with an editable label.
2. Select the node. Say the initial label is "initial"
3. Hit F2 and start typing a new label, "new label", then hit "Esc" to cancel the current edition.

**Expected:** The edition should be canceled and the label reverted to its original value ("initial") without any call to the backend (except for getting the initial text, but not after hitting "Esc").

**Actual:** The label edition tool is invoked with the current text at the time "Esc" is hit, validating the label change.

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2580

To test, check (on a React Flow diagram and an element which supports direct edit):
1. Finishing a direct edit with an explicit _Return_ triggers the rename.
2. Finishing a direct edit by simply clicking elsewhere ("on blur") triggers the rename.
3. Finishing a direct edit by hitting _Escape_ *cancels* the operation, does not trigger the mutation and reverts to the initial state on the front.

### DIFF
--- a/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/direct-edit/DiagramDirectEditInput.tsx
+++ b/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/direct-edit/DiagramDirectEditInput.tsx
@@ -77,10 +77,10 @@ export const DiagramDirectEditInput = ({ labelId, editingKey, onClose, transform
     newLabel: initialLabel,
   });
 
+  const editionFinished = useRef<boolean>(false);
+
   const { addErrorMessage, addMessages } = useMultiToast();
-
   const { hideDiagramElementPalette } = useDiagramElementPalette();
-
   const { editingContextId, diagramId } = useContext<DiagramContextValue>(DiagramContext);
 
   const textInput = useRef<HTMLInputElement | HTMLTextAreaElement | null>(null);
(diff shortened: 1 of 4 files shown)


packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/direct-edit/DiagramDirectEditInput.tsx
```diff
@@ -77,10 +77,10 @@ export const DiagramDirectEditInput = ({ labelId, editingKey, onClose, transform
     newLabel: initialLabel,
   });
 
-  const { addErrorMessage, addMessages } = useMultiToast();
+  const cancelled = useRef<boolean>(false);
```
gcoutable: Why are you using a `useRef` instead of putting this in the state?

pcdavid: I tried several approaches, the first one being obviously to add `cancelled` into the state, but the way `onBlur` is invoked, it never sees the state change.

I'm not really satisfied with `useRef`, but it's the only way I found to make it work.

frouene:
I've slightly changed the behavior to avoid `editLabel` mutation to be called two times on enter (it was called on onBlur action too)

frouene (approved):

PR #2719, opened 2023-12-01
[2600] Add a small margin at the top of a page

### LINKED ISSUE #2600: Add some padding the PageAction
The page action buttons overlaps the Page bottom bar. 

image

The icon should be smaller to avoid this effect.

pcdavid:
This helps delineate the page's tab from the content, in particular
when the page contains action buttons.

Bug: https://github.com/eclipse-sirius/sirius-web/issues/2600

**Before**

before

**After**

after

Note the added space above the "Group Action" button which is no longer too close to the page tabs separator.


AxelRICHARD:
I will review it soon

AxelRICHARD (approved):

sbegaudeau:
I have changed the implementation of the fix for a couple reasons. First, try not to use `margin` since it breaks the encapsulation of the component. With a margin top, the page component imposes a constraint to those which reuse it. A component should not make any hypothesis on the place it will be used. Layouting sibling elements is the role of their parent element, including adding a gap between elements.

PR #2811, opened 2023-12-11
[2765] Add support for portal representation

pcdavid:

PR #2838, opened 2023-12-19
[2765] Add support for user-configurable layout

### LINKED ISSUE #2765: Add support for a "Portal" representation
A portal can be created anywhere (on any semantic element in the project) and is just a kind of dashboard where the end-user can add other representations (of any kind) so that they can be all opened at the same time.

See https://github.com/eclipse-sirius/sirius-web/pull/2691 for the corresponding shape.

A more detailed ADR focusing on the technical aspects will be proposed soon.

pcdavid:
Add support for user-configurable layout on portals using `react-grid-layout`.


pcdavid:
It's advanced enough that it no longer makes sense to have a separation "inital version". All the changes are now in #2811.

PR #2868, opened 2024-01-03
[2848] Provide a common ShareRepresentationModal

### LINKED ISSUE #2848: Provide a common ShareRepresentationDialog
Most representation kinds have (or should/will have) a "share" button in their toolbar, which opens a modal dialog with the direct URL of the representation and copies it into the system clipboard.

* Diagrams have [`ShareDiagramDialog`](https://github.com/eclipse-sirius/sirius-web/blob/master/packages/diagrams/frontend/sirius-components-diagrams-reactflow/src/renderer/ShareDiagramDialog.tsx)
* Gantt have [`ShareGanttModal.tsx`](https://github.com/eclipse-sirius/sirius-web/blob/master/packages/gantt/frontend/sirius-components-gantt/src/share-gantt/ShareGanttModal.tsx) which is very similar but takes the URL to share as a prop instead of hard-coding the use of `window.location.href`.
* Portals and Deck representation will need one.
* Forms should also probably have one, but they currently do not have a toolbar where to place it.

Anyway, there is nothing specific to the nature of the representation shared in this modal, so it would be better to have a single common implementation.

In addition, the use of `window.location.href` (either hard-coded or passed as a prop) will soon become incorrect. For representations embedded inside a portal, sharing the representation should not share the current URL, which is the portal's URL.

Proposed interface for the shared implementation:

```js
export interface ShareRepresentationModalProps {
  editingContextId: string;
  representationId: string;
  onClose: () => void;
}
```

The URL shared would need to be computed from `editingContextId` and `representationId`.
The issue is that this mapping between `editingContextId` and `representationId` on one side and the URL on the other side is defined in `sirius-web-application/src/router/Router.tsx` at the application level.

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2848


pcdavid:
The build is red because of a single Cypress test failure, which is unrelated ("ResizeObserver loop limit exceeded").

doc/adrs/124_common_share_representation_modal.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+== Decision
+
+We will provide a common implementation of `ShareRepresentationModal` in `sirius-components-core`.
+It will be based on the existing implementation from `ShareDiagramDialog` but with a different API:
+
+```js
+export interface ShareRepresentationModalProps {
+  editingContextId: string;
+  editingContextId: string;
```
sbegaudeau: `editingContextId` -> `representationId`

packages/gantt/frontend/sirius-components-gantt/src/Gantt.tsx
```diff
@@ -0,0 +1,154 @@
+/*******************************************************************************
```
sbegaudeau: Is that a duplicate of `src/representation/Gantt.tsx` ? I'll check

doc/adrs/124_common_share_representation_modal.adoc
sbegaudeau: Done

packages/gantt/frontend/sirius-components-gantt/src/Gantt.tsx
sbegaudeau: I think it's just a small detail from the rebase conflict that can be removed.

packages/portals/frontend/sirius-components-portals/src/representations/PortalToolbar.tsx
```diff
@@ -10,58 +10,112 @@
...
+        representationId={representationId}
+        onClose={closeModal}
+      />
+    );
+  }
+
   return (
-    <Paper>
-      {fullscreen ? (
+    <>
```
sbegaudeau, line 50: It seems that you have duplicated the whole portal toolbar, every button is present twice now.

sbegaudeau: I've fixed it

sbegaudeau:
There seems to be an issue with the commit message or a bug with Github because it looks very odd:

Capture d’écran 2024-01-24 à 10 07 23

I'll fix it too. I'm investigating one last issue, once it will be done I'll merge it.

packages/portals/frontend/sirius-components-portals/src/representations/PortalToolbar.tsx
pcdavid: Probably a mistake during yesterday's rebase, sorry.

pcdavid:
> There seems to be an issue with the commit message or a bug with Github because it looks very odd

Looks like I amended the then-current master when I rebased yesterday.

sbegaudeau:
> > There seems to be an issue with the commit message or a bug with Github because it looks very odd
> 
> Looks like I amended the then-current master when I rebased yesterday.

That may be it indeed, I was very surprised to see that 😛, I'll wait for the build and merge the PR this morning

sbegaudeau (approved):

PR #2876, opened 2024-01-05
[2875] Make the Representations view a plain FormBasedView

### LINKED ISSUE #2875: The Representations view should be a normal FormBasedView
It shares 99% of the code (except for some bug fixes and improvements which where only made on FormBasedViews, e.g. #1068).

The only tricky part is that although the _Representations_ view uses a Form-based List widget, its frontend does not display the whole form with the tabs/pages and groups.

I was able to get the same result by adding an optional `postProcessor` function to `FormBasedView` to extract the relevant part of the _Form_ returned by the backend and only display that. This feels a little hackish but I'll push a draft PR to start a discussion.

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2875


pcdavid:
@sbegaudeau I'd like your opinion on [this change](https://github.com/eclipse-sirius/sirius-web/pull/2876/files#diff-795976045b5fede0c4ee5b5c1583da46721a6810c629593cc6605a10000a95b3).

packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/RepresentationsEventProcessorFactory.java
AxelRICHARD, line 1: Should it be possible to also remove this RepresentationsEventProcessorFactory?
There are no RepresentationsEventInput anymore, so may be this RepresentationsEventProcessorFactory could also be removed.

pcdavid: We need it because it takes the FormDescription to use from a `IRepresentationsDescriptionProvider` which is specific to the "Representations" view.

`RelatedElementsEventProcessorFactory` has the same "issue"

Neither can just use `PropertiesEventProcessorFactory` (which has its own logic of where to find the FormDescription, including defaulting to `IPropertiesDefaultDescriptionProvider`) nor `FormEventProcessorFactory` (which looks into the generic `IRepresentationDescriptionSearchService` and will find among others FormDescription defined in studios).

All of there are very similar and only differ on where and how they find the FormDescription to use. There is probably be a way to reduce the code duplication here, but I don't feel it's in the scope of this PR/issue.

AxelRICHARD: Ok thank you for the details

AxelRICHARD (approved):

pcdavid:
The Cypress tests failed twice on the same, which looks like it could be related; I'll have a look before merging.

pcdavid:
> The Cypress tests failed twice on the same, which looks like it could be related; I'll have a look before merging.

Indeed, the PR also uses the new post-processing step to "unwraps" the contents of the Form used for the "Related Elements" view so that only the intersting part is displayed. The Cypress test needed to be updated to reflect that.

PR #2878, opened 2024-01-05
[2926] Support for libraries in Sirius Web

### LINKED ISSUE #2926: Support for libraries in Sirius Web

### Expected behavior

Users want to store and have access to read only data (a.k.a libraries).
Users should be able to click on a template/project and have in their projects a set of documents with objects referencing objects from libraries.

### Cuttings backs

In a first iteration, the objects from libraries and the libraries will not be visible by default in Sirius Web UI.

AxelRICHARD:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/adrs/124_add_support_for_libraries.adoc
+++ b/doc/adrs/124_add_support_for_libraries.adoc
@@ -1,0 +1,37 @@
+= ADR-124 - Add support for libraries
+
+== Context
+
+Users want to store and have access to read only data (a.k.a libraries).
+
+== Decision
+
+=== Backend
+
+Introducing a new `IEditingContextProcessor` allowing to execute operations on editing context before and after its loading in `EditingContextSearchService`.
+
+```
+public interface IEditingContextProcessor {
+    void preProcess(IEditingContext editingContext);
+    void postProcess(IEditingContext editingContext);
+}
+```
+
+This will allow specifiers to load as many resources as they want in the resource set of an editing context while loading it.
+
+If these resources does not have the scheme `sirius://`, they will not be considered as documents in Sirius Web.
+
+Libraries as defined in the context above should be handled by `IEditingContextProcessor`.
+
+Specifiers will be allowed to declare as many `IEditingContextProcessor` as they want.
+
+For the version number of a library, its the responsibility of the specifier to handle it from an `IEditingContextProcessor`.
+
+=== Frontend
+
+Nothing to do in a first version.
+In a second version, users should be able to view the libraries in the Explorer.
+
+== Status
+
+Work in progress
--- a/doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
+++ b/doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
@@ -1,0 +1,47 @@
+= Add the support for libraries
+
+== Context
+
+Users want to store and have access to read only data (a.k.a libraries).
+
+== Key Result
+
+Users should be able to click on a template/project and have in their projects a set of documents with objects referencing objects from libraries. As a result, without even a single interaction, users should be able to have a valid resource set.
+
+The libraries are immutable.
+
+== Solution
+
+1. When declaring templates, specifiers should be able to specify a list of libraries that will be accessible in the projects created from the templates.
+
+2. The `Blank project` is not a template. So with the proposed solution, it will not be possible to use libraries in a `Blank project`.
+To handle this specific case of the `Blank project`, it should be possible to add libraries to every project.
+This will be done programmatically, by the specifier.
+
+3. The objects from libraries and the libraries should be visible somewhere from a project (e.g. Explorer view, ...).
+
+4. Once a project created, users/specifiers should be able to modify the list of libraries associated (i.e. load/unload) to a project.
+
+5. Libraries have version number.
+At the end, it should be possible to update the version of a library for a project.
+
+=== Breadboarding
+
+Same UI as before.
+
+=== Cutting backs
+
+In a first iteration, the objects from libraries and the libraries will not be visible.
+In a first iteration, it won't be be possible to update the version of a library for a project.
+
+== Rabbit holes
+
+A template may add a lot of libraries, and thus a lot a model elements.
+These model elements will be accessible in the reference widgets from the Detail view.
+In some case the list of elements accessible in the reference widgets will be huge.
+Specifiers may have to programmatically filter the candidates of reference widgets.
+Note that this is also the case without the introduction of libraries, but libraries may emphasize the case.
+
+== No-gos
+
+N.A.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,7 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: vscode
-          path: vscode-extension/sirius-web-2023.12.10.vsix
+          path: vscode-extension/sirius-web-2023.12.11.vsix
           retention-days: 20
 
       - name: Compress the code coverage results
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -82,6 +82,8 @@ The default implementation of the editing context for Sirius Web, which had Siri
 This default implementation will now also contain all the view models properly loaded which should be considered during the lifecycle of the editing context.
 - https://github.com/eclipse-sirius/sirius-web/issues/2796[#2796] [sirius-web] All Flow related configurations have been moved to `sirius-components-flow-starter` module.
 - https://github.com/eclipse-sirius/sirius-web/issues/2883[#2883] [diagram] Add the attribute `resizedByUser` to reactflow nodes and graphql nodes.
+- https://github.com/eclipse-sirius/sirius-web/issues/2926[#2926] [sirius-web] A new `IEditingContextProcessor` has been introduced.
+It allows to execute operations before/after the load of an `IEditingContext` in `EditingContextSearchService`.
 
 
 === Dependency update
@@ -126,6 +128,7 @@ Now, for a multi-valued feature, the values are properly displayed and the delet
 +
 The size of nodes that do not have been resized manually tries to fit the node default size.
 - https://github.com/eclipse-sirius/sirius-web/issues/2792[#2792] [diagram] Add a first version for arrange all support.
+- https://github.com/eclipse-sirius/sirius-web/issues/2926[#2926] [sirius-web] Add a first version of the support for {pre|post}processing of the editing context.
 
 === Improvements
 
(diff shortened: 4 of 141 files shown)


doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
```diff
@@ -0,0 +1,39 @@
+= Add the support for libraries
+
+== Context
+
+Users want to have access to read only data (a.k.a libraries).
```
sbegaudeau: I would say store and retrieve read only data because we will need to discuss about the storing part too

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+== Key Result
+
+In the Details View, users could reference these read only data from their models' elements.
```
sbegaudeau: You are detailing a use case where I can reference libraries with the details view but that's not even the first use case. I should be able to click on a template and have in my project a et of documents with objects referencing objects from libraries. As a result, without even a single interaction, I should be able to have a valid resource set.

The key technical result will be what's given to `EditingContextSearchService`, it does not involve in any way the Details view which is not involved at all in this feature.

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+== Solution
+
+1. When declaring templates, specifiers should be able to specify a list of libraries that will be accessible in the projects created from the templates.
+
+2. Specifiers should be able to associate libraries to projects' natures. Thus, when a project with a specific nature is created/uploaded, the libraries associated to this specific nature will be accessible.
```
sbegaudeau: Why such coupling? Two projects with the same nature may have a different set of libraries (see Papyrus Web).

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+3. The `Blank project` has no nature and is not a template. So with the proposed solution, it will not be possible to use libraries in a `Blank project`. To handle this specific case of the `Blank project`, it should be possible to add libraries to every project without nature. This will be done programmatically, by the specifier.
+
+These libraries are models stored as files.
```
sbegaudeau: We will have a lot to discuss with regard to the technical aspect of the solution. While I don't think the solution would be far from this, I would remove that to keep it in the ADR since there will be way more thinks to consider than just "stored as files"

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
sbegaudeau: Another important key result is the fact that the libraries should be immutable, the read only aspect is important.

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
```diff
@@ -0,0 +1,39 @@
...
+Then, in the Details View, users could reference these read only data from their models' elements.
+Furthermore, specifiers will be able to reference these read only data programmatically (e.g. when invoking creation tools, set operations,...).
```
sbegaudeau: Once loaded in the editing context, nothing should separate those objects from regular objects. We will have regular EMF objects used in any representation just like other objects. The only difference will come from how their resource is considered by Sirius Web (The only difference may come from the schema of the URI of their resource or something as small as that).

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+=== Cutting backs
+
+== Rabbit holes
+
+A template may add a lot of libraries, and thus a lot a model elements.
+These model elements will be accessible in the reference widgets from the Detail view.
+In some case the list of elements accessible in the reference widgets will be huge.
+Specifiers may have to programmatically filter the candidates of reference widgets.
+Note that this is also the case without the introduction of libraries, but libraries may emphasize the case.
```
sbegaudeau, line 43: I'm not sure that's related at all with this issue. We could talk more to say on "how to keep people from modifying those read only objects" for example.

sbegaudeau (changes requested):

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
AxelRICHARD: done

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
AxelRICHARD: done

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
AxelRICHARD: done

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
AxelRICHARD: done

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
AxelRICHARD: done

doc/iterations/2024.1/shapes/add_support_for_libraries.adoc
AxelRICHARD: That's why I've added the last sentence. The problem raised is not introduced by the libraries may emphasize it.

doc/adrs/124_add_support_for_libraries.adoc
```diff
@@ -0,0 +1,147 @@
+= ADR-124 - Add support for libraries
+
+== Context
+
+Users want to store and have access to read only data (a.k.a libraries).
+
+== Decision
```
sbegaudeau, line 7: This starts very quickly with tons of implementation details while we should step back a little and see the big picture. I'm not even sure that any of those technical details are relevant.


A library is a set of read only objects, at the end of the day Sirius Components has no opinion on the various objects it can interact with (EMF, non EMF, etc) so this will only be related to Sirius Web. On top of that, the Sirius Components part of the frontend does not care about the origin of the data, it only interacts with representations.

The first feature necessary is to be able to use existing features with a library. In order to work in Sirius Web, a library has to be an EMF resource loaded properly into the resource set. Since users may end up with libraries in their project from various source (a project template, an action performed during the manipulation of the project), we will need to ensure the consistency of the editing content while being loaded and saved.

In the case of a project template creating a project with some data and libraries, the task will be mostly about creating a resource set with the proper data and save it properly. When the project will really be open, then the loading and resolution should be performed properly. In the case of something more dynamic like adding or removing a library on the fly, we will modify the resource of the editing context directly and then we should save everything correctly. We will not add directly a way to add / remove libraries on the fly but specifiers may end up coding something like that for their needs.

### Loading

So the first key part to be modified is the `EditingContextSearchService#findById` in order to load properly a resource set with references to a library regardless of how this reference has been created.

Pretty much everything in this topic is simply related to EMF URI resolution thanks to URIHandlers and URIConverters. Such mechanisms are more than enough to allow us to load libraries from static resources stored in jars or from a specific place in the database (see `URIHandler#canHandle` and `URIHandler#createInputStream` to load a specific kind of resource from anywhere).

If the resource set is saved with proper URIs, everything should work quite easily.

### Saving

Since libraries are not be to modified and since they may come from external projects, they should not be saved at all. We should be able to identify them from their URI just like we should be able to tell from their URI whether they are coming from the database or a static file in a jar.

If a specifier wants to create a Java service to add / remove a library from the editing context, the only thing to do will be to update the resource set to add / remove the relevant resources with some specific kind of URI.

We should specify how to identify the URI of a resource coming from the Sirius Web database as a library and a mechanism to let others specify how to resolve libraries from static resources. In a Sirius Web application containing a jar with a folder `/MY_PROFILES/MyProfile.profile.uml`, a specifier should be able to say that every reference to `pathmap://MY_PROFILES/MyProfile.profile.uml` should be resolved to this static file. As such an end user could import an existing project created with any kind of EMF based tool with a reference to this UML profile.

In a similar manner, it should be possible to export from Sirius Web, EMF resources with proper references to stay compatible with other tools.

doc/adrs/124_add_support_for_libraries.adoc
```diff
@@ -0,0 +1,147 @@
...
+public interface IEditingContextLibrary {
+  String getLibraryId();
+  String getEditingContextId();
+}
+```
+
+=== GraphQL
+
+A library is like a document, but with immutable data.
+A library is not visible in the Explorer view.
```
sbegaudeau: I don't see the need for 90% of the code here, as for this part, if we can identify the resources of the resource sets which are libraries thanks to their URIs, we can decide to hide them (or show them but change their colors or anything) in the explorer fairly easily.

I don't see why the frontend would need to know anything specific to libraries at all for them to work.

Now there can be specific kind of libraries for example UML profile, which have a very specific workflow with versioning, a custom dialog to add and remove them, multiple versions at the same time, etc) but I don't see how we could standardize in Sirius Web such exotic requirements which are very specific to some kinds of libraries and which can be done one top of Sirius Web if we provide the mechanisms that I've discussed above.

doc/adrs/124_add_support_for_libraries.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+=== Backend
+
+Introducing a new `IEditingContextProcessor` allowing to execute operations on editing context before and after its loading in `EditingContextSearchService`.
+
+```
+public interface IEditingContextProcessor {
+    void preProcess(IEditingContext editingContext);
+    void postProcess(IEditingContext editingContext);
+}
```
pcdavid, line 17: Its not clear to me from the ADR when these are invoked wrt to the editing context's lifecycle.

AxelRICHARD: The sentence above says: "Introducing a new `IEditingContextProcessor` allowing to execute operations on editing context before and after its loading in `EditingContextSearchService`."
So `preProcess` and `postPrcoess` are invoked in `EditingContextSearchService`.
I can make it more explicit if you want.

pcdavid: Sorry, my bad, I read to quickly.

sbegaudeau (approved):

PR #2879, opened 2024-01-05
[2877] Add support for breadcrumbs in Portals

### LINKED ISSUE #2877: Add support for breadcrumbs in Portals
When a representation is embedded inside a Portal, its frame should not display only the representation's name, but also the path from the root of the project to the representation's target element.

In other words, if "Topography unsynchronized" from below were embedded in a portal, the frame header should be "FlowNewModel > NewSystem > CompositeProcessor1 > Topography unsynchronized", with each segment a link to select the corresponding element (and in the case of the last segment, open the representation itself "in full").

image

pcdavid:


pcdavid:
Note that this version completely crashes the backend.

For some reason I don't yet completely understand, the editing context event processor stays blocked indefinitly on `future.get()` after submitting the `() -> this.doHandle(payloadSink, input)`.
Well, not indefinitely in this version as I added a 5s timeout (which seems like a good idea anyway, though the actual value of delay is debatable and should probably be configurable).

Anyway, I don't really see how the new GraphQL field, data fetcher and handler I added are different from what's already done for `RepresentationMetadataDescriptionDataFetcher`.

In summary, the changes are:

```
// portal.graphqls
extend type RepresentationMetadata {
  breadcrumbs: [WorkbenchSelectionEntry!]!
}
```

The corresponding data fetcher is `RepresentationMetadataBreadcrumbsDataFetcher` and follows the same patterns as `RepresentationMetadataDescriptionDataFetcher`:
* read the data from the GraphQL environment
* create a custom input record, here a `RepresentationBreadcumbsInput` instead of a `GetRepresentationDescriptionInput`
* the actual handling is done in `RepresentationBreadcrumbEventHandler` which emit a `RepresentationBreadcrumbsSuccessPayload` (instead of a `GetRepresentationDescriptionPayload`) and `ChangeKind.NOTHING` change description.

The only difference I see is that on the front:
* I fetch (possibly) multiple "instances" of this field, one for each representation embedded inside a portal:
```
views {
  id
  representationMetadata {
    id
    kind
    label
    breadcrumbs {
      id
      kind
      label
    }
  }
}
```
* this is done in the context of a `subscription` and not a plain `query`.

@sbegaudeau do you see what I'm doing wrong?

pcdavid:
PS: ignore the fact that in this version the handler hard-codes the computation of the breadcrumbs using EMF-specific code. This was just for the initial PoC.

PR #2913, opened 2024-01-11
[2910] Make the initial drop area larger and more visible on empty portals

### LINKED ISSUE #2910: Make the initial drop area larger and more visible on empty portals
Currently on an empty portal only a small and invisible zone at the top of the portal is valid to add/drop new representations. It's hard to discover for users.

[Capture vidéo du 2024-01-09 09-54-23.webm](https://github.com/eclipse-sirius/sirius-web/assets/10608/59afd1c6-3e93-4ce2-990b-050b611d72c0)

pcdavid:
It's better that the empty zone before, but I can't (yet) quite get an active zone which actually matches the full area (hence the visible background to clearly identify the limits of the active zone).

[Capture vidéo du 2024-01-11 16-38-43.webm](https://github.com/eclipse-sirius/sirius-web/assets/10608/d0551717-a0de-4732-aa16-b0a79212ba02)


frouene (approved):

frouene:
The first drop is now much more explicit for the user.
What can be confusing is that the drop ghost can exceed the area shown.
portal

pcdavid:
> The first drop is now much more explicit for the user. What can be confusing is that the drop ghost can exceed the area shown.

Should be fixed (or at least improved) by https://github.com/eclipse-sirius/sirius-web/pull/2936

pcdavid:
Failing e2e tests are unrelated.

PR #2924, opened 2024-01-15
[2909] Add support for read-only mode on portals

### LINKED ISSUE #2909: Add support for read-only on portals
Portals pass the `readOnly` prop to their embedded representations, but do not use it for themselves.
In read-only mode, it should not be possible to add or remove representations in a portal, nor to change their layout. It should no be possible to switch to the "Edit portal configuration" mode at all.

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/2909


frouene:
When a portal is in a portal, if the main portal is in edit mode, the child portal should be in read-only and should not allow resize
portal

frouene (approved):

frouene:
Failed tests are not related to the PR

PR #3034, opened 2024-01-29
[3000] Make portals with representations open in "direct" mode

### LINKED ISSUE #3000: Portals should open in "direct" mode when non-empty
When a portal is first created and is empty, it makes sense to default to the "edit" mode so that the user can add the appropriate representations and configure its layout.
However once this has been done, the actual _usage_ of a portal will be dominated by interactions with the representations it contains, with changes to the portal configuration itself only once in a while.
Thus when a portal is opened, it should be in "edit" mode if empty, but in "direct" mode otherwise.

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/3000


frouene (changes requested):
When adding an existing portal in a new portal, the existing portal is always in read-only mode
portal_default_mode

pcdavid:
> When adding an existing portal in a new portal, the existing portal is always in read-only mode

I *think* this is fixed in the new version.

frouene (approved):

pcdavid:
I've relaunched tests, there was 1 failure in `portal.cy.ts`. They all worked locally.

Issue #3036, opened 2024-01-30
IObjectServiceDelegate are not composable


pcdavid:
If two different extensions (E1 and E2) define their own custom `IObjectServiceDelegate`, they can not both work completely at the same time.

In `ComposedObjectService.getObject(IEditingContext, String)` we do:

```java
    @Override
    public Optional<Object> getObject(IEditingContext editingContext, String objectId) {
        var optionalDelegate = this.objectServiceDelegates.stream()
                .filter(delegate -> delegate.canHandle(editingContext))
                .findFirst();
        if (optionalDelegate.isPresent()) {
            return optionalDelegate.get().getObject(editingContext, objectId);
        }
        return this.defaultObjectService.getObject(editingContext, objectId);
    }
```

If E1 and E2 want to have any chance to get their `getObject(editingContext, objectId)` method called, they currently *must* return `true` to `canHandle(editingContext)`. However, when this is the case the "first" (on the order Spring will decide to inject them) which is found by `ComposedObjectService` will be the only one invoked. If it's E1, it can resolve its own objects and the defaults (by extending `DefaultObjectService`) but  will not know about potential other objects that E2 would be able to resolve, and E2 will never be called (until maybe the next reboot if it is injected before E1 this time).

One option would be for `IObjectServiceDelegate.canHandle(IEditingContext)` to also take the `String objectId` as it is the "guard method" for `getObject(IEditingContext editingContext, String objectId)` and `getContents(IEditingContext editingContext, String objectId)`. However this means that to answer this more precise question, `canHandle(IEditingContext, String)` some (most? all?) implementations will need to actully try to resolve the designated object just to return `true` and then be called again via `getObject(IEditingContext, String)` and re-do the same work. And a third and fourth time in the (relatively common) case where the next step will be to invoke `getContents()`:

1. canHandle(editinContext, objectId) => resolve the objectId, return true if found
2. getObject(editinContext, objectId) => resolve the objectId, this time to return it
3. canHandle(editinContext, objectId) => resolve the objectId, return true if found
4. getContents(editinContext, objectId) => resolve the objectId, this time to return its children

Another option would be to remove `IObjectServiceDelegate.canHandle(IEditingContext)`, leverage the fact that `IObjectServiceDelegate.getObject()` returns an optional and change `IObjectServiceDelegate.getContents()` to return an `Optional<List<Object>>` and return the first non-empty result from all the known delegates.

[...]

PR #3041, opened 2024-01-30
[fix] Fix TaskObjectServiceDelegate.canHandle(IEditingContext)

pcdavid:
The `IObjectServiceDelegate.canHandle(IEditingContext)` method is only used to guard the `getObject()` and `getContents()` methods.

`TaskObjectServiceDelegate` does not override these so it should not return `true` here as it prevents other delegates to be considered for these two methods.


pcdavid:
See https://github.com/eclipse-sirius/sirius-web/issues/3036 for more context and possible evolutions of this `IObjectServiceDelegate` API, but in the meantime this change should be more conform to the current API's intended usage.

AxelRICHARD (approved):

PR #3046, opened 2024-01-31
[doc] Add shape and ADR about the confirmation dialog on delete

AxelRICHARD:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
```diff
@@ -0,0 +1,52 @@
...
+All delete actions are executable in one click.
+Any accidental click on a delete action can lead to the unwanted loss of data.
+
+== Key Result
+
+No delete action should be done in one click.
+
+== Solution
+
+A confirmation dialog should be displayed when a delete action (click on the delete tool on a node, click on delete key on a node, click on the delete menu item on an object in the Explorer view) has been called.
```
mbats: Also when you use the DEL key in a diagram.

doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
```diff
@@ -0,0 +1,52 @@
...
+This dialog should avoid or drastically minimized unwanted loss of data.
+
+As a specifier, It should be possible to enable/disable this capability on every representation description (Explorer view, any diagram description...).
```
mbats: It should be possible to be enabled/disabled by the end-user on a project.

doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+== Scenario
+
+=== Breadboarding
+
+The confirmation dialog could look like this:
+
+image:images/delete_dialog/delete_dialog.svg[Delete dialog]
+
+No cancel button will be provided, to be consistent with other Sirius Web dialogs.
```
mbats: I think here the Cancel button is important. If you click out of the window it will cancel the action and if you click on the button it will cancel also.

doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
```diff
@@ -0,0 +1,52 @@
...
+A click outside the confirmation dialog will cancel the deletion.
+
+For the specific cas of diagrams tools, the confirmation dialog will be the same, adapted with the message and button label set by the specifier.
```
mbats: cas -> case

doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
```diff
@@ -0,0 +1,52 @@
...
+
+=== Cutting backs
```
mbats, line 41: You should be able also to see the consequences of the deletion (x elements will be deleted. Are your sure?)

doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
AxelRICHARD: Yes it is written in the paragraph: "click on delete key on a node". I will update the message to "click on DEL key on a node" to be more explicit.

doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
AxelRICHARD: Ok, in place of the capability to enable/disable on every representation description, or in addition to?

If it is in addition to, I think it will lead to complex cases, as for example when the specifier disabled it on a diagram description, but the user enabled it on a project, and so who won?

doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
AxelRICHARD: Ok, I will update it with a cancel button.

doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
AxelRICHARD: done

doc/iterations/2024.3/shapes/confirmation_dialog_on_delete.adoc
AxelRICHARD: After discussion with @mbats, the solution will be in place of the capability to enable/disable on every representation description.
The end-user will not enable/disable the capability on the project, but on the dialog itself.

doc/adrs/128_add_support_for_confirmation_dialogs.adoc
```diff
@@ -0,0 +1,53 @@
...
+All delete actions are executable in one click.
+Any accidental click on a delete action can lead to the unwanted loss of data.
+By extension, some one-click actions may lead to unwanted consequences.
+
+== Decision
+
+The first decision is about the delete actions (click on the delete tool on a node, click on DEL key on a node, click on the delete menu item on an object in the Explorer view).
+For these actions, a confirmation dialog will be added, asking to confirm the deletion.
+It will be possible to disable these dialogs by clicking on the "Don't ask me again" checkbox inside them.
+These checkboxes will be associated to cookies, allowing to remember which dialogs have been disabled or not, for each end-users.
```
pcdavid: Maybe use local storage instead of cookies? Cookies are useful when we need to send back some id/data along with a backend request, but here the decision to show the decision to show the dialog or not is purely frontend.

doc/adrs/128_add_support_for_confirmation_dialogs.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+A new `Confirmation` concept will be added into `Tool` interface in `diagram.graphqls`:
+
+```
+interface Tool {
+  id: ID!
+  label: String!
+  iconURL: [String!]!
+  confirmation: Confirmation!
```
sbegaudeau: Why not make the confirmation optional instead of making it compulsory but with an `enable` property? It would feel quite cumbersome to have to compute a non existant confirmation dialog for every single tool in a diagram. Just make it optional like the `selectionDescriptionId`.

I would also directly use the name `ConfirmationDialog` in the GraphQL API too but that's a details.

doc/adrs/128_add_support_for_confirmation_dialogs.adoc
```diff
@@ -0,0 +1,53 @@
...
+}
+
+interface Confirmation {
+  id: ID!
+  enable: Boolean! // if the confirmation dialog (or any other graphical element) should be enabled on this tool
```
sbegaudeau: Could be deleted

doc/adrs/128_add_support_for_confirmation_dialogs.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+The second decision is about the diagrams tools.
+A confirmation dialog mechanism will be available on any tool in diagrams, not only the delete tool.
+Thus, a specifier could associate a confirmation dialog on the execution of its tools. This capability will be available on the Diagram View DSL.
+On a tool description, the specifier could enable the confirmation dialog and set a message expression and a button label expression (i.e. AQL expressions), allowing to call the confirmation dialog when the tool is executed by end-users.
+
+
+== Consequences
+
+To enable a disabled confirmation dialog, the end-user will have to delete the associated cookie.
```
sbegaudeau: (local storage?), it's pretty much the same for the end user

doc/adrs/128_add_support_for_confirmation_dialogs.adoc
```diff
@@ -0,0 +1,53 @@
...
+  id: ID!
+  enable: Boolean! // if the confirmation dialog (or any other graphical element) should be enabled on this tool
+  message: String! // the message that will be displayed on confirmation
+  label: String! // the label that will be displayed on the confirmation button (or any other graphical element used to confirm the execution of the tool)
+}
+```
+
+On frontend side, a new react component will be created for the confirmation dialog.
+```
+<ConfirmationDialog message={message} label={label} onConfirm={// execution of the tool}/>
```
sbegaudeau: I think you would need to have more than that, for example the id of the confirmation dialog to know which dialog to ignore (you would probably end up with `GQLConfirmationDialog` directly as a props).

It does not matter right now in this ADR but since you will end up using this dialog at multiple location with its internal lifecycle (to open it or not and then running the confirmation directly instead depending on the local storage state), you may end up with a quite different API than this).

The current example visible here is fine for an ADR, no need to change anything.

doc/adrs/128_add_support_for_confirmation_dialogs.adoc
sbegaudeau: I agree, local storage would indeed be better 👍

sbegaudeau (commented):
This looks good, I've added a couple minor comments

sbegaudeau (approved):

PR #3049, opened 2024-02-01
[doc] Add shape and ADR about filters in Explorer view

AxelRICHARD:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


doc/adrs/129_add_support_for_filters_in_explorer_view.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+On the backend side, a filter is an AQL expression.
+This will allow to get the filter concept to be used in other contexts such as diagram for example (If we decide to add filters on diagrams).
+Depending on the context, variables will be accessible in the AQL expression.
+
+```java
+public interface IFilter {
+  String getLabel(); // the label that will be displayed to end-users to identify this filter
+  boolean canHandle(IEditingContext editingContext); // allows to propose this filter to end-users depending on the editing context
+  boolean filter(String predicateExpression); // what the filter filters
```
sbegaudeau: I think an API like this one would be better `boolean filter(Object object);`
The content of the filter may have an expression, if it uses the view DSL but what is being filtered is some object from the editing context or a representation or something else.

doc/adrs/129_add_support_for_filters_in_explorer_view.adoc
```diff
@@ -0,0 +1,67 @@
...
+  boolean initialState(); // the initial state of the filter: enabled or disabled.
```
sbegaudeau: I'm not sure where this API is supposed to be but I don't know why it should be stateful. Let's say that the system provides a set of filters to be applied anywhere we want (forms, explorer, diagrams, diagram elements like edge and node, etc). They have a label to be displayed to the end user and they can filter some objects.

Someone can define a filter using an API like that:

```
@Service
public RandomFilter implements IFilter {
    public String getLabel() {
        return "Random filter";
    }

    public boolean canHandle(IEditingContext editingContext) {
        return true;
    }

    public boolean filter(Object object) {
        return new Random().nextBoolean();
    }
}
```

Then in the tree representation we would have to both:

- return the list of available filters
- apply a filter in the representation, in this case the representation, not the person creating the filter would have to know that the filter has been applied. We would need a unique identifier too for the filter. every representation supporting filters would have to remember what has been applied (just like the diagram remembers what has been hidden or faded)

Once applied, the representation would be refreshed with the filter to hide elements.

doc/adrs/129_add_support_for_filters_in_explorer_view.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+```
+type EditingContext {
+  ...
+  filters: [Filter!]!
+}
+
+type Filter {
+  label: String!
+}
```
sbegaudeau: A unique identifier would be helpful to send back to the backend and as a key in React components. A label does not carry this intent of uniqueness.

doc/adrs/129_add_support_for_filters_in_explorer_view.adoc
```diff
@@ -0,0 +1,67 @@
...
+```
+
+In `tree.graphqls`, the tree item will be enriched with a new filter state:
+
+```
+type TreeItem {
+  ...
+  filtered: Boolean!
+}
+```
```
sbegaudeau: I don't think this concept is necessary filtered tree items could just not be sent to the frontend.

doc/iterations/2024.3/shapes/add_filters_on_explorer_view.adoc
```diff
@@ -0,0 +1,44 @@
...
+== Key Result
+
+End-users have the possibility to hide objects contained in trees in the Explorer view.
+
+== Solution
+
+Filters action should be available in the Explorer view toolbar, allowing to hide objects contained the trees in the Explorer view.
+The filters action is a menu, containing a list of filters.
+A filter menu item is a checkbox associated to a label.
+When the checkbox is enabled, the filter is enabled on all trees in the Explorer view.
```
sbegaudeau: There is only one tree in the explorer view (same comment for the next line)

doc/iterations/2024.3/shapes/add_filters_on_explorer_view.adoc
```diff
@@ -0,0 +1,44 @@
...
+When the checkbox is disabled, the filter is disabled on all trees in the Explorer view.
+
+== Scenario
+
+End-users want to hide all elements of some type, or based on some condition.
+The filter could lead to hide an element and all its children, recursively, or not.
+Depending on the filter, an element could be hidden, but not its children.
```
sbegaudeau, line 25: I would definitely put that in a cutting back because it brings so many questions that it could be a shape + adr on its own. It could add so much complexity depending on how we see that, that we would need additional discussion on this part (while the first part is quite straightforward). It explains why you may have wanted to send the filter to the frontend. I'm not sure that it's necessary at all but I would suggest that a first version should just make tree items and their children disappear (stop considering them in getElements / getChildren at all in the backend) since that's easy to do.

After that, we should talk about what filtering can mean because the API in the ADR is just considering a filter returning true/false and it may need to be way more complex than just that. We would need to talk with a real use case for that. And a simple API with true / false may even be enough given the right meaning or not.

If we have the following tree:

```
- A
  - B
    - C
    - D
  - E
    - F
      - G
        - H
          - C
            - B
```

Yes, I know the structure of the tree does not match a topological structure of EObject, that's 100% on purpose, we don't have to follow containment references all the time. Consumers of Sirius Components do not always do that.

- What would happen if I say that I filter `B`?

This (B is hidden and C and D becomes children of A):

```
- A
  - C
  - D
  - E
    - F
      - G
        - H
          - C
```

or that (B and its children disappear):

```
- A
  - E
    - F
      - G
        - H
          - C
```

Personally, I would say that both approaches have flaws. Using the first meaning (which would make the object disappear but not its children, I would have to explicitly filter C and D by myself but it would make C disappear from under H which may not be what I want. On the other hand, if filtering B means filtering its children too, then it's an issue too because I may not want the children to disappear at all.

So, if those two sentences are a core feature that you need, I would even suggest that we don't need filters at all but instead transformations of the tree because I don't think that a boolean can tell us how to transform the shape of an arbitrary tree in a satisfying manner for all specifiers.

We will need to discuss more about that.

doc/iterations/2024.3/shapes/add_filters_on_explorer_view.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+Filters action should be available in the Explorer view toolbar, allowing to hide objects contained the trees in the Explorer view.
+The filters action is a menu, containing a list of filters.
+A filter menu item is a checkbox associated to a label.
+When the checkbox is enabled, the filter is enabled on all trees in the Explorer view.
+When the checkbox is disabled, the filter is disabled on all trees in the Explorer view.
+
+== Scenario
+
+End-users want to hide all elements of some type, or based on some condition.
```
sbegaudeau: Small nitpick but it can be misinterpreted since I know some users would like something even more powerful. I would personally say something like "specifiers want to let end users filter elements using some predefined filters" because the end users does not get to decide anything except applying or not the filters. 

They may not understand anything to the condition or type of what is being filtered and there no mention of some user-specified filter from what I've seen here.

doc/iterations/2024.3/shapes/add_filters_on_explorer_view.adoc
```diff
@@ -0,0 +1,44 @@
...
+=== Breadboarding
+
+The filters action in the Explorer view could look like this:
+
+image:images/explorer_view_filters/explorer_view_filters.svg[Explorer view filters]
+
+== Cutting backs
+
+Nothing will be done to adapt the selection synchronization with the filters.
+If an element is hidden in the Explorer view because a filter hide it, then if the sam element is selected in a diagram, then it will not be revealed/displayed in the Explorer view.
```
sbegaudeau: sam -> same

sbegaudeau: I don't think that there anything to do, if an element is not in a representation, its selection is just ignored by the representation. Just like if you select in a diagram something that does not appear in the explorer or the other way around.

sbegaudeau (changes requested):

doc/iterations/2024.3/shapes/add_filters_on_explorer_view.adoc
sbegaudeau: And if we go with transformations of the tree instead of filtering, I'm not even sure that `transformation1(transformation2(tree))` would be equal to `transformation2(transformation1(tree))` so the order in which those transformations would be applied could matter.

doc/adrs/129_add_support_for_filters_in_explorer_view.adoc
AxelRICHARD: done

doc/adrs/129_add_support_for_filters_in_explorer_view.adoc
AxelRICHARD: done

doc/iterations/2024.3/shapes/add_filters_on_explorer_view.adoc
AxelRICHARD: I wanted to say each document in the Explorer view, which are the root of their own tree.
I will update this sentence to be more explicit.

doc/iterations/2024.3/shapes/add_filters_on_explorer_view.adoc
AxelRICHARD: done

doc/iterations/2024.3/shapes/add_filters_on_explorer_view.adoc
AxelRICHARD: Yes I also think there is nothing to do. I just wanted to be clear that nothing will be done here about the synchronization.
I will add a sentence to be more clear.

doc/iterations/2024.3/shapes/add_filters_on_explorer_view.adoc
AxelRICHARD: Indeed, we will need to discuss more about that.
I definitely want the first approach.

sbegaudeau (approved):

PR #3062, opened 2024-02-05
[3061] Re-resolve the current self from its id when refreshing a Form

### LINKED ISSUE #3061: Forms always render the same element
In `FormEventProcessor.refreshForm()`, we currently do:
```java
VariableManager variableManager = new VariableManager();
variableManager.put(VariableManager.SELF, this.formCreationParameters.getObject());
```
In other words, the `self` element on which the `FormRenderer` will be invoked will always be the initial object passed when the form was initialy created.

This works "by mistake" in the standard case where this `self` object is a mutable `EObject`: each render sees the same object, but each time in the correct/expected state.

However this breaks when the target object is immutable, and a new, different Java object is used to represent the successive states of that object (with the same id). For example when [creating a Form on the content of a Portal representation](#2890). In such a case, the `FormEventProcessor` will always render the same `Form` on every refresh as the original object from `formCreationParameters` is immutable.

Diagrams solve this (not sure that was exactly on purpose) by re-resolving the target object from its id on every refresh:
```java
var optionalObject = this.objectService.getObject(editingContext, previousDiagram.getTargetObjectId());
```

Forms (and other representations which may have the issue) should do the same.

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/3061


sbegaudeau (approved):
This looks good, I'll wait for the build to merge it

PR #3064, opened 2024-02-05
[3063] Re-resolve the current self from its id when refreshing a Selection

### LINKED ISSUE #3063: The Selection representation always render the same element
See https://github.com/eclipse-sirius/sirius-web/issues/3061

`SelectionEventProcessor` resolves its `selectionConfiguration.getTargetObjectId()` only once in `SelectionEventProcessorFactory.createRepresentationEventProcessor()` and then reuses the same Java `Object` for all refreshes in `SelectionEventProcessor`.

Not an issue in practice when these are mutable EObjects, each refresh sees the current state, but in general this can be a problem when the target object it immutable.
Other representations avoid this by re-resolving the target object from its id on each refresh to make sure they see the current version.

pcdavid:


sbegaudeau (approved):
Once the build will be done, I'll merge it

PR #3068, opened 2024-02-06
[3066] Add support for representations in DefaultObjectService

### LINKED ISSUE #3066: Add support for representations in DefaultObjectService
I'm not entirely sure this is the right approach, but for #2890 I need to avoid the current behavior where a representation which uses another representation as a semantic element is considered dangling/obsolete as soon as it is created, and thus disposed immediately.

One approach outlined in #3023 would be to handle specifically this behavior in by changing how `IDanglingRepresentationDeletionService` works. This would fix this particular unwanted behavior but feels ad-hoc.

Another approach, implemented here, is to consider at the _core_ of the framework, that representations can be considered as "objects" (as seen by `IObjectService`).
This would also fix the issue mentioned above, and remove the need for other "hacks" (i.e. in `RepresentationsEventProcessorFactory`, resolving the `targetObjectId` with either `IObjectService` _or_, explicitly, with `IRepresentationSearchService`).
Conceptually it also feels like a more "satisfying" approach as it removes a disctinction between "semantic elements" on one side and "representations" on the other.
However because it changes a rather fundamental assumption in the system, it can have some hard to predict side effects.

pcdavid:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/3066

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -20,6 +20,7 @@
 - [ADR-132] Add support for filters in Explorer View
 - [ADR-133] Add Tree Widget to the View DSL
 - [ADR-134] Split IObjectService in multiples sub services
+- [ADR-135] Add support for representations in object related services
 
 === Breaking changes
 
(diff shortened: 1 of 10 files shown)


sbegaudeau (approved):
Let's see what happens now 😄

PR #3073, opened 2024-02-06
[doc] Contribute a shape for initial Search support

pcdavid:

### DIFF
--- a/doc/iterations/2024.3/shapes/search.adoc
+++ b/doc/iterations/2024.3/shapes/search.adoc
@@ -1,0 +1,110 @@
+= (L) Add initial support for search
+
+== Problem
+
+Currently the only way a user to find elements of interest inside in his documents is to manually look into the explorer.
+
+For example he can "Expand all" from the root of all his documents and then use 'Ctrl-F' to open the filter, but:
+
+* this does not scale well for large models;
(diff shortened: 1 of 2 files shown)


doc/iterations/2024.3/shapes/search.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+Currently the only way a user to find elements of interest inside in his documents is to manually look into the explorer.
+
+For example he can "Expand all" from the root of all his documents and then use 'Ctrl-F' to open the filter, but:
+
+* this does not scale well for large models;
+* this only allows basic filtering/search on the element's labels (the only thing known to the frontend).
+
+We should provide support for search.
+This first version will keep things simple with a default implementation in Sirius Web Sample Application.
```
sbegaudeau, line 13: I'm pretty sure that we will need that in both versions of Sirius Web so if a sizable part of the code is Sirius Web specific, it can either affect the cost of the feature or a choice will have to be made on the supported runtime.

sbegaudeau (approved):
I agree with pretty much everything here but since I don't remember how much time we want to invest on this topic, I suspect that we may have to sacrifice some things to get a first version out.

I've approved the shape to let you merge it if you want but I'll let Mélanie have a look.

PR #3098, opened 2024-02-12
[doc] Contribute a shape to support different overflow label strategies

frouene:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
```diff
@@ -0,0 +1,61 @@
...
+== Key Result
+
+Allow the specifier to define a label that applies an auto wrap rather than enlarging the node.
+A label should be broken at any point if there are no otherwise-acceptable break points (like a long word or URL).
+
+== Solution
+
+Add to the view DSL, for inside label:
+
+* _maxLabelWidthExpression_
```
sbegaudeau: Can we easily provide such option? I would personally put it in a cutting back. We would also need a max line height too. In both case, we should probably express this using characters something like max number of character horizontally and max number of lines.

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
```diff
@@ -0,0 +1,61 @@
...
+** If the label is larger than the node default width but smaller than the _maxLabelWidthExpression_ result, then the node will increase to display the whole label
++
+image:images/auto_wrap_label/node_with_normal_label.png[label smaller than the max label length]
++
+** If the label is larger than the _maxLabelWidthExpression_ result, then the node will be sized to match the _maxLabelWidthExpression_ result and the _overflowStrategy_ will be
+ applied on the label
++
+*** _none_ option
++
+image:images/auto_wrap_label/overflowStrategy_none_option.png[None option]
```
sbegaudeau, line 52: This is considered as a bug by our users, if an inside label is not supposed to overflow from its parent node, then the size of the node has to be updated.

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
```diff
@@ -0,0 +1,61 @@
...
+== Problem
+
+In the existing node layouts, the inside label of a node will always be used to determine the minimum size.
+This means that in the case of a long label, the node will necessarily take a size that allows the label to be displayed without any modification.
+In some cases, specifier doesn't want that the label affects the node representation.
+
+== Key Result
+
+Allow the specifier to define a label that applies an auto wrap rather than enlarging the node.
+A label should be broken at any point if there are no otherwise-acceptable break points (like a long word or URL).
```
sbegaudeau, line 12: I would very much like to let CSS deal with that and not try anything funny on this topic at all, we tried once already. CSS already provides various options for that.

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
```diff
@@ -0,0 +1,61 @@
...
+== Solution
+
+Add to the view DSL, for inside label:
+
+* _maxLabelWidthExpression_
+* _overflowStrategy_: none, wrap or ellipsis
+
+If set, a label can only influence the size of its node up to _maxLabelWidthExpression_ result.
+There are two options for dealing with overflow situations, wrap text or display an ellipsis.
+With the _overflowStrategy_ _none_ option, the
```
sbegaudeau: The end of the sentence is missing :)

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
```diff
@@ -0,0 +1,61 @@
+= Overflow label strategy
+
+== Problem
+
+In the existing node layouts, the inside label of a node will always be used to determine the minimum size.
```
sbegaudeau, line 5: What about outside label? Should we just exclude them? I'm not opposed to that, but we could have an outside label of thousands of words...

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
```diff
@@ -0,0 +1,61 @@
...
+** If the label is larger than the _maxLabelWidthExpression_ result, then the node will be sized to match the _maxLabelWidthExpression_ result and the _overflowStrategy_ will be
+ applied on the label
++
+*** _none_ option
++
+image:images/auto_wrap_label/overflowStrategy_none_option.png[None option]
++
+*** _wrap_ option
++
+image:images/auto_wrap_label/overflowStrategy_wrap_option.png[Wrap option]
```
sbegaudeau, line 56: Let's say I don't have a default width, can I resize this node to be a 5px by 5px one? Since the size of the label does not influence anything. What happens if I start the direct edition of the label?

We may say that we don't want to do anything for such use case but we have to write strict rules to make it explicit.

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+Allow the specifier to define a label that applies an auto wrap rather than enlarging the node.
+A label should be broken at any point if there are no otherwise-acceptable break points (like a long word or URL).
+
+== Solution
+
+Add to the view DSL, for inside label:
+
+* _maxLabelWidthExpression_
+* _overflowStrategy_: none, wrap or ellipsis
```
sbegaudeau, line 18: Why not wrap+ellipsis?

- none: The node is horizontally resized (but not vertically) to display its label
- wrap: The node is not resized horizontally, the label is wrapped at the end of the line. The node can be resized vertically by its label.
- ellipsis: The node is not resized horizontally or vertically, the label is ending with an ellipsis at the end of the line
- wrap+ellipsis: The label is wrapped at the end of the line but it does not change the size of the node vertically either. An ellipsis is displayed if the node is full.

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
```diff
@@ -0,0 +1,61 @@
+= Overflow label strategy
+
+== Problem
+
+In the existing node layouts, the inside label of a node will always be used to determine the minimum size.
+This means that in the case of a long label, the node will necessarily take a size that allows the label to be displayed without any modification.
+In some cases, specifier doesn't want that the label affects the node representation.
```
sbegaudeau, line 7: Honestly I'm not sure that this should even be configurable at all. It may just be me but as an end user, I would find it very confusing if some labels behaved differently from others and as a specifier, I would find it very painful to have to configure every label manually.

I would personally put wrap and ellipsis everywhere and let the choice of the end user regarding the size of the elements be respected. If an end user creates a node of 50x50 nicely positioned next to other nodes, I would find it quite odd to break all the layout and move everything because its label would be too large.

But it may be just me...

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
frouene: This is a CSS option: `break-word`

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
frouene: I changed it to the current behavior, the node is expands to fit the label

doc/iterations/2024.3/shapes/overflow_label_strategy.adoc
frouene: For inside label, the maximum label width is defined by the node width, but for outside label we don't have such constraints.

sbegaudeau (approved):

PR #3099, opened 2024-02-13
[2999] Add support for read-only in FormDescriptionEditors

### LINKED ISSUE #2999: FormDescriptionEditor does not support read-only mode
Ok, I was trying something a bit fishy here but it works, it just shows a tiny error in the process. If that's the source of the issue, we could probably deactivate the drop area of the portal while we are not editing the representations? @pcdavid what do you think?

https://www.youtube.com/watch?v=jH35IwkkRbo

Anytime I try to drag something inside the form description editor, the portal starts to react which prevents it from having a great user experience. On the screenshot below, I'm trying to drop a checkbox in the form description editor and the portal wants to capture the drop which creates an error too:

Capture d’écran 2024-01-24 à 14 21 30

I'm starting to believe that we are "this close" from being able to drag and drop things between representations without issues 😛

pcdavid:


packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/hooks/FormDescriptionEditorContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+import React from 'react';
+import { GQLFormDescriptionEditor } from '../FormDescriptionEditorEventFragment.types';
+import { FormDescriptionEditorContextValue } from './useFormDescriptionEditor.types';
+
+export const FormDescriptionEditorContext = React.createContext<FormDescriptionEditorContextValue>(null!);
```
AxelRICHARD: Why do you add a postfix ! on a null value?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/hooks/FormDescriptionEditorContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type FormDescriptionEditorContextProviderProps = {
```
AxelRICHARD: Shouldn't the props type be exported and extracted in a *.types.ts file?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/WidgetEntry.tsx
```diff
@@ -315,28 +309,35 @@ export const WidgetEntry = ({
...
+    );
   } else if (widget.__typename === 'SplitButton') {
     widgetElement = (
       <SplitButtonWidget
         data-testid={widget.id}
         editingContextId={editingContextId}
         representationId={representationId}
         widget={widget as GQLSplitButton}
-        onDropBefore={onDropBefore}
+        onDropBefore={readOnly ? noop : onDropBefore}
```
AxelRICHARD, line 326: May be it would have been easier to modify `onDropBefore` instead on modify every call to it by `readOnly ? noop : onDropBefore`.
What do you think about it?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/ToolbarActions.tsx
```diff
@@ -186,14 +180,14 @@ export const ToolbarActions = ({
       <div
         data-testid={`ToolbarActions-DropArea-${containerId}`}
         className={classes.toolbarActionDropArea}
-        onDragEnter={handleDragEnter}
-        onDragOver={handleDragOver}
-        onDragLeave={handleDragLeave}
-        onDrop={handleDrop}
+        onDragEnter={readOnly ? noop : handleDragEnter}
```
AxelRICHARD, line 183: May be it would have been better/more readable to modify handleDragEnter/handleDragOver/handleDragLeave... instead.
What do you think about it?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/ToolbarActionWidget.tsx
```diff
@@ -350,12 +347,16 @@ export const ToolbarActionWidget = ({
       <div
         data-testid={`ToolbarAction-DropArea-${toolbarAction.id}`}
         className={classes.placeholder}
-        onDragEnter={handleDragEnter}
-        onDragOver={handleDragOver}
-        onDragLeave={handleDragLeave}
-        onDrop={handleDrop}
+        onDragEnter={readOnly ? noop : handleDragEnter}
```
AxelRICHARD, line 350: May be it would have been better/more readable to modify handleDragEnter/handleDragOver/handleDragLeave... instead.
What do you think about it?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/SplitButtonWidget.tsx
```diff
@@ -360,10 +363,10 @@ export const SplitButtonWidget = ({ widget, editingContextId, representationId }
       <div
         data-testid={`${widget.__typename}-Widgets-DropArea-${widget.id}`}
         className={classes.bottomDropArea}
-        onDragEnter={handleDragEnter}
-        onDragOver={handleDragOver}
-        onDragLeave={handleDragLeave}
-        onDrop={handleDrop}>
+        onDragEnter={readOnly ? noop : handleDragEnter}
```
AxelRICHARD, line 366: May be it would have been better/more readable to modify handleDragEnter/handleDragOver/handleDragLeave... instead.
What do you think about it?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/PageList.tsx
```diff
@@ -395,35 +396,30 @@ export const PageList = ({ editingContextId, representationId, formDescriptionEd
...
-                draggable={true}
-                onDragStart={handleDragStart}
-                onDragEnter={handleDragEnter}
-                onDragOver={handleDragOver}
-                onDragLeave={handleDragLeave}
-                onDrop={handleDropTab}
+                onClick={readOnly ? noop : handleClick}
+                onKeyDown={readOnly ? noop : handleDelete}
+                draggable={!readOnly}
+                onDragStart={readOnly ? noop : handleDragStart}
```
AxelRICHARD, line 402: May be it would have been better/more readable to modify handleDragEnter/handleDragOver/handleDragLeave... instead.
What do you think about it?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/Page.tsx
```diff
@@ -188,25 +191,16 @@ export const Page = ({ editingContextId, representationId, formDescriptionEditor
...
         })}
       </div>
       <div
         data-testid="Page-DropArea"
         className={classes.bottomDropArea}
-        onDragEnter={handleDragEnter}
-        onDragOver={handleDragOver}
-        onDragLeave={handleDragLeave}
-        onDrop={handleDrop}>
+        onDragEnter={readOnly ? noop : handleDragEnter}
```
AxelRICHARD, line 200: May be it would have been better/more readable to modify handleDragEnter/handleDragOver/handleDragLeave... instead.
What do you think about it?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/Group.tsx
```diff
@@ -466,19 +466,19 @@ export const Group = ({ editingContextId, representationId, formDescriptionEdito
       <div
         data-testid={`Group-DropArea-${group.id}`}
         className={classes.placeholder}
-        onDragEnter={handleDragEnter}
-        onDragOver={handleDragOver}
-        onDragLeave={handleDragLeave}
-        onDrop={handleDrop}
+        onDragEnter={readOnly ? noop : handleDragEnter}
```
AxelRICHARD, line 469: May be it would have been better/more readable to modify handleDragEnter/handleDragOver/handleDragLeave... instead.
What do you think about it?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/FormDescriptionEditorRepresentation.tsx
```diff
@@ -141,11 +132,19 @@ const useFormDescriptionEditorStyles = makeStyles((theme) => ({
   },
 }));
 
+export type WidgetDescriptor = {
```
AxelRICHARD: WidgetDescriptor seems declared here, and also in FormDescriptionEditorRepresentation.types.ts

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/coreWidgets.tsx
```diff
@@ -0,0 +1,101 @@
...
+    name: 'Tree',
+    icon: <AccountTreeIcon />,
+  },
+  {
+    name: 'SplitButton',
+    label: 'Split Button',
+    icon: <Button width={'24px'} height={'24px'} color={'secondary'} />,
+  },
+  {
+    name: 'TextArea',
```
AxelRICHARD, line 95: Could you add `label: 'Textarea'` please?

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/coreWidgets.tsx
```diff
@@ -0,0 +1,101 @@
...
+import RadioButtonCheckedIcon from '@material-ui/icons/RadioButtonChecked';
+import TextFieldsIcon from '@material-ui/icons/TextFields';
+import TextFormatIcon from '@material-ui/icons/TextFormat';
+import ViewColumnIcon from '@material-ui/icons/ViewColumn';
+import { WidgetDescriptor } from './FormDescriptionEditorRepresentation';
+import { Button } from './icons/Button';
+
+export const coreWidgets: WidgetDescriptor[] = [
+  {
+    name: 'BarChart',
```
AxelRICHARD, line 31: Could you add `label: 'Bar Chart'` please?

pcdavid: done.

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/coreWidgets.tsx
pcdavid: done.

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/FormDescriptionEditorRepresentation.tsx
pcdavid: done.

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/FormDescriptionEditorRepresentation.types.ts
AxelRICHARD, line 1: Could you add SplitButton and Tree in this list please? It seems they have been forgotten by the one who added them.

pcdavid: done.

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/WidgetEntry.tsx
pcdavid: I hesitated between the two approaches but felt this way made things more obvious as it makes the "disabled when in read-only" concern handled in one place where the widget is configured instead of hidden in every individual handler. I felt it would be easier to forget supporting read-only mode if every handler has to handle it, especially if we add new handlers later on.

I don't have strong feelings or hard data about this, if you believe the other way would be better I can change.

AxelRICHARD: Ok let's keep that approach then

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/hooks/FormDescriptionEditorContext.tsx
pcdavid: React.createContext *needs* a value, but at the time this is defined there is no meaningful value `FormDescriptionEditorContextValue` that I can give it.
The only options I see are:
* use `React.createContext<FormDescriptionEditorContextValue | undefined>` but then all users have to test for `undefined`
* set a non-null default `FormDescriptionEditorContextValue` value which is actually completely meaningless but type-checks:
```js
const defaultValue: FormDescriptionEditorContextValue = {
  editingContextId: '',
  representationId: '',
  readOnly: false,
  formDescriptionEditor: { id: '', pages: [] },
};
```
* use this "trick" to bypass the null-check just for this location.

I can revert to the second option if you prefer, it's what we have in e.g. `DiagramContext` or `DeckContext`. It just felt uglier in this can because of the additional level of "broken but makes TS happy" for `formDescriptionEditor`.

packages/formdescriptioneditors/frontend/sirius-components-formdescriptioneditors/src/hooks/FormDescriptionEditorContext.tsx
pcdavid: Done.

AxelRICHARD (approved):

PR #3123, opened 2024-02-16
[3096] Add support for filters in Explorer view

### LINKED ISSUE #3096: Add support for filters in Explorer View

### Expected behavior

Filters action should be available in the Explorer view toolbar, allowing to hide objects contained the trees in the Explorer view.
The filters action is a menu, containing a list of filters.
A filter menu item is a checkbox associated to a label.
When the checkbox is enabled, the filter is enabled on all objects in the Explorer view.
When the checkbox is disabled, the filter is disabled on all objects in the Explorer view.

### Cuttings backs

...

### Potential side effects

...

...

### LINKED ISSUE #3102: Make support for transient representations more explicit
= ADR-XXX - Make support for transient representations more explicit

== Context

Sirius Web contains a set of transient representations which are not as connected to the Sirius Components architecture as persistent representations.
As a result, it is difficult to obtain both the metadata and the description of a transient representation.

== Decision

We should be able to find the description of any transient representation.
For that, we will introduce a dedicated service to find them.

```
package org.eclipse.sirius.web.services.api.representations;

public interface ITransientRepresentationMetadataSearchService {
    Optional<RepresentationMetadata> findById(String editingContextId, String representationId);
}
```

```
package org.eclipse.sirius.web.services.representations;

@Service
public class TransientRepresentationMetadataSearchService implements ITransientRepresentationMetadataSearchService {
    public Optional<RepresentationMetadata> findById(String editingContextId, String representationId) {
        // hardcoded metadata for explorer, validation, details, representations, related elements
    }
}
```

We could then use this service in the `EditingContextRepresentationDataFetcher` in order to retrieve the metadata of the transient representations.

After that, we should be able to find the description of said transient representations in `EditingContext#getRepresentationDescriptions` as for any representation descriptions. For that, the editing context should be populated with the various representation descriptions when it is loaded.

== Status

Work in progress

== Consequences

We should update the behavior of `EditingContextRepresentationDataFetcher` in the future to extract it in a dedicated service `IRepresentationMetadataSearchService` which should look in either the `ITransientRepresentationMetadataSearchService` or in the database to find the metadata. I don't see why we should look in the editing context event processor with both solutions, as it is inefficient performance wise and can provide unsynchronized information.

AxelRICHARD:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/3096

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management
- [x] Have the relevant API breaks been described in the `CHANGELOG.adoc`? (Including changes in the GraphQL API)
- [x] In case of a change with a visual impact, are there any screenshots in the `CHANGELOG.adoc`? For example in `doc/screenshots/2022.5.0-my-new-feature.png`

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:
	- [x] `useQuery<DATA_TYPE, VARIABLE_TYPE>(…)`
	- [x] `useState<STATE_TYPE>(…)`
- [x] No useless optional chaining with `?.` (if the GraphQL API specifies that a field cannot be `null`, do not treat it has potentially `null` for example)
- [x] Nullable values have a proper type (for example `let diagram: Diagram | null = null;`)

## Backend

### General purpose

### Architecture

### How to test this PR?


packages/sirius-web/backend/sirius-web-graphql/src/main/java/org/eclipse/sirius/web/graphql/datafetchers/editingcontext/EditingContextRepresentationDataFetcher.java
```diff
@@ -77,9 +80,9 @@ public DataFetcherResult<RepresentationMetadata> get(DataFetchingEnvironment env
                                                           representation.getDescriptionId());
                     })
                     .findFirst();
-        // @formatter:on
-
-        // @formatter:off
+        if (representationMetadata.isEmpty()) {
+            representationMetadata = this.transientRepresentationMetadataSearchService.findTransientRepresentationById(representationId);
+        }
```
sbegaudeau, line 85: If it works, this should be in replacement of the previous block

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideRepresentationsTreeItemAlteredContentProvider.java
```diff
@@ -0,0 +1,43 @@
...
+    public boolean canHandle(Object object, List<String> activeFilterIds) {
+        if (activeFilterIds.contains(SampleTreeFilterProvider.HIDE_REPRESENTATIONS_TREE_FILTER_ID)) {
+            return true;
+        }
+        return false;
+    }
+
+    @Override
+    public List<Object> apply(List<Object> computedChildren, VariableManager variableManager) {
+        computedChildren.removeIf(child -> child instanceof RepresentationMetadata);
```
sbegaudeau: Are you supposed to modify the input list?

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideRepresentationsTreeItemAlteredContentProvider.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+@Service
+public class HideRepresentationsTreeItemAlteredContentProvider implements IExplorerTreeItemAlteredContentProvider {
+
+    @Override
+    public boolean canHandle(Object object, List<String> activeFilterIds) {
+        if (activeFilterIds.contains(SampleTreeFilterProvider.HIDE_REPRESENTATIONS_TREE_FILTER_ID)) {
+            return true;
+        }
+        return false;
```
sbegaudeau: `return activeFilterIds.contains(SampleTreeFilterProvider.HIDE_REPRESENTATIONS_TREE_FILTER_ID);`

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideStudioColorPalettesTreeAlteredContentProvider.java
```diff
@@ -0,0 +1,58 @@
...
+                .natures()
+                .stream()
+                .map(Nature::natureId)
+                .anyMatch("siriusComponents://nature?kind=studio"::equals);
+        return isStudioProjectNature && activeFilterIds.contains(SampleTreeFilterProvider.HIDE_STUDIO_COLOR_PALETTES_TREE_FILTER_ID);
+    }
+
+    @Override
+    public List<Object> apply(List<Object> computedElements, VariableManager variableManager) {
+        List<Object> result = new ArrayList<>(computedElements);
```
sbegaudeau: You don't need to create a copy of the input list since `toList()` a couple lines later will already create a new list.

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideRepresentationsTreeItemAlteredContentProvider.java
```diff
@@ -0,0 +1,43 @@
...
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.web.sample.configuration.treefilters;
+
+import java.util.List;
+
+import org.eclipse.sirius.components.core.RepresentationMetadata;
+import org.eclipse.sirius.components.representations.VariableManager;
+import org.eclipse.sirius.web.services.explorer.api.IExplorerTreeItemAlteredContentProvider;
```
sbegaudeau, line 20: The API used for the filtering is in Sirius Web not Sirius Components? So every tree based representation will have to create its own filtering API?

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideRepresentationsTreeItemAlteredContentProvider.java
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Sample {@link IExplorerTreeItemAlteredContentProvider} for the Hide Representations tree item filter.
+ *
+ * @author arichard
+ */
+@Service
+public class HideRepresentationsTreeItemAlteredContentProvider implements IExplorerTreeItemAlteredContentProvider {
+
+    @Override
+    public boolean canHandle(Object object, List<String> activeFilterIds) {
```
sbegaudeau, line 32: it's a bit odd to have such a difference between the signature of the `canHandle(...)` and the `handle(...)` methods. Why give the object here but the variable manager instead in the other method.

I would have imagine something like that:

```
boolean canHandle(VariableManager variableManager, List<Object> computedChildren, List<String> activeFilterIds);
List<Object> canHandle(VariableManager variableManager, List<Object> computedChildren, List<String> activeFilterIds);
```

to make more informed decisions in both use case (access to all the variables in canHandle, access to the list of active filters in handle, etc).

Now that I think about it, why are you giving the list of the active filters? What is the link between a filter and this kind of provider? Is that supposed to be `1..1` or is it `1..*` on purpose? It's a bit odd to have many providers but they each should care about all the filters that are active.

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/explorer/ExplorerChildrenProvider.java
```diff
@@ -0,0 +1,143 @@
...
+            hasChildren = !resource.getContents().isEmpty();
+        } else if (self instanceof EObject eObject) {
+            hasChildren = !eObject.eContents().isEmpty();
+
+            if (!hasChildren) {
+                String id = this.objectService.getId(eObject);
+                hasChildren = this.representationService.hasRepresentations(id);
+            }
+        }
+        return hasChildren;
```
sbegaudeau, line 67: Unless I missed something, this would not be valid anymore. This code would say that an Object without any content would be empty while a provider could add some children. On the other hand a provider could hide all its children and this method could still answer true.

The API must ask someone if any filter changes the value of hasChildren (and we don't want to resolve the children to say so). And If the API needs to handle both getChildren and hasChildren it starts to look a lot like our own API we may be missing other parts.

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/explorer/ExplorerChildrenProvider.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    @Override
+    public List<Object> getChildren(VariableManager variableManager) {
+        List<Object> children = new ArrayList<>();
+        Object self = variableManager.getVariables().get(VariableManager.SELF);
+        if (self != null) {
+            children = this.getDefaultChildren(variableManager);
+            List<String> activeFilterIds = this.getActiveFilterIds(variableManager);
```
sbegaudeau, line 76: I did not catch that the list of active filters is in the variable manager, no need to retrieve them by ourselves, we can just give the variable manager with all its content directly.

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/explorer/ExplorerChildrenProvider.java
```diff
@@ -0,0 +1,143 @@
...
+        String id = null;
+        if (self instanceof RepresentationMetadata representationMetadata) {
+            id = representationMetadata.getId();
+        } else if (self instanceof Resource resource) {
+            id = resource.getURI().path().substring(1);
+        } else if (self instanceof EObject) {
+            id = this.objectService.getId(self);
+        }
+        return id;
+    }
```
sbegaudeau, line 142: The current API of the providers let specifiers return anything but it will only work if they return EObjects, Resources or RepresentationMetadatas otherwise that would not work with the current API.

On one hand, it's generic on all tree representations (given that it impacts sirius-components-collaborative-tree and its GraphQL API for example) and on the other hand it's very specific to the Sirius Web implementation of the explorer. I could add a custom object as a child but since I would not have the ability to give it an id, a label or a kind it would not work.

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/representations/RepresentationService.java
```diff
@@ -50,7 +53,7 @@
  * @author gcoutable
  */
 @Service
-public class RepresentationService implements IRepresentationService, IRepresentationPersistenceService, IDanglingRepresentationDeletionService {
+public class RepresentationService implements IRepresentationService, IRepresentationPersistenceService, IDanglingRepresentationDeletionService, ITransientRepresentationMetadataSearchService {
```
sbegaudeau, line 56: Just for information, I would not add even more responsibilities to this class but I don't care that much since it will be deleted soon so we can keep it that way but in the new architecture it gets replaced by multiple smaller classes.

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/representations/RepresentationService.java
```diff
@@ -202,12 +202,23 @@ public Optional<RepresentationMetadata> findByRepresentation(IRepresentation rep
...
+        Optional<RepresentationMetadata> representationMetadata = Optional.empty();
+        if (representationId.startsWith(ExplorerDescriptionProvider.REPRESENTATION_ID)) {
+            representationMetadata = Optional.of(new RepresentationMetadata(ExplorerDescriptionProvider.REPRESENTATION_ID, Tree.KIND, ExplorerDescriptionProvider.REPRESENTATION_NAME, ExplorerDescriptionProvider.DESCRIPTION_ID));
+        } else if (representationId.startsWith(ModelBrowsersDescriptionProvider.MODEL_BROWSER_CONTAINER_KIND)) {
+            representationMetadata = Optional.of(new RepresentationMetadata(ModelBrowsersDescriptionProvider.MODEL_BROWSER_CONTAINER_KIND, Tree.KIND, ModelBrowsersDescriptionProvider.REPRESENTATION_NAME, ModelBrowsersDescriptionProvider.CONTAINER_DESCRIPTION_ID));
+        } else if (representationId.startsWith(ModelBrowsersDescriptionProvider.MODEL_BROWSER_REFERENCE_KIND)) {
+            representationMetadata = Optional.of(new RepresentationMetadata(ModelBrowsersDescriptionProvider.MODEL_BROWSER_REFERENCE_KIND, Tree.KIND, ModelBrowsersDescriptionProvider.REPRESENTATION_NAME, ModelBrowsersDescriptionProvider.REFERENCE_DESCRIPTION_ID));
+        }
```
sbegaudeau, line 221: Ok, this is why there's a dependency to the reference widget. It wouldn't work for some of the use cases we want to support because some of our consumers are extending the list of transient representations so you can't just have one `ITransientRepresentationMetadataSearchService` for the whole application.

You would also need to consider the other transient representations provided in Sirius Web (validation view, details view, related elements view, representations view). Mixing both #3102 and #3096 in the same commit makes it more difficult to think about the big picture.

Unless my query is invalid because of a typo but with #3102, something like this should work without opening the project with the id of the editing context in a web browser:

```
query {
  viewer {
    editingContext(editingContextId: "...") {
      representation(representationId: "validation") { // the id of the validation view seems to be "validation"
        id
        label
        kind
        description {
          id
        }
      }
    }
  }
}
```

packages/trees/backend/sirius-components-collaborative-trees/src/main/java/org/eclipse/sirius/components/collaborative/trees/api/ITreeFilterProvider.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.eclipse.sirius.components.core.api.IEditingContext;
+
+/**
+ * Interface allowing to provide new tree filters.
+ *
+ * @author arichard
+ */
+public interface ITreeFilterProvider {
+    List<TreeFilter> get(IEditingContext editingContext);
```
sbegaudeau: I would have imagined this API to be parameterized by the representation or its description at least. Different tree based representations could have different set of filters

packages/trees/backend/sirius-components-collaborative-trees/src/main/java/org/eclipse/sirius/components/collaborative/trees/handlers/TreeFiltersEventHandler.java
```diff
@@ -0,0 +1,81 @@
...
+
+        if (treeInput instanceof TreeFiltersInput) {
+            List<TreeFilter> treeFilters = this.treeFiltersProviders.stream()
+                    .map(provider -> provider.get(editingContext))
+                    .filter(Objects::nonNull)
+                    .flatMap(Collection::stream)
+                    .toList();
+            payload = new TreeFiltersSuccessPayload(treeInput.id(), treeFilters);
+        } else {
+            String message = "TreeFiltersEventHandler cannot deal with " + treeInput.getClass().getSimpleName();
```
sbegaudeau: Use our internalized error messages please

packages/trees/backend/sirius-components-trees-graphql/src/main/java/org/eclipse/sirius/components/trees/graphql/datafetchers/tree/TreeDescriptionFiltersDataFetcher.java
```diff
@@ -0,0 +1,68 @@
...
+    public CompletableFuture<List<TreeFilter>> get(DataFetchingEnvironment environment) throws Exception {
+        Map<String, Object> localContext = environment.getLocalContext();
+        String editingContextId = Optional.ofNullable(localContext.get(LocalContextConstants.EDITING_CONTEXT_ID)).map(Object::toString).orElse(null);
+        String representationId = Optional.ofNullable(localContext.get(LocalContextConstants.REPRESENTATION_ID)).map(Object::toString).orElse(null);
+
+        if (editingContextId != null && representationId != null) {
+            String representationWithUniqueId = representationId;
+            if (!representationId.contains("&uniqueId=")) {
+                String uniqueId = "&uniqueId=" + UUID.nameUUIDFromBytes((editingContextId + "[][]").getBytes()).toString();
+                representationWithUniqueId += uniqueId;
```
sbegaudeau: What is the purpose of this code? This feels very odd to have something like that here.

packages/trees/frontend/sirius-components-trees/src/toolbar/TreeToolBar.tsx
```diff
@@ -57,6 +58,7 @@ export const TreeToolBar = ({
           const element = React.createElement(component, props);
           return element;
         })}
+        {treeFiltersMenu}
```
sbegaudeau, line 68: Why does the toolbar takes one of its children as a props instead of creating it?

packages/trees/frontend/sirius-components-trees/src/views/TreeView.tsx
```diff
@@ -159,13 +195,47 @@ export const TreeView = ({
...
 
+  const {
+    loading: allTreeFiltersLoading,
+    data: allTreeFiltersData,
+    error: allTreeFiltersError,
+  } = useQuery<GQLGetAllTreeFiltersData, GQLGetAllTreeFiltersVariables>(getAllTreeFiltersQuery, {
+    variables: {
+      editingContextId,
+      treeId,
+    },
```
sbegaudeau: Since the only relevant variables are editingContextId and treeId, which are given by its parent and since the transient metadata providers are supposed to do their job, I don't know why the code used to fetch the filters is here and not above nor why does it still need the treeView to have been fetched.

Is the fix for #3102 that you have implemented working?

sbegaudeau (changes requested):

sbegaudeau:
Your work on #3102 should also impact `RepresentationApplicationService#findRepresentationMetadataById`

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideRepresentationsTreeItemAlteredContentProvider.java
AxelRICHARD: Done

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideRepresentationsTreeItemAlteredContentProvider.java
AxelRICHARD: Done

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideStudioColorPalettesTreeAlteredContentProvider.java
AxelRICHARD: Done

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideRepresentationsTreeItemAlteredContentProvider.java
AxelRICHARD: `IExplorerTreeItemAlteredContentProvider` is for the `ExplorerView` which is only for Sirius Web, and we also agreed that it should be in Sirius Web in the ADR.
Do you think it should be for all views declaring a `TreeDescription`?

packages/trees/backend/sirius-components-collaborative-trees/src/main/java/org/eclipse/sirius/components/collaborative/trees/api/ITreeFilterProvider.java
AxelRICHARD: I followed what we agreed in the related ADR. But I will change it, if you think it is best.

packages/sirius-web/backend/sirius-web-graphql/src/main/java/org/eclipse/sirius/web/graphql/datafetchers/editingcontext/EditingContextRepresentationDataFetcher.java
AxelRICHARD: I don't understand how this could replace the previous block.

packages/trees/backend/sirius-components-collaborative-trees/src/main/java/org/eclipse/sirius/components/collaborative/trees/handlers/TreeFiltersEventHandler.java
AxelRICHARD: Done

packages/trees/backend/sirius-components-trees-graphql/src/main/java/org/eclipse/sirius/components/trees/graphql/datafetchers/tree/TreeDescriptionFiltersDataFetcher.java
AxelRICHARD: The problem here is that we can receive a `representationId` equals to `explorer://`. It corresponds to a representation that has not been created yet in the background but will be soon form what I understand.
But, when a Tree is created, it produces an Id with a `&uniqueId=` part (see `TreeConfiguration`).
So,  when the case occurs, I generate the `&uniqueId=` part corresponding to a tree representation that has just been created (with no tree item expanded).

packages/trees/frontend/sirius-components-trees/src/toolbar/TreeToolBar.tsx
AxelRICHARD: Because of the `treeFilters` list which is in the state of `ExplorerView`.
I will try to fix that.

packages/trees/frontend/sirius-components-trees/src/views/TreeView.tsx
AxelRICHARD: I put it here because the retrieving of the filters of a tree should be done after the subscription of the tree succeeds.
I will try to change that but seems difficult to get the filters before the subscription exists.
Indeed, in `EditingContextEventProcessor#handleRepresentationInput`, the `representationInput` may contains a `representationId` that doesn't exist yet if the subscription has not been executed before.

AxelRICHARD: Is the fix for https://github.com/eclipse-sirius/sirius-web/issues/3102 that you have implemented working? -> Yes

AxelRICHARD:
> Your work on #3102 should also impact `RepresentationApplicationService#findRepresentationMetadataById`

Ok thank you, at the time this PR was pushed, the `RepresentationApplicationService` didn't exist. I will rebase this PR and will try to modify  `RepresentationApplicationService#findRepresentationMetadataById`.

packages/trees/frontend/sirius-components-trees/src/views/TreeView.tsx
AxelRICHARD: As expected, this version get the filters before the subscription exists.
Indeed, in EditingContextEventProcessor#handleRepresentationInput, the representationInput may contains a representationId that doesn't exist yet if the subscription has not been executed before.

packages/sirius-web/backend/sirius-web-application/pom.xml
```diff
@@ -144,6 +144,16 @@
...
+		<dependency>
+			<groupId>org.eclipse.sirius</groupId>
+			<artifactId>sirius-components-trees</artifactId>
+			<version>2024.1.4</version>
+		</dependency>
+		<dependency>
+			<groupId>org.eclipse.sirius</groupId>
+			<artifactId>sirius-components-collaborative-widget-reference</artifactId>
+			<version>2024.1.4</version>
+		</dependency>
```
sbegaudeau: I'll remove that for now since the reference widget cannot be reused without tons of dependencies see my work [here](https://github.com/eclipse-sirius/sirius-web/pull/3199) to see what would be necessary for that.

packages/sirius-web/backend/sirius-web-application/pom.xml
```diff
@@ -144,6 +144,16 @@
 			<artifactId>sirius-components-view-gantt-edit</artifactId>
 			<version>2024.1.4</version>
 		</dependency>
+		<dependency>
+			<groupId>org.eclipse.sirius</groupId>
+			<artifactId>sirius-components-trees</artifactId>
+			<version>2024.1.4</version>
+		</dependency>
```
sbegaudeau: There is already a dependency to sirius-cmponents-tree line 94 :)

sbegaudeau: Done

packages/sirius-web/backend/sirius-web-application/pom.xml
sbegaudeau: Done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/representation/services/RepresentationApplicationService.java
```diff
@@ -83,4 +90,16 @@ public Optional<String> findEditingContextIdFromRepresentationId(String represen
...
+        Optional<RepresentationMetadata> representationMetadata = Optional.empty();
+        if (representationId.startsWith(ExplorerDescriptionProvider.REPRESENTATION_ID)) {
+            representationMetadata = Optional.of(new RepresentationMetadata(ExplorerDescriptionProvider.REPRESENTATION_ID, Tree.KIND, ExplorerDescriptionProvider.REPRESENTATION_NAME, ExplorerDescriptionProvider.DESCRIPTION_ID));
+        } else if (representationId.startsWith(ModelBrowsersDescriptionProvider.MODEL_BROWSER_CONTAINER_KIND)) {
+            representationMetadata = Optional.of(new RepresentationMetadata(ModelBrowsersDescriptionProvider.MODEL_BROWSER_CONTAINER_KIND, Tree.KIND, ModelBrowsersDescriptionProvider.REPRESENTATION_NAME, ModelBrowsersDescriptionProvider.CONTAINER_DESCRIPTION_ID));
+        } else if (representationId.startsWith(ModelBrowsersDescriptionProvider.MODEL_BROWSER_REFERENCE_KIND)) {
+            representationMetadata = Optional.of(new RepresentationMetadata(ModelBrowsersDescriptionProvider.MODEL_BROWSER_REFERENCE_KIND, Tree.KIND, ModelBrowsersDescriptionProvider.REPRESENTATION_NAME, ModelBrowsersDescriptionProvider.REFERENCE_DESCRIPTION_ID));
+        }
```
sbegaudeau: We talked about potentially removing that for now but since the widget reference cannot be reused as is, I'll do it

sbegaudeau: Done

packages/sirius-web/backend/sirius-web-graphql/src/main/java/org/eclipse/sirius/web/graphql/datafetchers/editingcontext/EditingContextRepresentationDataFetcher.java
sbegaudeau: I've tried and it can't right now since, contrary to the explorer, the details view does not have an identifier that we can easily detect (it doesn't start with something like details://). But it gives me an idea, transient representations are not being persisted so they don't need to have an UUID as a representation id so we could make it mandatory for transient representations to have an identifier which we can understand (`explorer://`, `details://`, `relatedelements://`, `validation://`, etc). It would make things way easier.

packages/sirius-web/backend/sirius-web-sample-application/src/main/java/org/eclipse/sirius/web/sample/configuration/treefilters/HideRepresentationsTreeItemAlteredContentProvider.java
sbegaudeau: > Do you think it should be for all views declaring a TreeDescription?

I thought about it about more and I think that those two specific interfaces should stay as they are for now. While filters should be available on all trees, how filters are applied may be very specific to each tree. We will see how this evolve in the future so this API may move.

sbegaudeau (approved):

PR #3170, opened 2024-02-27
[3019] Clear the domain events collector before the tests

### LINKED ISSUE #3021: Simplify the API used to extend parts of the Sirius Web backend

pcdavid:
Without this, running the complete suite of tests in `sirius-web` fails on `ProjectControllerIntegrationTests.givenValidProjectToCreateWhenMutationIsPerformedThenTheSemanticDataAreCreated()` because the `IDomainEventCollector` is reused and the previous test in the suite, from another test class, leaves an event in the collector:

Capture d’écran du 2024-02-27 15-20-39


sbegaudeau:
That's odd I never had this issue but since tests can be executed randomly I suppose it can indeed happen. Isn't there another test where the collector is used in a similar fashion and where a before each should be added too?

pcdavid:
I guess it depends on the order in which the test runner executes them, and Eclipse's JUnit runner orders them differently that IntelliJ's. I get the same order every time under STS.

I'll look for other places where the same could occur with other orderings.

pcdavid:
Same behavior (and same fix) with `RepresentationControllerIntegrationTests` if it happens to be executed right after `ProjectControllerIntegrationTests` (tested by disabling `ProjectControllerIntegrationTests`):

Capture d’écran du 2024-02-28 16-40-50

sbegaudeau:
Just like I thought the test containers dependency could also be updated for `sirius-web-sample-application` and `sirius-web-persistence`. I've updated both just like you did and I'll merge this PR once the build is good.

sbegaudeau (approved):

PR #3209, opened 2024-03-11
[doc] Contribute a shape for default collapse and collapse operation

gdaniel:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/adrs/140_add_collapse_and_expand_services.adoc
+++ b/doc/adrs/140_add_collapse_and_expand_services.adoc
@@ -1,0 +1,39 @@
+= ADR-140 Add collapse and expand services
+
+== Context
+
+Sirius Web doesn't provide a way for specifiers to define tools that collapse/expand nodes as part of other operations (e.g. to expand a collapsed node when a child node is created).
+
+We will add a new `diagramServices` variable  with services to collapse/expand nodes from AQL expressions and Java services.
+
+== Decision
+
+We will create a new interface `IDiagramServices` in `org.eclipse.sirius.components.collaborative.diagrams.api` with the following signature:
+```java
+public interface IDiagramServices {
+    void collapse(IDiagramServicesContext context, List<Node> nodes);
+    void expand(IDiagramServicesContext context, List<Node> nodes);
+}
+```
+
+We will add a new variable `diagramServices : IDiagramService` in the `VariableManager`.
+The `IDiagramService` would be a wrapper for objects that can be accessed by `IDiagramServices` services.
+For the moment it would only define a getter on the `DiagramContext`, which is required in our implementation of collapse/expand.
+
+This will allows developers to call these services from their AQL expressions and Java services.
+
+The implementation of the `IDiagramServices` interface will be located in `org.eclipse.sirius.components.collaborative.diagrams`.
+
+We will make `DiagramContext#diagramEvent`, `DiagramComponentProps#diagramEvent`, and `NodeComponentProps#diagramEvent` lists to store multiple events.
+
+The collapse/expand services will store instances of `UpdateCollapsingStateEvent` in the `DiagramContext#diagramEvent` list.
+
+We will update the `NodeComponent#doRender` and `NodeComponent#computeCollapsingState` methods to use this information (stored in `NodeComponent#props`) to compute the collapsing state of a node.
+
+== Status
+
+Work in progress
+
+== Consequences
+The new interface `IDiagramServices` should be used in the future to define diagram-level services.
+
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -5,8 +5,13 @@
 
 === Shapes
 
+- Configurable collapse
+
 
 === Architectural decision records
+
+- [ADR-139] Add Is Collapsed By Default Expression in NodeDescription entity in the view DSL
+- [ADR-140] Add collapse and expand services
 
 
 === Deprecation warning
@@ -52,6 +57,7 @@
 - [ADR-136] Add variables to the view form dsl
 - [ADR-137] Use a common identification strategy for transient representations
 - [ADR-138] Allow consumers to reuse smaller parts of the Sirius Web backend
+
 
 === Deprecation warning
 
--- a/doc/adrs/139_add_is_collapsed_by_default_in_view_dsl.adoc
+++ b/doc/adrs/139_add_is_collapsed_by_default_in_view_dsl.adoc
@@ -1,0 +1,29 @@
+= ADR-139 Add Is Collapsed By Default Expression in NodeDescription entity in the view DSL
+
+== Context
+
+Sirius Web doesn't provide a way for specifiers to set the default collapsible state of a node. 
+We will add new field _Is Collapsed By Default Expression_ in NodeDescription to handle it. 
+The field contains an boolean expression which is evaluated on the first rendering of the node to set its initial collapsed/expanded state.
+
+== Decision
+
+We will add a new field `isCollapsedByDefaultExpression : InterpretedExpression` in the `NodeDescription` element of the diagram metamodel. 
+This expression should return a boolean value, where `true` means collapsed, and `false` means expanded.
+This field is empty by default, which is evaluated to `false` (expanded) to keep the existing default behavior of Sirius Web.
+
+We will add the field `Predicate<VariableManager> collapsedByDefaultPredicate` in `org.eclipse.sirius.components.diagrams.description.NodeDescription` and its builder, and update the `ViewDiagramDescriptionConverter` to handle it. 
+
+We will update the `NodeComponent#doRender` and `NodeComponent#computeCollapsingState` methods to evaluate this expression as part of the rendering process. 
+The `isCollapsedByDefault` expression should be used only for the first rendering of a node (i.e. if the `optionalPreviousNode` is empty in `NodeComponent#computeCollapsingState`).  
+
+We won't update the GraphQL schemas nor the frontend, since collapsed state is handled via visibility, and the existing code already handles it.
+
+== Status
+
+Work in progress
+
+== Consequences
+
+No consequences detected for now.
+
--- a/doc/iterations/2024.5/configurable_collapse.adoc
+++ b/doc/iterations/2024.5/configurable_collapse.adoc
@@ -1,0 +1,48 @@
+= Configurable collapse
+
+== Problem
+
+Sirius Web doesn't provide a way for specifiers to set the default collapsible state of a node. 
+It also doesn't provide a way to let specifiers define tools that collapse/expand nodes as part of other operations (e.g. to expand a collapsed node when a child node is created).
+Nodes are always created expanded, and the only way to collapse them right now is that the end user does it manually.
+
+== Key Result
+
+- Specifiers must be able to define NodeDescriptions that are collapsed/expanded by default.
+- Specifiers must be able to call collapse/expand services in their tool to collapse/expand a set of nodes.
+
+== Solution
+
+A new field _Is Collapsed By Default Expression_ should be displayed in NodeDescription. 
+The field contains an boolean expression which is evaluated on the first rendering of the node to set its initial collapsed/expanded state.
+If the expression is empty it is evaluated to `false`, which means that the node is expanded by default (this preserves Sirius Web's default behavior to expand new nodes by default). 
+
+A new `diagramServices` variable  with services to collapse/expand nodes should be accessible from AQL expressions and Java services.
+While this shape is focused on collapsing/expanding nodes, this variable should ultimately contain several diagram-level services (e.g. pin, hide, etc).
+
+== Scenario
+
+1. Default Collapsing State
+- A specifier adds a NodeDescription in the view DSL
+- The _Is Collapsed By Default Expression_ field is available and its default value is empty, which means expanded
+- The specifier can define an AQL expression to that returns a boolean value to set the default collapsing state of the node (`true` for collapsed, `false` for expanded)
+- When an end user creates a node represented in this manner in a diagram, the node is collapsed/expanded according to this expression
+- If the user manually expands/collapses the node the value returned by the expression is not taken into account (the expression can't override user-defined state).
+
+2. Collapse/Expand operation
+- A specifier creates a tool in the view DSL
+- The specifier creates an operation, for example ChangeContext
+- The specifier fills the expression of its operation with a call to the collapse/expand service, for example `diagramServices.collapse(nodeId)`
+- When an end user uses the tool defined this way, the appropriate nodes are collapsed/expanded
+
+=== Cutting backs
+
+Nothing identified for now
+
+== Rabbit holes
+
+Nothing identified for now
+
+== No-gos
+
+Nothing identified for now


doc/iterations/2024.3/shapes/configurable_collapse.adoc
```diff
@@ -0,0 +1,47 @@
...
+== Key Result
+
+- A specifier must be able to define NodeDescriptions that are collapsed/expanded by default.
+- A specifier can define tools that collapse/expand other nodes.
+
+== Solution
+
+A new field _Default Collapsing State Expression_ should be displayed in NodeDescription. The field contains an expression which is evaluated on each diagram render to collapse/expand a node.
+
+This expression would be used to set the Collapsing State iff it hasn't been manually set by the end user or updated by a tool.
```
AxelRICHARD: iff -> if

doc/iterations/2024.3/shapes/configurable_collapse.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+A new operation _Set Collapsing State_ should be available in tools to expand/collapse nodes. This operation would have a _Graphical Candidate Expression_ field used to retrieve the nodes to expand/collapse, as well as a _CollapsingState_ field containing the value to set (Collapsed or Expanded).
```
AxelRICHARD: _Graphical Candidates Expression_ since it retrieves a lis of nodes?

doc/iterations/2024.3/shapes/configurable_collapse.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+== Scenario
+
+1. Default Collapsing State
+- A specifier adds a NodeDescription in the view DSL
+- The _Default Collapsing State Expression_ field is available and its default value is empty
```
AxelRICHARD: Could you add here that _empty_ means expanded please?

doc/iterations/2024.3/shapes/configurable_collapse.adoc
```diff
@@ -0,0 +1,47 @@
...
+- The specifier can define an AQL expression to set the default Collapsing State of the node
+- When an end user creates a node represented in this manner in a diagram, the node is collapsed/expanded according to this expression.
+- If the user manually expands/collapses the node the default _Collapsing State Expression_ is not taken into account on the next renders.
```
AxelRICHARD: From what I understand, the _Collapsing State Expression_ will be evaluated on each render. So why is it called default collapsing state expression? Shouldn't it be just collapsing state expression?

gdaniel: If we want to make sure the collapse/expand actions triggered by the user have precedence over this expression we need to reuse the previous state of the node on a render: if a previous collapsing state value exists we reuse it for the next render. We cannot override it with the _Collapsing State Expression_ otherwise we lose user interaction information.

Based on that, the only time this expression will be evaluated is the first render of the node (because there isn't a previous value for the collapsing state), which is similar to a default value.  To clarify, I copied the behavior of the default width/height expressions, where the expressions are used only if the node doesn't have a set width/height, which only happens on the first rendering.

If this behavior is fine for you I suggest to rewrite the first 2 paragraphs of the solution to clarify that this expression is evaluated on the first render.

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: done

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: done

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: done

doc/iterations/2024.3/shapes/configurable_collapse.adoc
```diff
@@ -0,0 +1,47 @@
+= Configurable collapse
+
+== Problem
+
+Sirius Web doesn't provide a way for specifiers to set the default collapsible state of a node. It also doesn't provide a way to let specifiers define tools that collapse/expand nodes as part of other operations (e.g. to expand a collapsed node when a child node is created).
```
sbegaudeau: one sentence per line please (I won't repeat this comment everywhere but it is valid at multiple locations)

doc/iterations/2024.3/shapes/configurable_collapse.adoc
```diff
@@ -0,0 +1,47 @@
...
+Nodes are always created expanded, and the only way to collapse them right now is that the end user does it manually.
+
+== Key Result
+
+- A specifier must be able to define NodeDescriptions that are collapsed/expanded by default.
+- A specifier can define tools that collapse/expand other nodes.
+
+== Solution
+
+A new field _Default Collapsing State Expression_ should be displayed in NodeDescription. The field contains an expression which is evaluated on each diagram render to collapse/expand a node.
```
sbegaudeau: What would this operation return? A `org.eclipse.sirius.components.diagrams.CollapsingState`? How could end users figure out how to return one from an AQL expression written without Java services?

Something like `isCollapsedByDefaultExpression` asking for a `boolean` in return would be easier to use. The default value of a boolean being `false`, it would make sense to have `false` by default in case this expression is empty.

doc/iterations/2024.3/shapes/configurable_collapse.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+This expression would be used to set the Collapsing State if it hasn't been manually set by the end user or updated by a tool.
+
+A new operation _Set Collapsing State_ should be available in tools to expand/collapse nodes. This operation would have a _Graphical Candidates Expression_ field used to retrieve the nodes to expand/collapse, as well as a _CollapsingState_ field containing the value to set (Collapsed or Expanded).
```
sbegaudeau: While you are definitely right in thinking that model operations are the way to add new behavior to be executed by specifier, it is the worst way to do it. Trust me, we did this mistake several times already.

What happens if the behavior that I want to execute is in a Java service? I need to execute my Java service and then go back to the model operations world once again. It is very cumbersome. If you want to let specifiers change the collapsing state, give them a service to do it.

You will need to build some API to change the collapsing state (API which you would have to leverage in your model operation) but instead of building a model operation around it, simply expose it to specifiers programmatically.

This way, it would work nicely for those who are building their diagrams programmatically (you definitely can't forget those users), it would work in AQL expressions, it would work in Java services, it would just work everywhere. Apart from control flow (if, for, let) and some basic features to let newcomers get started (setValue, changeContext), we could remove every single model operation available in Sirius Desktop and Web and offer them as services and it would be a massive improvement in flexibility for specifiers. Don't bother with a model operation, we did this mistake several time over the past few years. Just add a specific variable (to not pollute existing objects with a new service) and add your service on this. Something like that (which should belong in the ADR not the shape):

```
public interface ISiriusWebServices {
  void collapse(List<String> nodeIds);
  void expand(List<String> nodeIds);
}
```

To let specifiers write:

```
aql:siriusWebServices.collapse(Sequence{selectedNode.getId()})
```

or in custom Java services:

```
public EObject doStuff(Diagram previousDiagram, EObject eObject, ISiriusWebServices siriusWebServices) {
  var nodeIds = previousDiagram.getNodes().stream().map(Node::getId).toList();
  siriusWebServices.collapse(nodeIds);
  return eObject;
}
```

We started this pattern with PCD with the `DomainServices` and the variable `input` during our work on Sirius Desktop and EEF and its way more powerful given its flexibility. It's also easier to build and maintain, you don't need anything special for the view DSL. I would not be a fan of naming such variable `input` again because we have tons of `inputs` already and this name has another meaning in our case.

doc/iterations/2024.3/shapes/configurable_collapse.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+== Scenario
+
+1. Default Collapsing State
+- A specifier adds a NodeDescription in the view DSL
+- The _Default Collapsing State Expression_ field is available and its default value is empty, which means expanded to keep the existing default behavior of Sirius Web
```
sbegaudeau: empty means expanded is the most logical choice but not very explicit with this expression name

doc/iterations/2024.3/shapes/configurable_collapse.adoc
sbegaudeau: > if a previous collapsing state value exists we reuse it for the next render

You will always have a collapsing state value, we always store something and should always know the current value. If you want to detect the first rendering, it's the moment where the `optionalPreviousNode` is empty for example in `org.eclipse.sirius.components.diagrams.components.NodeComponent#computeCollapsingState`

doc/iterations/2024.3/shapes/configurable_collapse.adoc
```diff
@@ -0,0 +1,47 @@
...
+- The specifier can define an AQL expression to set the default Collapsing State of the node
+- When an end user creates a node represented in this manner in a diagram, the node is collapsed/expanded according to this expression.
+- If the user manually expands/collapses the node the default _Collapsing State Expression_ is not taken into account on the next renders.
+
+2. Collapse/Expand operation
+- A specifier creates a tool in the view DSL
+- The operation _Set Collapsing State_ is available in the "new" dropdown menu and creates a _Set Collapsing State_ element under the tool
+- The operation has two fields _Graphical Candidate Expression_ (a text field accepting an AQL expression) and _CollapsingState_ (a dropdown with 2 values: Collapsed and Expanded)
```
sbegaudeau: You see with such an approach, I could not easily select the collapsing state conditionally (for example expand all nodes if there is an odd number of nodes in the diagram).

doc/adrs/137_add_default_collapsing_state_in_view_dsl.adoc
```diff
@@ -0,0 +1,26 @@
+= ADR-137 Add Default Collapsing State Expression in NodeDescription entity in the view DSL
+
+== Context
+
+Sirius Web doesn't provide a way for specifiers to set the default collapsible state of a node. We will add new field _Default Collapsing State Expression_ in NodeDescription to handle it. The field will contain an expression which is evaluated on each diagram render to collapse/expand a node.
```
sbegaudeau: Once sentence per line in asciidoc please

doc/adrs/137_add_default_collapsing_state_in_view_dsl.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+== Decision
+
+We will add a new field `defaultCollapsingStateExpression : InterpretedExpression` in the `NodeDescription` element of the diagram metamodel. This field is empty by default.
```
sbegaudeau: You are never telling us what this expression should return and I don't think that it can work as its name suggest at all.

doc/adrs/137_add_default_collapsing_state_in_view_dsl.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+We will add the equivalent field in `NodeDescription` and its builder, and update the `ViewDiagramDescriptionConverter` to handle this new field.
```
sbegaudeau: What would be the equivalent field in NodeDescription? Given my comments in the shape on your suggested expression, I think it should be something like `Predicate<VariableManager> collapsedByDefaultPredicate`

doc/adrs/138_add_set_collapsing_state_operation_in_view_dsl.adoc
```diff
@@ -0,0 +1,25 @@
+= ADR-138 Add Set Collapsing State Operation in View DSL
+
+== Context
+
+Sirius Web doesn't provide a way for specifiers  to let specifiers define tools that collapse/expand nodes as part of other operations (e.g. to expand a collapsed node when a child node is created). We will add a new _Set Collapsing State_ operation that can be used in tools to expand/collapse nodes as part of the tool execution process.
```
sbegaudeau: Once sentence per line in asciidoc please

doc/adrs/138_add_set_collapsing_state_operation_in_view_dsl.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+== Decision
+
+We will add a new `SetCollapsingState` class in the diagram metamodel. This class extends `Operation`, and contains 2 fields:
+
+- `graphicalCandidateExpression : InterpretedExpression`: returns the list of graphical elements to expand/collapse
+- `collapsingState : InterpretedExpression` : returns the CollapsingState to set for the candidates
+
+We will create a new `ViewSetCollapsingStateRequest` class in `org.eclipse.sirius.components.diagrams` with an `elementId` and `collapsingState` attributes. We will update the `DiagramContext` to store these requests.
```
sbegaudeau: Do not make any reference to operations, the view DSL or anything like that in `sirius-components-diagrams`. Diagrams are supposed to be usable programmatically without expressions, without odesigns, and without the view DSL.

sbegaudeau: Why create an entirely new mechanism to update the collapsing state instead of building on the existing one? Why this `ViewSetCollapsingStateRequest` instead of reusing `UpdateCollapsingStateEvent` which already as the same exact properties?

sbegaudeau (changes requested):
Start by reading my comments on the shape since it is the entry point of the whole discussion.

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: Agreed, the idea is to evaluate it on the first rendering. To me this is the only time where the default value makes sense. As you said, besides the first rendering, we always know the current value, and we don't want to update it based on the default value.

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: You are right, a boolean value makes more sense. I assumed `org.eclipse.sirius.components.diagrams.CollapsingState` had a matching enumeration in the metamodel that could be reused in AQL, but it is not the case and with your explanations I realize it's not a good idea.

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: Agreed, it would make more sense if the expression was returning a boolean, as you suggest above

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: Couldn't you do something like `aql: if('diagram.nodes->size() % 2 == 1) then diagram.nodes else Sequence{} endif`? I agree it is not pretty though, and in any case it would be easier/cleaner with a service.

doc/adrs/137_add_default_collapsing_state_in_view_dsl.adoc
gdaniel: Based on the previous comments it should be something like `defaultIsCollapsedExpression : InterpretedExpression`, which should return a boolean

doc/adrs/137_add_default_collapsing_state_in_view_dsl.adoc
gdaniel: Agreed

doc/adrs/138_add_set_collapsing_state_operation_in_view_dsl.adoc
gdaniel: I am not sure I fully understand how to address the second comment. I can indeed reuse `UpdateCollapsingStateEvent`, but should it be stored in a collection like the `ViewCreationRequests`?
What drove me away from `UpdateCollapsingStateEvent` is that it is used in the `UpdateCollapsingStateHandler` to set `DiagramContext#diagramEvent`. I can't do this in this context because I may have multiple `UpdateCollapsingStateEvents`, and in any case the `DiagramContext#diagramEvent` will be overriden by the `InvokeSingleClickHandler` at the end of the tool execution.

I think I am missing some elements on how `DiagramContext#diagramEvent` is used, and when should a tool set it instead of using a mechanism like the one used for `ViewCreationRequests`

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: I agree on the solution you propose, I realize my solution based on an operation only was a bit short sighted, and we should provide a tool that can be used in all the context you mention. I have a question to clarify your suggestion:

> Just add a specific variable (to not pollute existing objects with a new service) and add your service on this.

The variable here is an AQL variable right? If this is the case I assume we want a single variable for a variety of services? This shape only discusses collapse/expand, but if we want to add similar services in the future it should be on the same object/on a nested object, right? Something like

```java
public interface ISiriusWebServices {
    void collapse(List<String> nodeIds);
    void expand(List<String> nodeIds);
    void hide(List<String> nodeIds);
    void pin(List<String> nodeIds)
    // etc
}
```
Some services may not make sense, but I want to clarify that you don't want a new variable that would only contain services to expand/collapse.

doc/iterations/2024.3/shapes/configurable_collapse.adoc
AxelRICHARD: Ok, so it is `initialCollapsingStateExpression` or `defaultCollapsingStateExpression`

doc/iterations/2024.3/shapes/configurable_collapse.adoc
sbegaudeau: Yes definitely, we will end up with tons of other features in there like `pin`, `hide`, `createView`, etc. The variable will be usable in AQL but it will simply be a new variable in the `VariableManager`. For example, in `org.eclipse.sirius.components.collaborative.diagrams.DiagramCreationService#doRender` we have a custom variable named `environment` (which is not fully finished) but which would let specifiers figure out in which application their code is executed (Sirius Web vs SysOn for example).

You could add there a new variable `siriusWebServices` or something like that which would then be accessible by everybody (programmatic diagram description, AQL expression, Java services, etc). To make its methods available in AQL expressions, you would have to register its services in the instance of `AQLInterpreter` created for diagrams.

When we convert representation descriptions, we provide the same instance of AQLInterpreter to all representation descriptions (see org.eclipse.sirius.components.view.emf.ViewConverter#convert(...)) while you would need to create an instance per representation description converted and for diagrams you would have to register as services the methods of `ISiriusWebServices`.

This makes me realize that this interface is badly named since it should be `ISiriusWebDiagramServices` or `IDiagramServices` with a variable named `diagramServices` for example. We will end up with form-related services, tree-related ones etc in the near future.

doc/iterations/2024.3/shapes/configurable_collapse.adoc
sbegaudeau: Yes you could indeed but if my use case becomes a tiny bit more complex with if it's an even number of node collapse everything or do something even more complicated you end up having to write complex stuff in AQL or even in model operations (you would end up having maybe two model operations and conditionally decide which one to call).

None of our advanced users wants to code in AQL or model operations.   (✖╭╮✖)

doc/adrs/137_add_default_collapsing_state_in_view_dsl.adoc
sbegaudeau: I would rather have predicate starting by `is` or `has` as a matter of conventions (even if I'm pretty sure that some other predicate may not be following said convention already -.-) so I prefer something like `isCollapsedByDefaultExpression`

doc/adrs/137_add_default_collapsing_state_in_view_dsl.adoc
gdaniel: done

doc/adrs/137_add_default_collapsing_state_in_view_dsl.adoc
gdaniel: done

doc/adrs/137_add_default_collapsing_state_in_view_dsl.adoc
gdaniel: updated the ADR to integrate this feedback

doc/adrs/138_add_set_collapsing_state_operation_in_view_dsl.adoc
gdaniel: done

doc/adrs/138_add_set_collapsing_state_operation_in_view_dsl.adoc
gdaniel: updated the ADR with a new proposal. I am not sure where the service should be located, I put sirius-web-services in the ADR, but I don't know if it is an appropriate location.

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: done

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: updated the shape to integrate the feedback

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: updated the shape to integrate the feedback (part of the feedback has been moved to an ADR as suggested)

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: updated the shape to integrate the feedback

doc/iterations/2024.3/shapes/configurable_collapse.adoc
gdaniel: @sbegaudeau suggested `isCollapsedByDefaultExpression` to clarify that the expression should return a boolean

AxelRICHARD (approved):

doc/adrs/138_add_collapse_and_expand_services.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+We will create a new interface `IDiagramServices` in `org.eclipse.sirius.components.collaborative.diagrams.api` with the following signature:
+```java
+public interface IDiagramServices {
+    void collapse(IDiagramServicesContext context, List<Node> nodes);
+    void expand(IDiagramServicesContext context, List<Node> nodes);
+}
+```
+
+We will add a new variable `diagramServices : IDiagramServiceContext` in the `VariableManager`.
```
sbegaudeau: We have tons of things name `XxxContext` just use something like diagram service. I'm trying to limit the usage of diagram context which we had to provide as an API a bit too quickly in my mind so I'm not fan of another context. This term should be necessary for our consumers.

doc/adrs/140_add_collapse_and_expand_services.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+public EObject doStuff(Diagram previousDiagram, EObject eObject, IDiagramServicesContext diagramServicesContext) {
+  diagramServices.collapse(diagramServicesContext, previousDiagram.getNodes());
+  return eObject;
+}
+```
+
+The implementation of the `IDiagramServices` interface will be located in `org.eclipse.sirius.components.collaborative.diagrams`.
+
+The collapse/expand services will store instances of `UpdateCollapsingStateEvent` in a new attribute `DiagramContext#updateCollapsingStateEvents`.
```
sbegaudeau: Just make the existing diagram events a list, no need to create one new attribute per event type.

doc/adrs/140_add_collapse_and_expand_services.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+The same attribute will be defined in `DiagramComponentProps` and `NodeComponentProps`, and initialized from the value stored in the `DiagramContext`.
```
sbegaudeau: I don't want one API break per kind of operation we will want to support, make DiagramComponentProps#diagramEvent a list instead.

doc/adrs/140_add_collapse_and_expand_services.adoc
```diff
@@ -0,0 +1,62 @@
...
+
+We will update the `NodeComponent#doRender` and `NodeComponent#computeCollapsingState` methods to use this information (stored in `NodeComponent#props`) to compute the collapsing state of a node:
+
+```java
+public CollapsingState computeCollapsingState(String nodeId, Optional<Node> optionalPreviousNode, Optional<IDiagramEvent> optionalDiagramEvent, Optional<List<UpdateCollapsingStateEvent>> optionalUpdateCollapsingStateEvents) {
```
sbegaudeau: `Optional<List<xxx>>` why not a list?

sbegaudeau (commented):
At this point, I think it would be better to add less code in the ADR and just merge it. You don't have to write any or all your code changes in the ADR and I feel like I'm reviewing the implementation already. Let's juste simplify the ADR to merge it and contribute / review the real code.

doc/adrs/138_add_collapse_and_expand_services.adoc
gdaniel: renamed to IDiagramService

doc/adrs/140_add_collapse_and_expand_services.adoc
gdaniel: done

doc/adrs/140_add_collapse_and_expand_services.adoc
gdaniel: done

doc/adrs/140_add_collapse_and_expand_services.adoc
gdaniel: removed this part to simplify the ADR

sbegaudeau (approved):

sbegaudeau (changes requested):
I'll update the ADR to add them to 2024.5 instead of 2024.3

sbegaudeau (approved):
Now it's ok :)

PR #3264, opened 2024-03-20
[doc] Add shape on diagram assertions and navigators

gdaniel:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -7,6 +7,7 @@
 - Configurable collapse
 - Add support for EMF models migration
 - Add support for representations migration
+- Add Assertions and Navigators to improve Diagram testability
 
 === Architectural decision records
 
--- a/doc/iterations/2024.5/shapes/add_assertions_and_navigator_to_improve_diagram_testability.adoc
+++ b/doc/iterations/2024.5/shapes/add_assertions_and_navigator_to_improve_diagram_testability.adoc
@@ -1,0 +1,53 @@
+= Add Assertions and Navigators to improve Diagram Testability
+
+== Problem
+
+Defining diagram tests that require complex diagram navigation is cumbersome.
+Existing assertions could be improved to ease the verification of diagram elements.
+
+== Key Result
+
+Testers can write simpler tests that rely on new navigation capabilities to retrieve the diagram elements involved in their tests.
+Testers can validate diagram elements using custom assertions.
+
+== Solution
+
+=== Navigator APIs
+
+A set of navigator APIs to:
+
+- Access a list of elements regardless of their position in the diagram, and matching a given condition (label, description, semantic target label)
+- Access an unique element regardless of its position in the diagram, and matching a given condition
+- Access the direct children of an element matching a given condition
+- Access the direct parent or the list of ancestors of an element
+- Access the outgoing/incoming edges of a node, matching a given condition
+- Access the label of an element
+
+All these operations should have two modes: one that stops the navigation and returns the list of elements (e.g. `findBorderNodes()`), and one that allows to continue the navigation (e.g. `borderNodes().ancestors()`).
+
+Navigators will need to cache information for performance reasons (we don't want to actually navigate the model on every query), as well as to provide navigation capabilities that aren't directly accessible from diagram elements (e.g. a node doesn't know its ancestors).
+
+=== Assertion APIs
+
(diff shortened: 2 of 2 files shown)


doc/iterations/2024.3/shapes/add_assertions_and_navigator_to_improve_diagram_testability.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+=== Navigator APIs
+
+A set of navigator APIs (`DiagramNavigator`, `NodeNavigator`, etc) with the same approach as the existing `FormNavigator`.
+
+For example, these APIs should provide ways to:
+
+```java
+// Get the list of elements with a given description regardless of their location in the diagram
+List<Node> = new DiagramNavigator(diagram)
```
sbegaudeau: In the FormNavigator, I distinguish two use cases:

- either my methods return another navigator
- or they find something for the end user

All methods used to find stuff are prefixed by `find` which is the difference in:

```
new FormNavigator(form)
    .page("Label")
    .group("Label")
    .findWidget("Label", Textfield.class)
```

You will need such distinction even more than me because your navigator classes will be even more relevant. Contrary to the FormNavigator, the DiagramNavigator and the NodeNavigator will have some serious job to do.

doc/iterations/2024.3/shapes/add_assertions_and_navigator_to_improve_diagram_testability.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+// Get the list of elements with a given label/targetObjectLabel regardless of their location in the diagram
+new DiagramNavigator(diagram)
+  .nodesWithLabel("label");
+
+// Get the unique element with a given label regardless of its location in the diagram
+new DiagramNavigator(diagram)
+  .nodeWithLabel("label")
+
+// Get the direct parent and the ancestors of a node
```
sbegaudeau: This is a part of the API where you'll start having problem I think. Since a node does not know its parent, you won't be able to move up easily. I'm pretty sure that you will end up with an API like this:

```
var diagramNavigator = new DiagramNavigator(diagram); // Navigate and index the whole diagram
var nodeNavigator = diagramNavigator.nodeWithLabel("stuff") // returns a new node navigator with new cache data
Node parentNode = nodeNavigator.findParentNode(); // Use the data cached to find the parent node because you can't find it otherwise
```

I don't know what the exact API will be but we could end up with something like that:

```
public class DiagramNavigator {
    private final Diagram diagram;
    private final Map<String, List<Node>> nodeId2Ancestors;
    private final Map<String, List<Node>> nodeDescriptionId2Nodes;
    private final Map<String, List<Edge>> nodeId2Edges;
    private final Map<String, Node> edgeId2SourceNode;
    private final Map<String, Node> edgeId2TargetNode;

    public DiagramNavigator(Diagram diagram) {
      this.diagram = Objects.requireNonNull(diagram);
      // Analyze the diagram to populate the cache
    }
}
```

In my mind, when we have to provide an answer to a "find" method, the response must be accessible instantaneously, we will have massive diagrams and I want to be able to perform tons of queries on them without having to navigate the diagram again and again. At the same time, we may not compute everything in DiagramNavigator and some additional cache may be computed only when we switch to another {Node|Edge}Navigator.

doc/iterations/2024.3/shapes/add_assertions_and_navigator_to_improve_diagram_testability.adoc
```diff
@@ -0,0 +1,108 @@
...
+  .nodeWithLabel("parent")
+  .borderNode()
+  .insideLabel();
+```
+
+Methods to access `DiagramElement` should provide multiple ways to filter the element(s) to access:
+
+- Using the description
+- Using the element's label
+- Using the target object label
```
sbegaudeau: Yes, completely agree and I think we could end up with similar recommendations as in Spring Data with query methods naming `findByXxxAndYyy(String xxx, String yyy)` etc.

sbegaudeau (approved):

PR #3281, opened 2024-03-25
[doc] Add ADR on model migration

mcharfadi:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,6 +17,8 @@
 - [ADR-141] Factorize GraphQL queries in tests
 - [ADR-142] Add report on document upload
 - [ADR-143] Handle documents upload with other formats than JSON and XMI
+- [ADR-144] Add the support for EMF models migration
+
 
 === Deprecation warning
 
--- a/doc/adrs/144_add_support_for_emf_models_migration.adoc
+++ b/doc/adrs/144_add_support_for_emf_models_migration.adoc
@@ -1,0 +1,98 @@
+= ADR-144 - Add the support for EMF models migration
+
+== Context
+
+We want to be able to load any model created with a previous version of Sirius Web. 
+
+== Decision
+
+Since models are serialized / deserialized with EMF, we will use the `ExtendedMetaData` mechanism from EMF.
+
+== Consequences
+
+We will extend `ExtendedMetaData` with another class  `MigrationExtendedMetaData` that will be instantiated with the current version of the model expected by Sirius Web and a list of `IMigrationParticipant`.
+
+This class will also override the necessary methods from `BasicExtendedMetaData` that will allow us to migrate the model (like `getLocalElement`)
+
+One important thing to consider is that BasicExtendedMetaData methods assume that the model is stored in XMl format, since we're using JSON format, we should be really cautious about using the default methods and might have to override most of them even if we don't use them all for migrating, a BasicJSonExtendedMetaData might even be needed.
+
+The version of the last migration used on a model will be stored at the root of the JSON document on EMFJson side.
+
+[source,java]
+----
+public class MigrationExtendedMetaData extends BasicExtendedMetaData {
+
+    private final String version;
+
+    private final List<IMigrationParticipant> migrationParticipants;
+
+    public ViewExtendedMetaData(List<IMigrationParticipant> migrationParticipants, String version) {
+        this.migrationParticipants = Objects.requireNonNull(migrationParticipants);
+        this.version = Objects.requireNonNull(version);
+    }
+
+    private boolean isCandidateVersion(IMigrationParticipant migrationParticipant) {
+        return migrationParticipant.getVersion().compareTo(this.version) < 0;
+    }
+
+    @Override
+    public EStructuralFeature getLocalElement(EClass eClass,  String namespace, String eStructuralFeatureName) {
+        EStructuralFeature structuralFeature = eClass.getEStructuralFeature(eStructuralFeatureName);
+        var migrationParticipants = iMigrationParticipants.stream()
+                .filter(this::isCandidateVersion)
+                .sorted(Comparator.comparing(IMigrationParticipant::getVersion))
+                .toList();
+
+        for (IMigrationParticipant migrationParticipant : migrationParticipants) {
+            structuralFeature = migrationParticipant.getStructuralFeature(eClass, eStructuralFeatureName);
+        }
+
+        return structuralFeature;
+    }
+}
+----
+
+In each method overridden, we call the corresponding method from `IMigrationParticipant` if it has a lesser version than the one used to instantiate the `MigrationExtendedMetaData`.
+
+[source,java]
+----
+public interface IMigrationParticipant {
+
+    String getVersion();
+
+    EStructuralFeature getStructuralFeature(EClass eClass, String eStructuralFeatureName);
+}
+----
+
+
+This is an example of implementation where this NameFeatureChangelMigrationParticipant will be used if the model version is superior to `202404221458`.
+If the feature name was changed from "name2" to something else, we provide the new correct value.
+
+[source,java]
+----
+@Service
+public class NameFeatureChangelMigrationParticipant implements IMigrationParticipant {
+
+    private static final String PARTICIPANT_VERSION = "202404221458";
+
+    @Override
+    public String getVersion() {
+        return PARTICIPANT_VERSION;
+    }
+
+    @Override
+    public EStructuralFeature getStructuralFeature(EClass eClass, String eStructuralFeatureName) {
+        EStructuralFeature feature = null;
+        if (DomainPackage.eINSTANCE.getDomain().equals(eClass)) {
+            if (eStructuralFeatureName.equals("name2")) {
+                feature = DomainPackage.eINSTANCE.getNamedElement_Name();
+            }
+        }
+        return feature;
+    }
+}
+----
+
+== Status
+
+Work in progress


doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+One important thing to consider is that BasicExtendedMetaData methods assume that the model is stored in XMl format, since we're using JSON format, we should be really cautious about using the default methods and might have to override most of them even if we don't use them all for migrating, a BasicJSonExtendedMetaData might even be needed.
+
+[source,java]
+----
+public class MigrationExtendedMetaData extends BasicExtendedMetaData {
+    private final String version;
+    private final List<IMigrationParticipant> iMigrationParticipants;
+    public MigrationExtendedMetaData(List<IMigrationParticipant> iMigrationParticipants, String version) {
+        this.iMigrationParticipants = Objects.requireNonNull(iMigrationParticipants);
+        this.version = version;
```
sbegaudeau: The version should be required too, we can initialize it to something arbitrary old for existing models lacking a version

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public EStructuralFeature getLocalElement(EClass eClass,  String namespace, String oldName) {
+        EStructuralFeature structuralFeature = eClass.getEStructuralFeature(oldName);
+        var migrationParticipants = iMigrationParticipants.stream().
+                filter(iMigrationParticipant -> iMigrationParticipant.canHandle(this.version))
```
sbegaudeau: Just use the version provided by the migration participant and check that its above the version currently consider. No need for a canHandle method here.

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+We will extend `ExtendedMetaData` with another class  `MigrationExtendedMetaData` that will be instantiated with the current version of the model expected by Sirius Web and a list of `IMigrationParticipant`.
+This class will also override the necessary methods from `BasicExtendedMetaData` that will allow us to migrate the model (like `getLocalElement`)
+
+One important thing to consider is that BasicExtendedMetaData methods assume that the model is stored in XMl format, since we're using JSON format, we should be really cautious about using the default methods and might have to override most of them even if we don't use them all for migrating, a BasicJSonExtendedMetaData might even be needed.
+
+[source,java]
+----
+public class MigrationExtendedMetaData extends BasicExtendedMetaData {
+    private final String version;
+    private final List<IMigrationParticipant> iMigrationParticipants;
```
sbegaudeau: `iMigrationParticipant(s)` -> `migrationParticipant(s)`

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+    }
+}
+----
+
+In each method overridden, we call the corresponding method from `IMigrationParticipant` if it has a lesser version than the one used to instantiate the `MigrationExtendedMetaData`.
+
+[source,java]
+----
+public class MigrationExtendedMetaData extends BasicExtendedMetaData {
+    boolean canHandle(String currentFileVersion);
```
sbegaudeau: No need for every single migration participant to compare their version and the current version, you can remove this `canHandle` method

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+    String getMigrationParticipantVersion();
```
sbegaudeau: `getVersion` is enough, you know that it's a migration participant already

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+    EStructuralFeature getLocalElement(EClass eClass,  String namespace, String oldName);
```
sbegaudeau: Use meaningful names contrary to the one provided by EMF. Here, it should be something like:

```
EStructuralFeature getStructuralFeature(EClass eClass, String featureName)
```

What is the `namespace`? Is that relevant for us?

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+        return structuralFeature;
+    }
+}
+----
+
+In each method overridden, we call the corresponding method from `IMigrationParticipant` if it has a lesser version than the one used to instantiate the `MigrationExtendedMetaData`.
+
+[source,java]
+----
+public class MigrationExtendedMetaData extends BasicExtendedMetaData {
```
sbegaudeau: Shouldn't that be the signature of `IMigrationParticipant`?

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+[source,java]
+----
+@Service
+public class NameFeatureChangelMigrationParticipant implements IMigrationParticipant {
+
+    private static final String PARTICIPANT_VERSION = "202404221458";
+    @Override
+    public boolean canHandle(String currentFileVersion) {
+        return PARTICIPANT_VERSION.compareTo(currentFileVersion) < 0;
+    }
```
sbegaudeau: We don't need to code that in every migration participant

doc/adrs/142_add_support_for_emf_models_migration.adoc
sbegaudeau: You will need to ensure that we don't need to override every single methods of the interface `IMigrationParticipant`. If one needs to just rename an attribute, they should not be asked to retrieve classes, packages, etc.

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+----
+
+[source,java]
+----
+public class MigrationExtendedMetaData extends BasicExtendedMetaData {
+    boolean canHandle(String currentFileVersion);
+    String getMigrationParticipantVersion();
+    EStructuralFeature getLocalElement(EClass eClass,  String namespace, String oldName);
+}
+----
```
sbegaudeau: Duplicate? Or did I miss anything

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+        for (IMigrationParticipant migrationParticipant : migrationParticipants) {
+            structuralFeature = migrationParticipant.getLocalElement(eClass, namespace, oldName);
+        }
+
+        return structuralFeature;
+    }
+}
+----
+
+In each method overridden, we call the corresponding method from `IMigrationParticipant` if it has a lesser version than the one used to instantiate the `MigrationExtendedMetaData`.
```
sbegaudeau: You are not talking about where this version will be stored? Something like a `migration` property at the root of the JSON document would be helpful but you need to serialize it and retrieve it which will require an update to sirius-emf-json. Preferably just a simple API to add whatever we need in the JSON document as a post process for exemple (but that will require a simple ADR on the EMFJson side). Anyway, you need to store something, somewhere for view, domain and other models that indicate the last version used for the migration. We could even store the name of the migration participant used to know quickly what has been the latest migration

```
{
  "json": {
    "version": "1.0",
    "encoding": "UTF-8"
  },
  "migration": {
    "lastMigrationPerformed": "NameFeatureChangelMigrationParticipant",
    "version": "202404221458"
  },
  "ns": {},
  "content": []
}
```

doc/adrs/142_add_support_for_emf_models_migration.adoc
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public EStructuralFeature getLocalElement(EClass eClass,  String namespace, String oldName) {
+        EStructuralFeature feature = null;
+        if (DomainPackage.eINSTANCE.getDomain().equals(eClass)) {
+            if (oldName.equals("name2")) {
+                feature = DomainPackage.eINSTANCE.getNamedElement_Name();
+            }
+        }
```
sbegaudeau: Start thinking about the first concrete use case that this API should fulfill.

An attribute has disappeared (Node#labelExpression) and for those who had some expressions in this attribute, we need instead to create a new child (Node#insideLabel: InsideLabel) and set the expression on the newly created child (InsideLabel#labelExpression). I have a couple of other use cases in mind (there are a couple of attributes to rename for example but that's the first use case for the iteration).

No need to update the ADR for that use case, but when you will start implementing it, this is the first use case.

sbegaudeau (changes requested):

doc/adrs/142_add_support_for_emf_models_migration.adoc
mcharfadi: Yes indeed

doc/adrs/142_add_support_for_emf_models_migration.adoc
mcharfadi: Okay I thought I would be using the same name as the one provided by EMF. 
I also wanted to have the same methods in MigrationExtendedMetaData and the MigrationParticipant but I should only offer a few methods in MigrationParticipant to be implemented then, right ?

sbegaudeau (approved):

PR #3314, opened 2024-04-02
[3312] Make the details view more compact

### LINKED ISSUE #3312: Make the details view more compact
The widget used in the detail view are quite generous in their consumption of space. This can make the use of Sirius Web more tedious when objects have a lot of properties since the user it often required to scroll down to edit or view a widget.

This issues aims to explore some possible solution to reduce the space consumption. The mains leads are explored:

* Reduce the padding/margin inside the widgets or between them
* Make some widgets on one line (Label and widget in one line) instead of the usual two line style (Label + EOL+Widget)

adaussy:
https://github.com/eclipse-sirius/sirius-web/issues/3312

### DIFF
--- a/packages/forms/frontend/sirius-components-forms/src/groups/Group.tsx
+++ b/packages/forms/frontend/sirius-components-forms/src/groups/Group.tsx
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019, 2023 Obeo.
+ * Copyright (c) 2019, 2024 Obeo.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
--- a/packages/forms/frontend/sirius-components-forms/src/groups/Group.tsx
+++ b/packages/forms/frontend/sirius-components-forms/src/groups/Group.tsx
@@ -54,7 +54,7 @@ const useGroupStyles = makeStyles<Theme, GroupStyleProps>((theme) => ({
     display: 'flex',
     flexDirection: 'column',
     '& > *': {
-      marginBottom: theme.spacing(2),
+      marginBottom: theme.spacing(1),
     },
   },
   adaptableSections: {
@@ -121,7 +121,6 @@ export const Group = ({ editingContextId, formId, group, widgetSubscriptions, re
       </div>
     );
   }
-
   return (
     <div className={classes.group} data-testid={`group-${group.label}`}>
       <div className={classes.groupLabelAndToolbar}>
--- a/packages/forms/frontend/sirius-components-forms/src/propertysections/ButtonPropertySection.tsx
+++ b/packages/forms/frontend/sirius-components-forms/src/propertysections/ButtonPropertySection.tsx
@@ -43,6 +43,8 @@ const useStyle = makeStyles<Theme, ButtonStyleProps>((theme) => ({
     fontStyle: ({ italic }) => (italic ? 'italic' : null),
     fontWeight: ({ bold }) => (bold ? 'bold' : null),
     textDecorationLine: ({ underline, strikeThrough }) => getTextDecorationLineValue(underline, strikeThrough),
+    paddingTop: theme.spacing(0.5),
+    paddingBottom: theme.spacing(0.5),
     '&:hover': {
       backgroundColor: ({ backgroundColor }) =>
         backgroundColor ? getCSSColor(backgroundColor, theme) : theme.palette.primary.main,
--- a/packages/forms/frontend/sirius-components-forms/src/propertysections/ButtonPropertySection.tsx
+++ b/packages/forms/frontend/sirius-components-forms/src/propertysections/ButtonPropertySection.tsx
@@ -58,6 +58,11 @@ const useStyle = makeStyles<Theme, ButtonStyleProps>((theme) => ({
   icon: {
     marginRight: ({ iconOnly }) => (iconOnly ? theme.spacing(0) : theme.spacing(2)),
   },
+  propertySection: {
+    display: 'flex',
+    flexDirection: 'row',
+    gap: theme.spacing(2),
+  },
 }));
 
 export const pushButtonMutation = gql`
@@ -194,7 +199,7 @@ export const ButtonPropertySection: PropertySectionComponent<GQLButton> = ({
   };
 
   return (
-    <div>
+    <div className={classes.propertySection}>
       <PropertySectionLabel
         editingContextId={editingContextId}
         formId={formId}
--- a/packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
+++ b/packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
@@ -27,9 +27,13 @@ const useStyle = makeStyles<Theme, LabelStyleProps>((theme) => ({
     fontWeight: ({ bold }) => (bold ? 'bold' : null),
     textDecorationLine: ({ underline, strikeThrough }) => getTextDecorationLineValue(underline, strikeThrough),
     verticalAlign: 'baseline',
-    'align-items': 'center',
+    alignItems: 'center',
     display: 'flex',
-    margingLeft: theme.spacing(2),
+  },
+  propertySection: {
+    display: 'flex',
+    flexDirection: 'row',
+    gap: theme.spacing(2),
   },
 }));
 
@@ -49,7 +53,7 @@ export const LabelWidgetPropertySection: PropertySectionComponent<GQLLabelWidget
   };
   const classes = useStyle(props);
   return (
-    <div>
+    <div className={classes.propertySection}>
       <PropertySectionLabel
         editingContextId={editingContextId}
         formId={formId}
--- a/packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
+++ b/packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
@@ -26,6 +26,10 @@ const useStyle = makeStyles<Theme, LabelStyleProps>((theme) => ({
     fontStyle: ({ italic }) => (italic ? 'italic' : null),
     fontWeight: ({ bold }) => (bold ? 'bold' : null),
     textDecorationLine: ({ underline, strikeThrough }) => getTextDecorationLineValue(underline, strikeThrough),
+    verticalAlign: 'baseline',
+    'align-items': 'center',
+    display: 'flex',
+    margingLeft: theme.spacing(2),
   },
 }));
 
--- a/packages/forms/frontend/sirius-components-forms/src/propertysections/PropertySectionLabel.tsx
+++ b/packages/forms/frontend/sirius-components-forms/src/propertysections/PropertySectionLabel.tsx
@@ -22,6 +22,7 @@ const usePropertySectionLabelStyles = makeStyles((theme) => ({
     display: 'flex',
     flexDirection: 'row',
     alignItems: 'center',
+    marginRight: theme.spacing(2),
   },
   subscribers: {
     marginLeft: 'auto',
@@ -38,13 +39,18 @@ const usePropertySectionLabelStyles = makeStyles((theme) => ({
     width: theme.spacing(2),
     height: theme.spacing(2),
   },
+  typography: {
+    lineHeight: '1.5',
+  },
 }));
 
 export const PropertySectionLabel = ({ editingContextId, formId, widget, subscribers }: PropertySectionLabelProps) => {
   const classes = usePropertySectionLabelStyles();
   return (
     <div className={classes.propertySectionLabel}>
-      <Typography variant="subtitle2">{widget.label}</Typography>
+      <Typography className={classes.typography} variant="subtitle2">
+        {widget.label}
+      </Typography>
       {widget.hasHelpText ? (
         <HelpTooltip editingContextId={editingContextId} formId={formId} widgetId={widget.id} />
       ) : null}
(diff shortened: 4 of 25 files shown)


packages/forms/frontend/sirius-components-forms/src/propertysections/ButtonPropertySection.tsx
```diff
@@ -55,6 +57,10 @@ const useStyle = makeStyles<Theme, ButtonStyleProps>((theme) => ({
   icon: {
     marginRight: ({ iconOnly }) => (iconOnly ? theme.spacing(0) : theme.spacing(2)),
   },
+  div: {
```
AxelRICHARD: Could you rename this to something more specific like `propertySection` please?

packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
```diff
@@ -24,6 +24,14 @@ const useStyle = makeStyles<Theme, LabelStyleProps>((theme) => ({
     fontStyle: ({ italic }) => (italic ? 'italic' : null),
     fontWeight: ({ bold }) => (bold ? 'bold' : null),
     textDecorationLine: ({ underline, strikeThrough }) => getTextDecorationLineValue(underline, strikeThrough),
+    verticalAlign: 'baseline',
+    'align-items': 'center',
```
AxelRICHARD: `'align-items'` -> `alignItems`

packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
```diff
@@ -24,6 +24,14 @@ const useStyle = makeStyles<Theme, LabelStyleProps>((theme) => ({
...
+    display: 'flex',
+    margingLeft: theme.spacing(2),
```
AxelRICHARD: `margingLeft` -> `marginLeft`

packages/forms/frontend/sirius-components-forms/src/propertysections/ButtonPropertySection.tsx
AxelRICHARD: Could you also rename every other new `div` rules in others  `**PropertySection.tsx` please?

packages/forms/frontend/sirius-components-forms/src/propertysections/PropertySectionLabel.tsx
```diff
@@ -38,13 +39,18 @@ const usePropertySectionLabelStyles = makeStyles((theme) => ({
     width: theme.spacing(2),
     height: theme.spacing(2),
   },
+  typography: {
+    lineHeight: '1.5',
```
AxelRICHARD, line 42: It seems strange to only update the `lineHeight` property here? why didn't you change the `variant="subtitle2"` for another typography like `body2` for example? (the full list of typographies is in `theme.ts`)

packages/forms/frontend/sirius-components-forms/src/propertysections/SplitButtonPropertySection.tsx
```diff
@@ -243,7 +245,7 @@ export const SplitButtonPropertySection = ({
   };
 
   return (
-    <div>
+    <div style={{ display: 'flex', flexDirection: 'row' }}>
```
AxelRICHARD: Why did you use an inline style here instead of creating a new style rule like you did in `ButtonPropertySection.tsx` for example?

packages/forms/frontend/sirius-components-forms/src/propertysections/TextfieldPropertySection.tsx
```diff
@@ -61,6 +61,14 @@ const useStyle = makeStyles<Theme, TextfieldStyleProps>((theme) => ({
     fontWeight: ({ bold }) => (bold ? 'bold' : null),
     textDecorationLine: ({ underline, strikeThrough }) => getTextDecorationLineValue(underline, strikeThrough),
   },
+  input: {
```
AxelRICHARD, line 62: It seems very strange to add two new style rules (`input` and `textfield`) here, in addition to the existing one (`style`).
Did you try to put the padding/margin directly in the `style` rule?

packages/forms/frontend/sirius-components-widget-reference/src/components/ValuedReferenceAutocomplete.tsx
```diff
@@ -265,6 +265,12 @@ export const ValuedReferenceAutocomplete = ({
           placeholder={placeholder}
           error={widget.diagnostics.length > 0}
           helperText={widget.diagnostics[0]?.message}
+          inputProps={{
```
AxelRICHARD: It seems strange to add a new `inputProps` while there is already an `InputProps` property declared on the component.
Did you try to add your new style properties in the existing `InputProps`?

packages/forms/frontend/sirius-components-forms/src/groups/Group.tsx
AxelRICHARD, line 1: Could you please update the changelog file?

AxelRICHARD, line 1: Could you please please fix the typos in your commits messages? Or juste remove the messages?

packages/forms/frontend/sirius-components-forms/src/propertysections/PropertySectionLabel.tsx
```diff
@@ -22,6 +22,7 @@ const usePropertySectionLabelStyles = makeStyles((theme) => ({
     display: 'flex',
     flexDirection: 'row',
     alignItems: 'center',
+    marginRight: theme.spacing(2),
```
sbegaudeau: A component cannot make hypothesis on the location where it will be used. Perform the layout on the parent not the children components. In this case, I suspect that a `gap: theme.spacing(2)` is missing in the places where you have added a `display: 'flex', flexDirection: 'row'`

packages/forms/frontend/sirius-components-forms/src/propertysections/ButtonPropertySection.tsx
```diff
@@ -55,6 +57,10 @@ const useStyle = makeStyles<Theme, ButtonStyleProps>((theme) => ({
...
+    display: 'flex',
+    'flex-direction': 'row',
```
sbegaudeau: why not `flexDirection`?

packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
```diff
@@ -24,6 +24,14 @@ const useStyle = makeStyles<Theme, LabelStyleProps>((theme) => ({
...
+  },
+  div: {
+    display: 'flex',
+    'flex-direction': 'row',
```
sbegaudeau: why not `flexDirection`?

sbegaudeau (dismissed):

packages/forms/frontend/sirius-components-forms/src/propertysections/ButtonPropertySection.tsx
adaussy: Done.

packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
adaussy: Done

packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
adaussy: Done

packages/forms/frontend/sirius-components-forms/src/propertysections/PropertySectionLabel.tsx
adaussy: I have look at  https://v4.mui.com/components/typography/#typography for other style of typographie. I wanted to keep a "bold" style because it makes it clear this is the label of the widget so I kept "subtitle2". If you prefer I can use "body2" but I think I would make it bold. Do you think it is better?

packages/forms/frontend/sirius-components-forms/src/propertysections/SplitButtonPropertySection.tsx
adaussy: Here the class variable is an array because it creates a class for each button of the split button.
In my case I want to apply a style the container div so I have only one instance.
If you have a better implementation I gladly change it.

packages/forms/frontend/sirius-components-widget-reference/src/components/ValuedReferenceAutocomplete.tsx
adaussy: Yes I tried but the "style" is applied to the main "div" of the component. In my case I wanted to reduce the space of the input element.

Current implementation
image

With style class modification
image

packages/forms/frontend/sirius-components-forms/src/groups/Group.tsx
adaussy: Done.

packages/forms/frontend/sirius-components-forms/src/propertysections/SplitButtonPropertySection.tsx
AxelRICHARD: Ok I see your problem.
You could have created a second makeStyle variable with only one style containing `display: 'flex', flexDirection: 'row',`.

adaussy: Yes I will do it mainly due to comment of Stephanne on the margin/gap of the PropertyLabelSection

packages/forms/frontend/sirius-components-forms/src/propertysections/PropertySectionLabel.tsx
AxelRICHARD: Ok, let's keep it that way.

packages/forms/frontend/sirius-components-forms/src/propertysections/TextfieldPropertySection.tsx
adaussy: Yes I tried but the "style" is applied to the main "div" of the component. In my case I wanted to reduce the space of the input element.

Current implementation
image

With style class modification
image

packages/forms/frontend/sirius-components-forms/src/propertysections/PropertySectionLabel.tsx
adaussy: Done. But in order to avoid getting the weird empty space when button or label widgets are used without a LabelExpression

image

I also modified the PropertySectionLabel to return a *null* component when it has nothing to display.


```typescript
export const PropertySectionLabel = ({ editingContextId, formId, widget, subscribers }: PropertySectionLabelProps) => {
  const classes = usePropertySectionLabelStyles();

  if (!widget.label && subscribers.length == 0 && !widget.hasHelpText) {
    return null;
  } else {
    return (
      <div className={classes.propertySectionLabel}>
        <Typography className={classes.typography} variant="subtitle2">
          {widget.label}
        </Typography>
        {widget.hasHelpText ? (
          <HelpTooltip editingContextId={editingContextId} formId={formId} widgetId={widget.id} />
        ) : null}
        <div className={classes.subscribers}>
          {subscribers.map((subscriber) => (
            <Tooltip title={subscriber.username} arrow key={subscriber.username}>
              <Avatar classes={{ root: classes.avatar }}>{subscriber.username.substring(0, 1).toUpperCase()}</Avatar>
            </Tooltip>
          ))}
        </div>
      </div>
    );
  }
};
```

What do you thinlk about that? It means that this behavior will also apply on all other widgets. For example for a TextField with no LabelExpression, NoSubscribers and No Help Expression

image

packages/forms/frontend/sirius-components-forms/src/propertysections/PropertySectionLabel.tsx
adaussy: Ok

packages/forms/frontend/sirius-components-forms/src/propertysections/SplitButtonPropertySection.tsx
adaussy: Done

packages/forms/frontend/sirius-components-widget-reference/src/components/ValuedReferenceAutocomplete.tsx
adaussy: Same reason than above.
The _InputProps_ are the props of the _div_ owning the _input_ whereas _inputProps_ are the props used for the _input_ element itself.

packages/forms/frontend/sirius-components-forms/src/propertysections/ButtonPropertySection.tsx
adaussy: Done.

packages/forms/frontend/sirius-components-forms/src/propertysections/LabelWidgetPropertySection.tsx
adaussy: Done.

packages/forms/frontend/sirius-components-forms/src/propertysections/TextfieldPropertySection.tsx
AxelRICHARD: Ok

packages/forms/frontend/sirius-components-widget-reference/src/components/ValuedReferenceAutocomplete.tsx
AxelRICHARD: Ok

packages/forms/frontend/sirius-components-forms/src/propertysections/PropertySectionLabel.tsx
sbegaudeau: I don't mind, a component can make the decision to do nothing if it has nothing to do. It would be better than every single component having to say `{widget.label ? <PropertySectionLabel ... /> : null}`

AxelRICHARD (approved):

PR #3355, opened 2024-04-09
[doc] Add shape on default visibility and visibility configuration

gdaniel:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/iterations/2024.5/shapes/configurable_visibility.adoc
+++ b/doc/iterations/2024.5/shapes/configurable_visibility.adoc
@@ -1,0 +1,56 @@
+= Configurable visibility
+
+== Problem
+
+Sirius Web doesn't provide a way for specifiers to set the default visibility of a node (hidden/visible/faded). 
+It also doesn't provide a way to let specifiers define tools that hide/reveal/fade/unfade nodes as part of other operations (e.g. to reveal a hidden node when a child node is created).
+Nodes are always created visible and unfaded, and the only way to hide/fade them right now is that the end user does it manually.
+
+== Key Result
+
+- Specifiers must be able to define NodeDescriptions that are hidden/faded/visible by default.
+- Specifiers must be able to call hide/reveal/fade/unfade services in their tool to hide/reveal/fade/unfade a set of nodes.
+
+== Solution
+
+A new field _Is Hidden By Default Expression_ should be displayed in NodeDescription. 
+The field contains an boolean expression which is evaluated on the first rendering of the node to set its initial hidden/visible state.
+If the expression is empty it is evaluated to `false`, which means that the node is visible by default (this preserves Sirius Web's default behavior to show new nodes by default). 
+
+A new field _Is Faded By Default Expression_ should be displayed in NodeDescription.
+The field contains a boolean expression which is evaluated on the first rendering of the node to set its initial faded/unfaded state.
+If the expression is empty it is evaluated to `false`, which means that the node is not faded by default (this preserves Sirius Web's default behavior to not fade new nodes by default).
+
+A node that is faded and not visible won't be displayed.
+Revealing a faded node keeps it faded.
+
+The `diagramServices` variable should contain new services to hide/reveal/fade/unfade nodes.
+
+== Scenario
+
+1. Default Visibility State
+- A specifier adds a NodeDescription in the view DSL
+- The _Is Hidden By Default Expression_ field is available and its default value is empty, which means visible
+- The _Is Faded By Default Expression_ field is available and its default value is empty, which means unfaded (i.e. normal)
+- The specifier can define an AQL expression that returns a boolean value to set whether a node is hidden by default (`true` for hidden, `false` for visible)
+- The specifier can define an AQL expression that returns a boolean value to set whether a node is faded by default (`true` for faded, `false` for unfaded)
+- When an end user creates a node represented in this manner in a diagram, the node is hidden/visible/faded/unfaded according to these expression
+- If the user manually hides/reveals/fades/unfades the node the value returned by the expression is not taken into account (the expression can't override user-defined state).
+
+2. Hide/Reveal/Fade/Unfade operation
+- A specifier creates a tool in the view DSL
+- The specifier creates an operation, for example ChangeContext
+- The specifier fills the expression of its operation with a call to the hide/reveal/fade/unfade service, for example `diagramServices.hide(nodeId)`
+- When an end user uses the tool defined this way, the appropriate nodes are hidden/revealed/faded/unfaded
+
+=== Cutting backs
+
+Nothing identified for now
+
+== Rabbit holes
+
+Nothing identified for now
+
+== No-gos
+
+Nothing identified for now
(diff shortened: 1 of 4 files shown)


AxelRICHARD (approved):

AxelRICHARD (approved):

doc/iterations/2024.5/shapes/configurable_visibility.adoc
```diff
@@ -0,0 +1,47 @@
+= Configurable visibility
```
sbegaudeau, line 1: Since the hidden is part of a trio of options (normal, faded, hidden) this topic is a bit more than just hidden / visible, faded must be part of the mix since it will be requested and it's a core part of this topic.

doc/iterations/2024.5/shapes/configurable_visibility.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+== Problem
+
+Sirius Web doesn't provide a way for specifiers to set the default visibility of a node (hidden/visible). 
+It also doesn't provide a way to let specifiers define tools that hide/reveal nodes as part of other operations (e.g. to reveal a hidden node when a child node is created).
+Nodes are always created visible, and the only way to hide them right now is that the end user does it manually.
+
+== Key Result
+
+- Specifiers must be able to define NodeDescriptions that are hidden/visible by default.
```
sbegaudeau: must be able to configure the state of node by default, hidden/faded/visible

doc/iterations/2024.5/shapes/configurable_visibility.adoc
```diff
@@ -0,0 +1,47 @@
...
+- Specifiers must be able to call hide/reveal services in their tool to hide/reveal a set of nodes.
```
sbegaudeau: fade too

doc/iterations/2024.5/shapes/configurable_visibility.adoc
```diff
@@ -0,0 +1,47 @@
...
+The field contains an boolean expression which is evaluated on the first rendering of the node to set its initial hidden/visible state.
+If the expression is empty it is evaluated to `false`, which means that the node is visible by default (this preserves Sirius Web's default behavior to show new nodes by default). 
+
+The `diagramServices` variable should contain new services to hide/reveal nodes.
+
+== Scenario
+
+1. Default Visibility State
+- A specifier adds a NodeDescription in the view DSL
+- The _Is Hidden By Default Expression_ field is available and its default value is empty, which means visible
```
sbegaudeau, line 33: This is where I agree, we need something simple and `isHiddenByDefaultExpression` asking for a boolean is simple. I just think we need a `isFadedByDefaultExpression` asking also for a boolean and if both are true, hidden wins. We can't ask specifiers to return an enumeration literal.

doc/iterations/2024.5/shapes/configurable_visibility.adoc
```diff
@@ -0,0 +1,47 @@
...
+- The specifier can define an AQL expression to that returns a boolean value to set the default visibility of the node (`true` for hidden, `false` for visible)
+- When an end user creates a node represented in this manner in a diagram, the node is hidden/visible according to this expression
+- If the user manually hides/reveals the node the value returned by the expression is not taken into account (the expression can't override user-defined state).
+
+2. Hide/Reveal operation
+- A specifier creates a tool in the view DSL
+- The specifier creates an operation, for example ChangeContext
+- The specifier fills the expression of its operation with a call to the hide/reveal service, for example `diagramServices.hide(nodeId)`
```
sbegaudeau: I would add fade too

doc/iterations/2024.5/shapes/configurable_visibility.adoc
```diff
@@ -0,0 +1,47 @@
...
+- When an end user uses the tool defined this way, the appropriate nodes are hidden/revealed
+
+=== Cutting backs
+
+This shape doesn't define any support for faded elements.
```
sbegaudeau: And this is the part where I will disagree since it does not bring any additional complexity at all :)

sbegaudeau (changes requested):

doc/iterations/2024.5/shapes/configurable_visibility.adoc
gdaniel: Done, I propose 2 duos of services:
- hide/reveal that only deals with the hidden aspect
- fade/unfade that only deals with the faded aspect

Let me know if the names don't work for you. There might be a confusion for reveal, which is used in Sirius Web for both faded and hidden elements.
image

doc/iterations/2024.5/shapes/configurable_visibility.adoc
gdaniel: removed

doc/iterations/2024.5/shapes/configurable_visibility.adoc
gdaniel: done

doc/iterations/2024.5/shapes/configurable_visibility.adoc
gdaniel: fixed

doc/iterations/2024.5/shapes/configurable_visibility.adoc
gdaniel: fixed

doc/iterations/2024.5/shapes/configurable_visibility.adoc
gdaniel: Added faded in the shape / ADR

doc/iterations/2024.5/shapes/configurable_visibility.adoc
sbegaudeau: Well I think there should be three services since there are only three states and not two different states that can be toggled. A node can be visible, hidden or faded. Unfade and reveal would do the same thing which is set the node back to the normal state. By having two ways to do the same thing, the specifier would have to figure out if the node is current hidden or faded to find out which service to use.

Something like this would be easier to use:

- hide(...)
- fade(...)
- show(...)

`show` would restore back a diagram element to its normal state regardless of whether it was faded or hidden because it does not make any difference.

gdaniel: As you said, a `ViewModifier` has only 3 modes, but a node can have multiple `ViewModifiers`. We can fade a node then hide it, and if we reveal it via the "reveal hidden elements" tool on the diagram it will make it visible, but still faded. This wouldn't be possible with a single `show(...)` method.

FadeAndHide

sbegaudeau: > As you said, a ViewModifier has only 3 modes, but a node can have multiple ViewModifiers.

Yes indeed, I forgot about that part -.-"

I'm not sure that a set of view modifiers was the best to do that but that's the API we have so yes, your proposal with two sets of operations is better, let's go for that. I'll have one last look at the shape and merge it if there are no other issues

sbegaudeau (approved):

PR #3387, opened 2024-04-15
[3304] Align the properties of transient and persisted representations

### LINKED ISSUE #3304: Align the properties of transient and persisted representations
We should remove `ISemanticRepresentation` and move `targetObjectId` to `IRepresentation`. We should be careful since `ISemanticRepresentation` is used at some places as the way to find out if a representation is transient or not, the issue #3303 should be considered first to stop relying on this.

pcdavid:
For now the changes are in separate steps, probably to be squashed into a single one once we're happy with the end result.

- [3304] Deprecate ISemanticRepresentation and pull getTargetObjectId up
- [3304] Remove ISemanticRepresentation


packages/trees/backend/sirius-components-trees/src/main/java/org/eclipse/sirius/components/trees/renderer/TreeRenderer.java
```diff
@@ -54,7 +55,9 @@ public Tree render() {
         }
 
         return Tree.newTree(treeId)
+                .targetObjectId(targetObjectId)
                 .descriptionId(this.treeDescription.getId())
+                .targetObjectId(label)
```
sbegaudeau: This one seems to be a mistake

sbegaudeau: Ah you've fixed it in the next commit, I'll update the first one instead.

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/form/ViewFormDescriptionConverterSwitch.java
```diff
@@ -443,6 +443,7 @@ public Optional<AbstractControlDescription> caseBarChartDescription(org.eclipse.
         BarChartDescription.Builder builder = BarChartDescription.newBarChartDescription(this.getDescriptionId(viewBarChartDescription))
                 .label(viewBarChartDescription.getName())
                 .labelProvider(this.getStringValueProvider(labelExpression))
+                .targetObjectIdProvider(variableManager -> variableManager.get(IEditingContext.EDITING_CONTEXT, IEditingContext.class).map(IEditingContext::getId).orElse(null))
```
sbegaudeau: Why not self? If a barchart is created here, it is under a group with a self.

sbegaudeau: I would even say that the targetObjectIdProvider here should be the same as the other widget since they all share a common one `.targetObjectIdProvider(this.semanticTargetIdProvider)`

sbegaudeau: I'll update both bar chart and pie chart here with that to merge the PR

CHANGELOG.adoc
```diff
@@ -56,6 +56,10 @@ They are also now used by the view builder code generator.
 - https://github.com/eclipse-sirius/sirius-web/issues/3382[#3382] [diagram] Remove the hardcoded show/hide tools in the frontend and move it to the backend.
 Specifiers now need to manually add the hide tools they want in their node/edge palette.
 New default implementations are available in `DefaultToolsFactory` to ease the migration.
+- https://github.com/eclipse-sirius/sirius-web/issues/3304[#3304] [core] The `ISemanticRepresentation` interface hase been removed.
```
sbegaudeau: typo -> "hase"

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/representations/RepresentationService.java
```diff
@@ -166,12 +165,13 @@ public void delete(UUID representationId) {
 
     @Override
     public boolean isDangling(IEditingContext editingContext, IRepresentation representation) {
-        if (representation instanceof ISemanticRepresentation semanticRepresentation) {
-            String targetObjectId = semanticRepresentation.getTargetObjectId();
-            Optional<Object> optionalObject = this.objectSearchService.getObject(editingContext, targetObjectId);
-            return optionalObject.isEmpty();
+        String targetObjectId = representation.getTargetObjectId();
+        if (Objects.equals(targetObjectId, editingContext.getId())) {
+            // Transient representation are never candidate for explicit removal
```
sbegaudeau: I'll update that since the details view, the representations view or the related elements view are all transient but they can be candidate for removal.

sbegaudeau: This also makes me realize that I've never handled dangling representations in the new runtime, I'll have to fix that before the release.

sbegaudeau (approved):

PR #3430, opened 2024-04-29
[3429] Add diagram filter view in diagram panel

### LINKED ISSUE #3429: Add a diagram filter view in the diagram panel

### Expected behavior

The diagram panel has a new icon that shows a tree representing the content of the diagram when clicked.
Elements in the tree can be checked, and a split button is available to apply some tools on the selected elements (e.g. hide, show, collapse).

A checkbox is available to (un) select all the elements in the tree at once.

The interface should look like

image

The tree and the diagram should be synchronized: an update in the diagram should update the content of the tree (in particular the icons next to each tree node item), and executing a tool from the form view should udpate the diagram accordingly.

This feature is a particular implementation of #3428 with a `FormDescription` initialized from the content of a diagram representation.

### Cuttings backs
- The split button should only support actions that have an associated event handler. In particular, accessing `DiagramServices` from the form is out of the scope of this issue.

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/3429

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?
1. Start a Sirius Web application (new architecture)
2. Create a studio
3. Open the diagram representation inside _DomainNewModel_
4. Click on the _diagram filters_ tool in the diagram panel
5. Select some nodes and apply tools on them
6. The diagram should be updated to reflect the tools you applied


packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/explorer/services/ExplorerDescriptionProvider.java
```diff
@@ -130,7 +130,7 @@ private String getExplorerTreeId(List<?> expandedObjects, List<?> activatedFilte
                 .map(id -> URLEncoder.encode(id, StandardCharsets.UTF_8))
                 .toList();
 
-        return "explorer://?expandedIds=[" + String.join(",", expandedObjectIds) + "]?&activeFilterIds=[" + String.join(",", activatedFilterIds) + "]";
+        return "explorer://?expandedIds=[" + String.join(",", expandedObjectIds) + "]&activeFilterIds=[" + String.join(",", activatedFilterIds) + "]";
```
gdaniel: This fix shouldn't be part of this PR, it fixes an issue on master when selecting elements in the explorer.
I added a quick fix here to make sure this PR can be properly tested

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanel.tsx
```diff
@@ -207,6 +211,15 @@ export const DiagramPanel = memo(
                 <UnpinIcon />
               </IconButton>
             </Tooltip>
+            {views.map((view) => {
+              return (
+                <DiagramPanelContribution
+                  view={view}
+                  editingContextId={editingContextId}
+                  diagramId={diagramId}
+                  key={view.title}></DiagramPanelContribution>
```
AxelRICHARD: Please replace `></DiagramPanelContribution>` by `/>`

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanelContribution.tsx
```diff
@@ -0,0 +1,77 @@
...
+import Paper from '@material-ui/core/Paper';
+import Popper from '@material-ui/core/Popper';
+import Tooltip from '@material-ui/core/Tooltip';
+
+import { Selection, useSelection } from '@eclipse-sirius/sirius-components-core';
+import { useRef, useState } from 'react';
+import { DiagramPanelContributionProps } from './DiagramPanelContribution.types';
+
+export const DiagramPanelContribution = ({ view, editingContextId, diagramId }: DiagramPanelContributionProps) => {
+  const [state, setState] = useState<boolean>(false);
```
AxelRICHARD: Please set meaningful names for `state` and `setState`

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/representations/services/DiagramFilterRefreshPolicyProvider.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author gdaniel
+ */
+@Service
+public class DiagramFilterRefreshPolicyProvider implements IRepresentationRefreshPolicyProvider {
+
+
+    @Override
+    public boolean canHandle(IRepresentationDescription representationDescription) {
+        return "diagram_filter_form_description".equals(representationDescription.getId());
```
AxelRICHARD: Shouldn't it be `DiagramFilterDescriptionProvider.FORM_DESCRIPTION_ID` instead of `"FORM_DESCRIPTION_ID"`?

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/explorer/services/ExplorerDescriptionProvider.java
AxelRICHARD: Ok

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/api/IDiagramFilterSplitButtonContributionProvider.java
```diff
@@ -0,0 +1,26 @@
...
+package org.eclipse.sirius.web.application.views.diagramfilter.services.api;
+
+import org.eclipse.sirius.components.forms.description.ButtonDescription;
+
+/**
+ * Provides the description of a button for the diagram filter's split button.
+ *
+ * @author gdaniel
+ */
+public interface IDiagramFilterSplitButtonContributionProvider {
```
AxelRICHARD: May be this could be renamed to something that is not related to the SplitButton, like `IDiagramFilterActionContributionProvider`? What do you think about?

packages/forms/backend/sirius-components-forms-tests/src/main/java/org/eclipse/sirius/components/forms/tests/graphql/DiagramFilterEventSubscriptionRunner.java
```diff
@@ -0,0 +1,52 @@
...
+    public DiagramFilterEventSubscriptionRunner(IGraphQLRequestor graphQLRequestor) {
+        this.graphQLRequestor = Objects.requireNonNull(graphQLRequestor);
+    }
+
+    @Override
+    public Flux<Object> run(PropertiesEventInput input) {
+        return this.graphQLRequestor.subscribe(DIAGRAM_FILTER_EVENT_SUBSCRIPTION, input);
+    }
+
+
```
AxelRICHARD: Please remove empty lines

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
AxelRICHARD, line 1: You have provided a `IDiagramFilterDescriptionProvider` bean only for the new architecture.
The result is that the current one can't start anymore because no `IDiagramFilterDescriptionProvider` bean is provided for this current architecture.
You should also provide one for this architecture or probably handle the case where there is no `IDiagramFilterDescriptionProvider` bean in `DiagramFilterEventProcessorFactory`

packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/api/DiagramFilterConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.eclipse.sirius.components.collaborative.api.IRepresentationConfiguration;
+
+/**
+ * The configuration used to create a diagram filter event processor.
+ *
+ * @author gdaniel
+ */
+public class DiagramFilterConfiguration implements IRepresentationConfiguration {
+
```
AxelRICHARD: There is problem if several users display the new filter elements menu on the same diagram at the same time.
The selection of diagram elements are synchronized but it should not be the case.
If user A select some checkboxes in his menu, user B should not view the checkboxes selected in its own menu.
the only thing that should be synchronized are the icons located after the checkboxes labels, indicating the states (hide,show,fade...) of diagram elements.

This is probably due to the fact that everybody shares the same formId here.
The frontend should probably use a random UUID when the subscription is called.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
AxelRICHARD: A third option, may be better, is to declare your `IDiagramFilterDescriptionProvider ` in `sirius-components` (besides DiagramFilterEventProcessorFactory), not in `sirius-web`.

gdaniel: I'm not sure I understand the third option: `IDiagramFilterDescriptionProvider` is already in `sirius-components` (in `org.eclipse.sirius.components.collaborative.forms.api`). 

If you refer to moving `DiagramFilterDescriptionProvider` to `sirius-components`, then we may have an issue: `DiagramDescriptionProvider` depends on both forms and diagrams. Adding it in `sirius-components-collaborative-forms` would imply to add a dependency between forms and diagrams, I don't know if this is something we want.

AxelRICHARD: I meant your `IDiagramFilterDescriptionProvider` implementations, which are currently in sirius-web

AxelRICHARD: So yes I refer to `DiagramFilterDescriptionProvider`. Ok so the third option is not the best considering your last answer.

packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/api/DiagramFilterConfiguration.java
gdaniel: I fixed it in `DiagramFilterConfiguration`, which now computes the form ID from the received object id and UUID passed by the frontend (paremeters `id` and `objectIds` in `PropertiesEventInput`).

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
gdaniel: Done, I have copied the `DiagramFilterDescriptionProvider` class in the old architecture. I also changes the `IDiagramFilterDescriptionProvider` parameter of `DiagramFilterEventProcessorFactory` to `Optional<IDiagramFilterDescriptionProvider>`. I don't know if it is the typical way to do it in sirius web, let me know if I should do it in a different way.

packages/forms/backend/sirius-components-forms-tests/src/main/java/org/eclipse/sirius/components/forms/tests/graphql/DiagramFilterEventSubscriptionRunner.java
gdaniel: done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/api/IDiagramFilterSplitButtonContributionProvider.java
gdaniel: done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/representations/services/DiagramFilterRefreshPolicyProvider.java
gdaniel: done

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanelContribution.tsx
gdaniel: done

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanel.tsx
gdaniel: done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
AxelRICHARD: It seems Ok with Optional.
It seems another option could be to use `@Autowired(required = false)` next to the parameter in the constructor, but I have not tested this way.

CHANGELOG.adoc
```diff
@@ -109,6 +109,11 @@ image:doc/screenshots/inside_outside_labels.png[Isinde outside label example, 70
 - https://github.com/eclipse-sirius/sirius-web/issues/3382[#3382] [diagram] Add a tool section in Node and Edge palettes with tools to hide/show/reset elements and their children.
 - https://github.com/eclipse-sirius/sirius-web/issues/3135[#3135] [diagram] Add new overflow strategies for node label : _wrap_ and _ellipsis_
 - https://github.com/eclipse-sirius/sirius-web/issues/3425[#3425] [diagram] Make the direction applied during _arrangeAll_ configurable.
+- https://github.com/eclipse-sirius/sirius-web/issues/3429[#3429] [diagram] Add a diagram filter view in the diagram panel.
+This view presents diagram elements in a tree and allows to select them and apply tools on the selection (e.g. hide, show, collapse).
```
sbegaudeau: > apply tools on the selection

I would rather say "update their visibility and status" or something like that because "tools" are defined by the specifier with any behavior they want. With this sentence, consumers of Sirius Web may believe that we have added the ability to execute tools on a batch of elements at once. It could be a nice addition in the future but it's not what you have done here.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanel.tsx
```diff
@@ -57,6 +60,7 @@ export const DiagramPanel = memo(
     });
 
     const { readOnly } = useContext<DiagramContextValue>(DiagramContext);
+    const { views } = useContext<DiagramPanelViewContextValue>(DiagramPanelViewContext);
```
sbegaudeau: Leverage the existing extension capabilities to add this new mechanism. See `useComponents` for example. You should end up with something like this:

```
const diagramPanelActionComponents: ComponentExtension<DiagramPanelActionProps>[] = useComponents(diagramPanelActionExtensionPoint);
```

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanel.tsx
```diff
@@ -207,6 +211,16 @@ export const DiagramPanel = memo(
...
+            {views.map((view) => {
+              return (
+                <DiagramPanelContribution
+                  view={view}
+                  editingContextId={editingContextId}
+                  diagramId={diagramId}
+                  key={view.title}
+                />
+              );
+            })}
```
sbegaudeau: Just let people contribute any component they want here but ask for one which accept specific props. Something like:

```
interface DiagramPanelActionProps {
  editingContextId: string;
  diagramId: diagramId;
}
```

You can thus remove `DiagramPanelContribution.tsx` and `DiagramPanelContribution.types.ts` from here (the code would be used specifically in your filter action) but not everybody wants that specific behavior

packages/diagrams/frontend/sirius-components-diagrams/src/contexts/DiagramPanelViewContext.ts
```diff
@@ -0,0 +1,20 @@
...
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+import React from 'react';
```
sbegaudeau: `View` has multiple very specific meaning in Sirius Web but it's mostly the root concept where representation description are created or the workbench views like the explorer or the details/properties view. Try not to use this term as much as possible, there's not reason to have anything in this contribution named view.

You don't need a new context for this feature, use the extension capabilities of `sirius-components-core`. You can remove this file entirely

packages/diagrams/frontend/sirius-components-diagrams/src/contexts/DiagramPanelViewContext.types.ts
```diff
@@ -0,0 +1,23 @@
...
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+import { WorkbenchViewComponentProps } from '@eclipse-sirius/sirius-components-core';
```
sbegaudeau: Can be removed in favor of the extension capabilities of `sirius-components-core`.  Do not use `WorkbenchViewComponentProps`, it exists to contribute new panels on the side of the workbench like the explorer. You can remove this file entirely

packages/diagrams/frontend/sirius-components-diagrams/src/index.ts
```diff
@@ -13,6 +13,8 @@
 
 export { DiagramContext } from './contexts/DiagramContext';
 export type { DiagramContextValue } from './contexts/DiagramContext.types';
+export { DiagramPanelViewContext } from './contexts/DiagramPanelViewContext';
```
sbegaudeau: Here you'll have to export the extension point instead

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanelContribution.tsx
```diff
@@ -0,0 +1,77 @@
...
+    setSelection(newSelection);
+    setIsOpen(() => true);
+  };
+
+  const anchorRef = useRef<HTMLButtonElement | null>(null);
+
+  let panel: JSX.Element | null = null;
+  if (isOpen) {
+    const PanelComponent = view.component;
+    panel = <PanelComponent editingContextId={editingContextId} readOnly={false}></PanelComponent>;
```
sbegaudeau: Since you were reusing props intended for something completely different, this readOnly props does not make sense for example. The way you did it, people can only contribute panels that will be opened and closed when you decide it, that's very restrictive.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanelContribution.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const newSelection: Selection = {
+      entries: [
+        {
+          id: diagramId,
+          label: '',
+          kind: '',
+        },
+      ],
+    };
+    setSelection(newSelection);
```
sbegaudeau: Why should opening the panel change the selection of the workbench?

packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/DiagramFilterEventProcessorFactory.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Used to create the diagram filter event processors.
+ *
+ * @author gdaniel
+ */
+@Service
+public class DiagramFilterEventProcessorFactory implements IRepresentationEventProcessorFactory {
+
+    private final Optional<IDiagramFilterDescriptionProvider> diagramFilterDescriptionProvider;
```
sbegaudeau: Do not make optional fields. This has no reason to be optional either as a field or as an argument in the constructor.

I suspect that the only reason this is optional is because it creates an issue otherwise. Let's consider that someone wants to reuse Sirius Components Forms support but without our diagrams like some of our consumers do, without  this optional field, they would have to provide a description for this feature that they do not need. On top of that something named Diagram in a forms is quite odd.

Adding diagram related stuff in forms is an issue on multiple level since one way or another it will end up with either:

- a dependency from forms to diagrams
- optional stuff that seems misplaced

A dependency from forms to diagrams is not possible since we will have the same issue the other way around when we will have to add forms in diagrams (and that's not theoretical because it has been asked multiple time) and optional stuff like that indicate that something is not in the right place.

The only scalable way to do it is to split responsibilities:

- The diagram parts will provide the necessary extension capabilities
- The form parts will provide the necessary extension capabilities (it may not needed in your case but supporting for forms in diagram it may be necessary)
- A third party depending on both forms and diagrams connect the two

In an ideal wolrld with no overhead and an infinite budget, I would say that most of the code of this feature should thus be in a dedicated `sirius-components-diagrams-filter` frontend and backend module depending on both forms and diagrams. I'm not sure that it would bring any value today to do so. As a result, this has to be in `sirius-web-application`.

Please move `DiagramFilterEventProcessorFactory`, `DiagramFilterConfiguration` and `IDiagramFilterDescriptionProvider` in `sirius-web-application` and add a new GraphQL schema file in `sirius-web-application/src/main/resources/schema/diagram-filter.graphqls` instead of adding diagram related stuff in forms.graphqls

sbegaudeau: And you will thus be able to remove the optional part of this field and constructor argument

sbegaudeau: In the near future, the code for `PropertiesEventProcessorFactory`, `RelatedElementsEventProcessorFactory` and `RepresentationsEventProcessorFactory` and their configurations, GraphQL definitions and frontend components. will move too in Sirius Web. Those are Sirius Web specific features.

packages/forms/backend/sirius-components-forms-graphql/src/main/java/org/eclipse/sirius/components/forms/graphql/datafetchers/subscription/SubscriptionDiagramFilterEventDataFetcher.java
```diff
@@ -0,0 +1,74 @@
...
+import graphql.execution.DataFetcherResult;
+import graphql.schema.DataFetchingEnvironment;
+
+/**
+ * The data fetcher used to send the refreshed diagram filters to a subscription.
+ *
+ * @author gdaniel
+ */
+@SubscriptionDataFetcher(type = "Subscription", field = "diagramFilterEvent")
+public class SubscriptionDiagramFilterEventDataFetcher implements IDataFetcherWithFieldCoordinates<Publisher<DataFetcherResult<IPayload>>> {
```
sbegaudeau: To be moved in `org.eclipse.sirius.web.application.diagram.controllers`

packages/forms/backend/sirius-components-forms-graphql/src/main/java/org/eclipse/sirius/components/forms/graphql/datafetchers/subscription/SubscriptionDiagramFilterEventDataFetcher.java
```diff
@@ -0,0 +1,74 @@
...
+        this.objectMapper = Objects.requireNonNull(objectMapper);
+        this.exceptionWrapper = Objects.requireNonNull(exceptionWrapper);
+        this.eventProcessorSubscriptionProvider = Objects.requireNonNull(eventProcessorSubscriptionProvider);
+    }
+
+    @Override
+    public Publisher<DataFetcherResult<IPayload>> get(DataFetchingEnvironment environment) throws Exception {
+        Object argument = environment.getArgument(INPUT_ARGUMENT);
+        var input = this.objectMapper.convertValue(argument, PropertiesEventInput.class);
+        var propertiesConfiguration = new DiagramFilterConfiguration(input.id().toString(), input.objectIds());
```
sbegaudeau: This is not the configuration of the properties view

packages/forms/backend/sirius-components-forms-graphql/src/main/java/org/eclipse/sirius/components/forms/graphql/datafetchers/subscription/SubscriptionDiagramFilterEventDataFetcher.java
```diff
@@ -0,0 +1,74 @@
...
+    public SubscriptionDiagramFilterEventDataFetcher(ObjectMapper objectMapper, IExceptionWrapper exceptionWrapper, IEventProcessorSubscriptionProvider eventProcessorSubscriptionProvider) {
+        this.objectMapper = Objects.requireNonNull(objectMapper);
+        this.exceptionWrapper = Objects.requireNonNull(exceptionWrapper);
+        this.eventProcessorSubscriptionProvider = Objects.requireNonNull(eventProcessorSubscriptionProvider);
+    }
+
+    @Override
+    public Publisher<DataFetcherResult<IPayload>> get(DataFetchingEnvironment environment) throws Exception {
+        Object argument = environment.getArgument(INPUT_ARGUMENT);
+        var input = this.objectMapper.convertValue(argument, PropertiesEventInput.class);
```
sbegaudeau: This should not ask for the input of the properties view but for its own

packages/forms/backend/sirius-components-forms-tests/src/main/java/org/eclipse/sirius/components/forms/tests/graphql/DiagramFilterEventSubscriptionRunner.java
```diff
@@ -0,0 +1,50 @@
...
+import reactor.core.publisher.Flux;
+
+/**
+ * Used to get the related elements event subscription with the GraphQL API.
+ *
+ * @author gdaniel
+ */
+@Service
+public class DiagramFilterEventSubscriptionRunner implements ISubscriptionRunner<PropertiesEventInput> {
```
sbegaudeau: To be moved in `sirius-web-tests`

packages/forms/frontend/sirius-components-forms/src/views/DiagramFilterView.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { DiagramFilterViewProps } from './DiagramFilterView.types';
+import { FormBasedView } from './FormBasedView';
+
+const useDiagramFilterViewStyles = makeStyles((theme) => ({
+  content: {
+    padding: theme.spacing(1),
+  },
+}));
+
+export const DiagramFilterView = (props: DiagramFilterViewProps) => {
```
sbegaudeau: To be moved in the frontend module `sirius-web-application` along with its `.types.ts` file. Don't call it a view, it's not a workbench view like the explorer or the details view. Use a name like `DiagramPanelFilterAction` instead.

packages/forms/frontend/sirius-components-forms/src/views/DiagramFilterView.tsx
```diff
@@ -0,0 +1,62 @@
+/*******************************************************************************
+ * Copyright (c) 2022, 2023 Obeo.
```
sbegaudeau: The pull request check should inform you that several files have copyright that are invalid

packages/forms/frontend/sirius-components-forms/src/views/DiagramFilterView.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+import { WorkbenchViewComponentProps } from '@eclipse-sirius/sirius-components-core';
```
sbegaudeau: Use you own props, those are not made for this

packages/forms/frontend/sirius-components-forms/src/views/DiagramFilterView.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { makeStyles } from '@material-ui/core';
+import { GQLForm, GQLWidgetSubscription } from '../form/FormEventFragments.types';
+import { Group } from '../groups/Group';
+import { useDetailsViewConfiguration } from './DetailsViewConfiguration';
```
sbegaudeau: Do not use anything from the details view, use your own stuff if needed (I'm not sure that you need any of that)

packages/view/backend/sirius-components-view-diagram-edit/src/main/resources/icons/full/obj16/BrightenTool.svg
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
sbegaudeau: None of those svg files belong in the support for diagrams in the view DSL since nothing is using them here. They should be with the pin icon.

packages/sirius-web/frontend/sirius-web-application/src/diagrams/DiagramRepresentationConfiguration.types.ts
```diff
@@ -10,16 +10,19 @@
...
   INodeConverter,
   INodeLayoutHandler,
   NodeData,
   NodeTypeContributionElement,
-} from '../../../../../diagrams/frontend/sirius-components-diagrams/dist';
+} from '@eclipse-sirius/sirius-components-diagrams';
 
 export interface DiagramRepresentationConfigurationProps {
   nodeTypeRegistry: NodeTypeRegistry;
+  diagramPanelViewRegistry?: DiagramPanelViewRegistry;
```
sbegaudeau: Use the extension registry instead of adding stuff to this mechanism that I'm trying to delete. The same comment would apply to all three files in this package

packages/sirius-web/frontend/sirius-web-application/src/diagrams/DiagramRepresentationConfiguration.types.ts
```diff
@@ -10,16 +10,19 @@
...
+
 import {
+  DiagramPanelView,
   GraphQLNodeStyleFragment,
   INodeConverter,
   INodeLayoutHandler,
   NodeData,
   NodeTypeContributionElement,
-} from '../../../../../diagrams/frontend/sirius-components-diagrams/dist';
+} from '@eclipse-sirius/sirius-components-diagrams';
```
sbegaudeau, line 20: hmmm that's odd that nobody fixed that before, good catch 👍

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterDescriptionProvider.java
```diff
@@ -0,0 +1,333 @@
...
+
+    private List<IDiagramFilterActionContributionProvider> splitButtonContributionProvider;
+
+    private DiagramFilterHelper diagramFilterHelper;
+
+    public DiagramFilterDescriptionProvider(IObjectService objectService, @Lazy IEditingContextEventProcessorRegistry editingContextEventProcessorRegistry, List<IDiagramFilterActionContributionProvider> splitButtonContributionProviders, DiagramFilterHelper diagramFilterHelper) {
+        this.objectService = Objects.requireNonNull(objectService);
+        this.editingContextEventProcessorRegistry = Objects.requireNonNull(editingContextEventProcessorRegistry);
+        this.splitButtonContributionProvider = splitButtonContributionProviders;
+        this.diagramFilterHelper = diagramFilterHelper;
```
sbegaudeau: Some people are using parts of our code without Spring. While spring will ensure that we don't end up with an odd state, I prefer that we enforce it by ourselves too. Make your fields `final` to ensure that the state is not modified and add `Objects.requireNonNull(...)` too to tell consumers of our code that `null` is not acceptable.

This comment is valid for pretty much all other services in this package.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterDescriptionProvider.java
```diff
@@ -0,0 +1,333 @@
...
+
+    private static final String PAGE_DESCRIPTION_ID = "defaultDiagramFilterPage";
+
+    private static final String FORM_TITLE = "Diagram Filters";
+
+    private final IObjectService objectService;
+
+    private IEditingContextEventProcessorRegistry editingContextEventProcessorRegistry;
+
+    private List<IDiagramFilterActionContributionProvider> splitButtonContributionProvider;
```
sbegaudeau: Since you have multiple providers, the name of the field should reflect that

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterDescriptionProvider.java
```diff
@@ -0,0 +1,333 @@
...
+
+    private DiagramFilterHelper diagramFilterHelper;
+
+    public DiagramFilterDescriptionProvider(IObjectService objectService, @Lazy IEditingContextEventProcessorRegistry editingContextEventProcessorRegistry, List<IDiagramFilterActionContributionProvider> splitButtonContributionProviders, DiagramFilterHelper diagramFilterHelper) {
```
sbegaudeau: Why should the IEditingContextEventProcessorRegistry be `@Lazy`?

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
```diff
@@ -0,0 +1,91 @@
...
+        Map<String, Boolean> checkMap = variableManager.get(DiagramFilterDescriptionProvider.SELECTED_TREE_NODES, Map.class).get();
+        return checkMap.entrySet().stream()
+                .filter(e -> e.getValue().equals(Boolean.TRUE))
+                .map(Map.Entry::getKey)
+                .collect(Collectors.toSet());
+    }
+
+    public IStatus sendDiagramEvent(VariableManager variableManager, IDiagramInput diagramInput) {
+        IRepresentationEventProcessor representationEventProcessor = variableManager.get("diagramEventProcessor", IRepresentationEventProcessor.class).get();
+        IEditingContextEventProcessor editingContextEventProcessor = variableManager.get("editingContextEventProcessor", IEditingContextEventProcessor.class).get();
```
sbegaudeau: Don't use `.get()` on optional without testing that the value is present first.

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/diagramfilter/ShowButtonDescriptionProvider.java
```diff
@@ -0,0 +1,74 @@
...
+        .diagnosticsProvider(variableManager -> List.of())
+        .kindProvider(variableManager -> "")
+        .messageProvider(variableManager -> "")
+        .styleProvider(variableManager -> ButtonStyle.newButtonStyle()
+                .backgroundColor("#ffffff")
+                .foregroundColor("#000000")
+                .build()
+        )
+        .helpTextProvider(variableManager -> "Show selected elements")
+        .build();
```
sbegaudeau: Invalid formatting, it has not the same formatting as other buttons

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterDescriptionProvider.java
```diff
@@ -0,0 +1,333 @@
...
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.web.application.views.diagramfilter.services;
```
sbegaudeau: None of the work being done here is related to the views of the workbench. All your code in Sirius Web Application should be in `sirius-web-application/src/main/java/org.eclipse.sirius.web.application.diagram.services.filter`

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
```diff
@@ -0,0 +1,91 @@
...
+import reactor.core.publisher.Sinks.Many;
+import reactor.core.publisher.Sinks.One;
+
+/**
+ * Utility methods for the "Diagram Filter" view.
+ *
+ * @author gdaniel
+ */
+@Service
+public class DiagramFilterHelper {
```
sbegaudeau: Add an interface too and use it for the dependency injection

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
```diff
@@ -0,0 +1,91 @@
...
+    public Set<String> getSelectedElementIds(VariableManager variableManager) {
+        Map<String, Boolean> checkMap = variableManager.get(DiagramFilterDescriptionProvider.SELECTED_TREE_NODES, Map.class).get();
+        return checkMap.entrySet().stream()
+                .filter(e -> e.getValue().equals(Boolean.TRUE))
+                .map(Map.Entry::getKey)
+                .collect(Collectors.toSet());
+    }
+
+    public IStatus sendDiagramEvent(VariableManager variableManager, IDiagramInput diagramInput) {
+        IRepresentationEventProcessor representationEventProcessor = variableManager.get("diagramEventProcessor", IRepresentationEventProcessor.class).get();
```
sbegaudeau: Use some constants to share the name of the variable between this class and DiagramFilterDescriptionProvider

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @author gdaniel
+ */
+@Service
+public class DiagramFilterHelper {
+
+    private final Logger logger = LoggerFactory.getLogger(DiagramFilterHelper.class);
+
+    public Set<String> getSelectedElementIds(VariableManager variableManager) {
+        Map<String, Boolean> checkMap = variableManager.get(DiagramFilterDescriptionProvider.SELECTED_TREE_NODES, Map.class).get();
```
sbegaudeau: Don't use .get() on optional without testing that the value is present first.

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/diagramfilter/PinButtonDescriptionProvider.java
```diff
@@ -0,0 +1,75 @@
...
+                })
+                .diagnosticsProvider(variableManager -> List.of())
+                .kindProvider(variableManager -> "")
+                .messageProvider(variableManager -> "")
+                .styleProvider(variableManager -> ButtonStyle.newButtonStyle()
+                        .backgroundColor("#ffffff")
+                        .foregroundColor("#000000")
+                        .build()
+                )
+                .helpTextProvider(variableManager -> "Pin selected elements")
```
sbegaudeau: Use the internationalization support for text sent back to the end user (this applies to other button description provider)

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/diagrams/DiagramFilterControllerTests.java
```diff
@@ -0,0 +1,203 @@
...
+                .filter(form -> {
+                    var diagramNavigator = new DiagramNavigator(diagramReference.get());
+                    var group = new FormNavigator(form)
+                        .page("ExpandCollapseDiagram")
+                        .group("Filter elements")
+                        .getGroup();
+                    var optTreeWidget = group.getWidgets().stream()
+                        .filter(TreeWidget.class::isInstance)
+                        .map(TreeWidget.class::cast)
+                        .findFirst();
```
sbegaudeau: =you can continue using the form navigator to retrieve the tree widget and ensure that it is present

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/BrightenButtonDescriptionProvider.java
```diff
@@ -0,0 +1,74 @@
...
+                    var nodeIds = this.diagramFilterHelper.getSelectedElementIds(variableManager);
+                    var editingContext = variableManager.get("editingContext", IEditingContext.class).get();
+                    return this.diagramFilterHelper.sendDiagramEvent(variableManager, new FadeDiagramElementInput(UUID.randomUUID(), editingContext.getId(), diagram.getId(), nodeIds, false));
+                })
+                .diagnosticsProvider(variableManager -> List.of())
+                .kindProvider(variableManager -> "")
+                .messageProvider(variableManager -> "")
+                .styleProvider(variableManager -> ButtonStyle.newButtonStyle()
+                        .backgroundColor("#ffffff")
+                        .foregroundColor("#000000")
```
sbegaudeau: Do you need to change the default style of the buttons? Because `#000000` is not the color of the text in Sirius Web and in most applications based on Sirius Web.

sbegaudeau (changes requested):
I have not executed the feature yet to play with it but the code seems nice. Given the complexity of the work that you've done, don't be surprised by the number of comments. It was expected.

CHANGELOG.adoc
gdaniel: Done

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanelContribution.tsx
gdaniel: This is a hack because `FormBasedView` uses the current selection to build the event input. We saw two solutions to address this:
- Make `objectIds` an optional parameter of `FormBasedView` to allow consumers to specify on which elements we want to open the subscription
- Select the diagram before we create the view so the current selection matches what the backend expects.

We went for the second one because we didn't want to change the API of `FormBasedView`, but we can go for another solution, including a different "FormBasedView" tailored to the DiagramFilter (see my other comment on why we use `PropertiesEventInput` as input of the subscription)

packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/DiagramFilterEventProcessorFactory.java
gdaniel: I moved the classes in `sirius-web-application`. I put them in the same package as the other classes related to this feature: `o.e.s.w.a.diagram.services.filter`. Let me know if this isn't the right place.

The optional wasn't needed anymore, it was there because I initially provided an implementation of `IDiagramFilterDescriptionProvider` for the new architecture, but not the old one. I should have removed it when I added the implementation for the old one.

This being said, I agree that it's odd to have something named "DiagramXXX" in forms. I didn't see the issue of consumer needing to provide something for a feature they don't want, because to me it looked similar for `PropertiesEventProcessorFactory` or `RelatedElementsEventProcessorFactory`, for example. Since all of them will be moved to Sirius Web I see the issue, thanks for the additional info 👍

packages/forms/backend/sirius-components-forms-graphql/src/main/java/org/eclipse/sirius/components/forms/graphql/datafetchers/subscription/SubscriptionDiagramFilterEventDataFetcher.java
gdaniel: done

packages/forms/backend/sirius-components-forms-graphql/src/main/java/org/eclipse/sirius/components/forms/graphql/datafetchers/subscription/SubscriptionDiagramFilterEventDataFetcher.java
gdaniel: done

packages/forms/backend/sirius-components-forms-graphql/src/main/java/org/eclipse/sirius/components/forms/graphql/datafetchers/subscription/SubscriptionDiagramFilterEventDataFetcher.java
gdaniel: I used `PropertiesEventInput` because in the frontend I rely on `FormBasedView` to create the view and open the subscription. The type of input in hardcoded in `FormBasedView`, so I didn't have an easy way to change it and still rely on `FormBasedView`.
To support this we can either:
- Add a props in `FormBasedView` for the input we want to use to open the subscription (this will probably have some impacts in other parts of the code)
- Create a new "FormBasedView" tailored to the diagram filter that would create the appropriate input (in this case we will have quite a lot of duplicated code between the two views)

packages/forms/backend/sirius-components-forms-tests/src/main/java/org/eclipse/sirius/components/forms/tests/graphql/DiagramFilterEventSubscriptionRunner.java
gdaniel: done

packages/view/backend/sirius-components-view-diagram-edit/src/main/resources/icons/full/obj16/BrightenTool.svg
gdaniel: done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterDescriptionProvider.java
gdaniel: Done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterDescriptionProvider.java
gdaniel: done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterDescriptionProvider.java
gdaniel: I made it lazy because of this dependency chain:
┌─────┐
|  editingContextEventProcessorRegistry 
↑     ↓
|  editingContextEventProcessorFactory
↑     ↓
|  editingContextEventProcessorFactoryParameters
↑     ↓
|  representationEventProcessorComposedFactory
↑     ↓
|  diagramFilterEventProcessorFactory
↑     ↓
|  diagramFilterDescriptionProvider
└─────┘

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
gdaniel: done

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/diagramfilter/ShowButtonDescriptionProvider.java
gdaniel: done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterDescriptionProvider.java
gdaniel: Done (and moved the interfaces in `diagram.services.filter.api`).
For the code in `sirius-web-services` I kept the package `services.diagramfilter`, mimicking `services.relatedelements`.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
gdaniel: done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
gdaniel: done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterHelper.java
gdaniel: done

packages/sirius-web/backend/sirius-web-services/src/main/java/org/eclipse/sirius/web/services/diagramfilter/PinButtonDescriptionProvider.java
gdaniel: done

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/diagrams/DiagramFilterControllerTests.java
gdaniel: I should know that 🙂
Done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/BrightenButtonDescriptionProvider.java
gdaniel: @mbats asked me to put the background in white, because the default colors for the buttons weren't working fine (see screenshot below).

image


The default foreground color for buttons is white, so I had to change it to make the text visible. I updated it to use `#261E58` to match the default text color in Sirius Web, but it will be hardcoded and won't change if the theme changes. Let me know if this is fine for you or if we need something else

packages/forms/backend/sirius-components-forms-graphql/src/main/java/org/eclipse/sirius/components/forms/graphql/datafetchers/subscription/SubscriptionDiagramFilterEventDataFetcher.java
gdaniel: We discussed it, we'll keep `PropertiesEventInput` for now

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanel.tsx
gdaniel: done

packages/diagrams/frontend/sirius-components-diagrams/src/contexts/DiagramPanelViewContext.ts
gdaniel: Renamed the contribution to "DiagramFilter", and the component containing the form to "DiagramFilterForm"

packages/diagrams/frontend/sirius-components-diagrams/src/contexts/DiagramPanelViewContext.types.ts
gdaniel: done

packages/diagrams/frontend/sirius-components-diagrams/src/index.ts
gdaniel: done

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanelContribution.tsx
gdaniel: I removed `DiagramPanelContribution` now that we use the extension mechanism provided by Sirius Web. Note that I still have a `readOnly` property in `DiagramFilter`, which is required by the underlying `FormBasedView` I use to render the form

packages/forms/frontend/sirius-components-forms/src/views/DiagramFilterView.tsx
gdaniel: Done, renamed it to "DiagramFilter" and the component containing the form to "DiagramFilterForm"

packages/forms/frontend/sirius-components-forms/src/views/DiagramFilterView.tsx
gdaniel: They are required by `FormBasedView`, I used my own props through the extension mechanism (see `DiagramPanelActionProps` in `DiagramPanel.types`), but in this particular file (which is now renamed `DiagramFilterForm`) I still need this `WorkbenchViewComponentProps` because it is required by the `FormBasedView`. 

I am maybe missing something, but I don't see how to get rid of it.

packages/forms/frontend/sirius-components-forms/src/views/DiagramFilterView.tsx
gdaniel: Removed, we didn't need the converter anyways

packages/sirius-web/frontend/sirius-web-application/src/diagrams/DiagramRepresentationConfiguration.types.ts
gdaniel: done

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanel.tsx
gdaniel: done

packages/forms/frontend/sirius-components-forms/src/views/DiagramFilterView.tsx
gdaniel: Done

gdaniel:
Added a new test in `DiagramFilterControllerTests` to improve test coverage. This implied to add:
- EditTreeCheckboxMutationRunner
- PushButtonMutationRunner
- DiagramNavigator#nodeWithId

gdaniel:
@sbegaudeau the test I added doesn't improve the coverage enough, I tested two use cases for one action (collapse), but to be closer to complete I should test all the actions that can be performed on the diagram filter (expand, hide, show, etc). All these tests will be quite similar, and may require a bit of time. Do you want me to look into it?

I also see an issue with the test coverage in `sirius-web-infrastructure`. The only change I made there is to add methods in `MessageService` to access i18n messages for help text in the split button. Existing methods in this class aren't tested yet, do you want me to do it? For the record these messages aren't visible to the end user because help text isn't supported by the split button yet.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/BrightenButtonDescriptionProvider.java
sbegaudeau: It's ok to be hardcoded to the Sirius Web theme for now. I'll just update the label of this feature because it has nothing to do with making things brighter :)

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/diagramfilter/services/DiagramFilterDescriptionProvider.java
sbegaudeau: Ok, I'll have to look into this a bit more, I'll review the other changes first.

sbegaudeau:
Ok, I've rebased the PR and reviewed your changes, there's still one or two minor details I'll have to look into but I've pushed a new version to see what the code coverage looks like on the CI. I'll fix potential code coverage issues and I'll merge this PR as soon as it's done (in a couple minutes / hours).

I think that we will have to either change some default CSS settings in the form representation or make it a bit more configurable because under some circumstances I have a "less than optimal" layout for the dialog. Some pieces of text are wrapped while they should not and I can create some overflow issues so some improvements will have to be made but we can move forward by merging the core feature first.

sbegaudeau (approved):

Issue #3574, opened 2024-06-05
No indicator that an arrangeAll action is in progress


frouene:
### Expected behavior

Add a loading indicator during arrangeAll action

PR #3635, opened 2024-06-17
[3604] Make node overlap resolution faster during "Arrange All"

### LINKED ISSUE #3604: Arrange All is extremely slow (unusable) on large diagrams
On a medium/large diagram (2831 nodes, no edges), launching an "Arrange All" essentially never finishes. On my machine (tested with Chromium and Firefox), the browser tab gets stuck eating CPU for several minutes and nothing can be done except killing it.

Under Firefox, when the page is stuck I can pause the JS and drop into the debugger, which shows this stack:

Capture d’écran du 2024-06-11 09-47-48

It should be confirmed with more tests/measures, but maybe `getIntersectingNodes` & co have bad algorithmic complexity which lead to unreasonable run times with "large" N ?

pcdavid:
- **[3604] Update CHANGELOG**
- **[3604] Only compute if nodes intersect**
- **[3604] Remove useless test**
- **[3604] Filter out non-siblings earlier**
- **[3604] Avoid creating intermediate rectangles**

On the following diagram (a simple diagram with recursive nodes for every `papaya::NamedElement`, created on the "Package java.lang" element in the "Sirius Web" Papaya example), I go from 60s to a little less than 20s.

image

Note that the first 6s or 7s seem to be spent in ELK? At least during this time I see the loading indicator added in #3574 spinning for these first seconds, and then it stops moving while my CPU is working for the rest of the duration.

The diagram is very specific and does not contain any actual overlap, so it's very possible my changes breaks some stuff. Take with a grain of salt for the moment.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -99,6 +99,7 @@ image:doc/screenshots/insideLabelPositions.png[Inside label positions, 70%]
 - https://github.com/eclipse-sirius/sirius-web/issues/3627[#3627] [form] Remove unused mutation in form
 - https://github.com/eclipse-sirius/sirius-web/issues/3606[#3606] [test] Improve error handling in ExecuteEditingContextFunctionRunner and ExecuteEditingContextFunctionEventHandler
 - https://github.com/eclipse-sirius/sirius-web/issues/3561[#3561] [diagram] Add support for background and border on diagram labels
+- https://github.com/eclipse-sirius/sirius-web/issues/3604[#3604] [diagram] Make node overlap resolution faster during "Arrange All"
 
 == v2024.5.0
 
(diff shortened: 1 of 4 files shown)


pcdavid:
After this first batch, a profiling in Chromium shows this:

Capture d’écran du 2024-06-17 17-40-19

* First (in red), about 7s waiting for the backend for the result of the `mutation layoutDiagram()` operation.
* Then (in blue), about 13s trying to detect and resolve overlaps (which do not exist in this case).

I'm wondering if the backend could pre-compute some of the required information? Even just a flag indicating if there *are* any overlap so that the front could simply avoid all of this in the (common?) case where there are none.

frouene (approved):

PR #3644, opened 2024-06-18
[DRAFT 3637] Only save the documents actually impacted by a change

### LINKED ISSUE #3637: Only save the documents actually impacted by a change
Currently on every mutation (technically any handler which publishes a `SEMANTIC_CHANGE` change description, see `EditingContextEventProcessor.shouldPersistTheEditingContext(ChangeDescription)`) we save all the documents (EMF resources) in the project.
This can be long and is blocking (because EMF resources can not be safely used by multiple threads).

Ideally we would only save the "delta" of the change itself, and this might be possible but probably require a significant effort. A middle ground that seem to be worth investigating is to only save the sub-set of documents which are impacted by the change. Of course it would only be useful for _some_ projects (with data split in multiple resources) and _some_ changes (which do not recursively impact all/most of the resources).

EMF has native support for resource modification tracking (`resource.setTrackingModification(true)` on load, and then `resource.isModified()` to test the flag and `resource.setModified(false)` to reset it on save), so identifying which resources have been directly changed is trivial.

What's less trivial (but not that complicated) is that if resource _A_ has been changed and resource _B_ has not, the change in _A_ might still have an impact on _B_'s serialisation, if _B_ contains a reference to at least one object in _A_ whose URI has been modified by the change.

I'll try to prototype such an approach just to see if it's worth.

pcdavid:


packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/ExternalReferencesAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+    public static ExternalReferencesAdapter getOrInstall(Resource res) {
+        var existingAdapter = res.eAdapters().stream().filter(ExternalReferencesAdapter.class::isInstance).map(ExternalReferencesAdapter.class::cast).findFirst();
+        if (existingAdapter.isPresent()) {
+            return existingAdapter.get();
+        } else {
+            var adapter = new ExternalReferencesAdapter();
+            res.eAdapters().add(adapter);
+            return adapter;
+        }
+    }
```
sbegaudeau, line 41: This could simply be a method in `EditingContextPersistenceService`

sbegaudeau (commented):
I can't comment on the performance improvements that this would bring but it seems interesting for the moment :+1:

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/ExternalReferencesAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public void update() {
+        if (this.getTarget() instanceof Resource res && res.getResourceSet() != null) {
+            res.getAllContents().forEachRemaining(sourceObject -> {
+                sourceObject.eCrossReferences().forEach(targetObject -> {
+                    var targetResource = targetObject.eResource();
+                    if (targetResource != res && targetResource != null) {
+                        this.references.add(targetResource.getURI());
+                    }
+                });
```
pcdavid, line 55: The list of external references/dependencies should probably be maintained on each save using `JsonResourceSerializationListener.onCrossReferenceURICreated(EObject, EReference, String)`. This would avoid this additional "eAllContents" walk here.

pcdavid:
I'm working on a new version, closing this one.

PR #3737, opened 2024-07-10
[doc] Add an ADR on allowing only coherents resize

mcharfadi:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
+++ b/doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
@@ -1,0 +1,33 @@
+= [ADR-152] Allow only coherents resizes on diagram
+
+== Context
+
+When resizing a node on a diagram it's possible for the user to set the dimension too low, this results in the end of the user interaction visually different than the actual final visual. 
+
+Also, there are issues on the layout algorithm that happen when the node is resized below his minimal size.
+
+For example a list nodes that contains 2 `iconLabel` nodes with 2 different `overflow strategy` : `None and Wrap`. Here the minimum size of the parent node cannot be inferior to the size of the sub node with a label strategy of None.
+
+By no allowing the user to interact with the nodes in a way that would cause an invalid layout (even if we try to ajust it to a correct one after), we can improve the user experience.
+
+It should also be easier to avoid invalids interactions with the diagram in the first place rather than correct them during the layout phase afterwards.
+
+== Decision
+
+We will use the props `minWidth` and `minHeigth` that are available on ReactFlow component's `NodeResizer` and `NodeResizeControl`.
+
+To get the correct value of theses properties we need to calculate them during the layout. As a remeinder, during the `layout` we render a `hidden div` and put a div per labels on the diagram inside. We then, calcultate the `BoundingClientRect` of theses div to have the `minimum size of each label`. For a label with an overflow strategy of None, we just have to add some constant like `labelVerticalPadding` to get the minimum size of the node containing the label.
+
+Doing this calcultation elsewhere (like during the conversion of a diagram) would add more complexity as we would have to redo most of the logic that is already done during the layout. Also since the temporaly div used for the calcultation is gone, we would have to wait for a first render of the nodes and then with a useffect redo some of the calculations and set the minWidth and minHeigth to the nodes, that would cause an additional rerender.
+
+To set the calculated values that we get during the layout of the diagram, we will send these when sending the `NodeLayoutData` to the backend and retreive them during the conversion of the diagram along with the whole diagram payload.
+
+== Status
+
+Work in progress
+
+== Consequences
+
+The user experience will change.
+
+We increase the size of the datas that are exchanged between the frontend and the backend.
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -10,6 +10,7 @@
 === Architectural decision records
 
 [ADR-151] Improve Diagram Subscription Lifecycle
+[ADR-152] Allow only coherents resizes on diagram
 
 === Deprecation warning
 


doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+== Context
+
+When resizing a node on a diagram it's possible for the user to set the dimension too low, this results in the end of the user interaction visually different than the actual final visual. 
+
+Also, there are issues on the layout algorithm that happen when the node is resized below his minimal size.
+
+For example a list nodes that contains 2 `iconLabel` nodes with 2 different `overflow strategy` : `None and Wrap`. Here the minimum size of the parent node cannot be inferior to the size of the sub node with a label strategy of None.
+
+By no allowing the user to interact with the nodes in a way that would cause an invalid layout (even if we try to ajust it to a correct one after), we can improve the user experience.
```
sbegaudeau, line 11: By not allowing

doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+It should also be easier to avoid invalids interactions with the diagram in the first place rather than correct them during the layout phase afterwards.
```
sbegaudeau, line 13: Sure but the layout algorithm cannot trust the input received and should always correct them anyway. Given the nature of the application with its collaborative approach, we can receive invalid inputs.

doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+== Decision
+
+We will use the props `minWidth` and `minHeigth` that are available on ReactFlow component's `NodeResizer` and `NodeResizeControl`.
+
+To get the correct value of theses properties we need to calculate them during the layout. As a remeinder, during the `layout` we render a `hidden div` and put a div per labels on the diagram inside. We then, calcultate the `BoundingClientRect` of theses div to have the `minimum size of each label`. For a label with an overflow strategy of None, we just have to add some constant like `labelVerticalPadding` to get the minimum size of the node containing the label.
```
sbegaudeau, line 19: one sentence per line please (here and in the whole document)

doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+Doing this calcultation elsewhere (like during the conversion of a diagram) would add more complexity as we would have to redo most of the logic that is already done during the layout. Also since the temporaly div used for the calcultation is gone, we would have to wait for a first render of the nodes and then with a useffect redo some of the calculations and set the minWidth and minHeigth to the nodes, that would cause an additional rerender.
```
sbegaudeau, line 21: Why can't we do that when a node is selected instead of computing the minWidth of every single label every time the diagram is refreshed? We have the node with its label in the dom for real, we can know the size of its label.

doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+To set the calculated values that we get during the layout of the diagram, we will send these when sending the `NodeLayoutData` to the backend and retreive them during the conversion of the diagram along with the whole diagram payload.
```
sbegaudeau, line 23: I still don't know why this information needs to be sent the backend. Currently what we have is:

- we need to share the layout with other subscriptions
- when the layout is computed, we send it to the backend to be shared

What you are trying to do is to use this mechanism used to share layout data in order to retrieve some values that happen to be computable during the computation of the layout. This layout sharing mechanism has not been made to compensate some lack of ability to share some data between two pieces of code in the frontend. I am still not convinced that there is a reason to send this value to every single subscriber of the diagram.

sbegaudeau (changes requested):

doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
mcharfadi: Yes we could do it like that, but that would require calling a part of the layout algorithm with each click.

We reduced the amount of things that happen when manipulating the diagram but it's still not perfect.

I'm afraid that if in the future we do more complex computation to know the correct minimum width or height it would impact the responsiveness of the app.

I could try to set this property when selecting a node why many children and see if that would cause a noticeable delay.

doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
mcharfadi: I just thought that it was a property related to the layout and happened to be more easily retrieved during the layout without adding too much computation, so it made sense to me. 

I was also afraid that putting some bit of the layout algorithm elsewhere would made the app more complex.

Are you concerned with the size of the payload ?

doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
mcharfadi: The issue I was trying to fix could have been resolved by fixing the layout after a resize but then I thought it would be a good opportunity to improve the UX

sbegaudeau: Improving the user experience is a good idea too, no issue with that

doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
sbegaudeau: > Are you concerned with the size of the payload ?

I'm not, what I'm afraid of is that we have a specific mechanism to share data collaboratively which would start getting exploited in order to compensate architectural weakness of the frontend code. I'm really not fan of such shortcut because it will get abused.

I'm still not sold on computing and sending thousands of those minWidth / minHeight to every subscription, every time a diagram is refreshed even if nobody has clicked on anything to try to resize any node. That seems highly inefficient.

doc/adrs/152_allows_only_coherents_resizes_on_diagram.adoc
sbegaudeau: I don't see how computing this information for every single label during every refresh and sending it uselessly to everybody would provide a better user experience than computing it once when needed only. I can open a diagram with 1000 nodes, not select anything at all and start changing some properties in the details view and this information will be computed and sent back and forth uselessly countless of time.

mcharfadi: Yes I have in mind that in the futur we will send only what changed but since we're not here yet you convinced me 100% to compute this information when selecting a node.

PR #3764, opened 2024-07-11
[doc] Add shape for adding tree description in view dsl

jerome-obeo:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/iterations/2024.9/add_tree_representation_view_model.adoc
+++ b/doc/iterations/2024.9/add_tree_representation_view_model.adoc
@@ -1,0 +1,46 @@
+= Add tree representation in the view DSL
+
+== Problem
+
+Sirius Web doesn't provide a way for specifiers to specify a tree representation using the view DSL.
+
+== Key Result
+
+* Specifiers must be able to create a new tree representation to describe how elements should be displayed and how they are organized in a tree structure.
+* Tree elements should be able to represent:
+** EMF based semantic elements
+** Other kind of elements (e.g. EMF Resources, Sirius Web representations or any kind of Java based objects)
+
+== Solution
+
+Define a Tree representation that will manage _all elements_ of the tree using the following specified AQL expressions:
+
+* a way to compute a *label*: `labelExpression` (`self` is the object associated to a tree element),
+* a way to compute an *icon* on both side of the label with possibly *overlays*: `beginIconExpression` and `endIconExpression` (`self` is a object associated to the tree element),
+* a way to compute *top level objects* of the tree representation: `elementsExpression` (using `editingContext` for the explorer or providing `self` in case the tree representation is created on a specific semantic element)
+* a way to compute an *id* : `idExpression` (`self` is the object associated to a tree element),
+* a way to compute an *object* associated to a tree element: `objectExpression` (`id` is the identifier of the element computed by the `idExpression`)
+* a way to compute the direct descendant *children*: `childrenExpression` (`self` is the object associated to a tree element),
+* a way to compute the *parent* object: `parentExpression` (`self` is the object associated to a tree element)
+
+The specifier is in charge of providing services for the expressions above that can handle all tree elements.
+
+== Scenario
+
+Using such definition, it should be possible for a specifier to:
+
+- Create a persisted tree representation like a regular representation on a semantic element
+- Contribute a definition of the explorer using a tree description based on the view DSL
+- Create a tree widget in a form using the tree description
+
+=== Cutting backs
+
+Nothing identified for now
+
+== Rabbit holes
+
+Nothing identified for now
+
+== No-gos
+
+Nothing identified for now
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -4,6 +4,7 @@
 
 === Shapes
 
+- Add tree representation in the view DSL
 
 === Architectural decision records
 
@@ -31,7 +32,8 @@
 - https://github.com/eclipse-sirius/sirius-web/issues/3744[#3744] [diagram] Add support for helper lines during multi selection move.
 - https://github.com/eclipse-sirius/sirius-web/issues/3770[#3770] [diagram] Enable helper lines by default.
 
-== v2024.7.0 (work in progress)
+
+== v2024.7.0
 
 === Shapes
 


doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
+= Add tree representation in the view DSL
+
+== Problem
+
+Sirius Web doesn't provide a way for specifiers to specify a tree representation using the view DSL.
+
+== Key Result
+
+* Specifiers must be able to create a new tree representation to describe how elements should be displayed and how they are organized each other in a tree structure.
```
sbegaudeau: "how they are organized each other in a tree structure" -> "how they are organized in a tree structure"?

doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
...
+* Tree elements should be able to represent:
+** a meta model element (semantic element)
```
sbegaudeau: Metamodel elements are semantic elements but not all semantic elements are meta model elements. If you want to distinguish both use cases for the Tree View DSL, just use something like "EMF based semantic elements" and "other kind of elements" because it can be anything else.

doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
...
+** a virtual element which is not part of the meta model (e.g. virtual group, EMF Resource or Sirius Web representation)
```
sbegaudeau: While there's nothing wrong with 99% of this shape, there are a couple of things that need to be kept in mind:

You need to consider both the behavior of the tree representation and the behavior that we will expose in the tree view DSL. The tree view DSL may not allow users to do everything the tree representation can do but it should not impact the tree representation in a way that would prevent those alternate use cases. For example, some of our users are creating tree based representations with semantic elements which are not based on EMF. Those are perfectly valid semantic elements but they do not use a meta-model or anything related to modeling. Even more interesting, some users have only a tree of representations. In their use cases, all the semantic elements are representations. So you can't identify semantic elements as something that is using a metamodel. This definition is too narrow.

The semantic elements are just anything used as the input of the representation and we have worked a lot to ensure that non-EMF based semantic elements can be used without issues.

doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
...
+* a way to compute an *id* : `idExpression` (self is the object associated to a tree element),
+* a way to compute an *object* associated to a tree element: `objectExpression` (self is an id)
+* a way to compute the direct descendant *children*: `childrenExpression` (self is the object associated to a tree element),
+* a way to compute the *parent* object: `parentExpression` (self is the object associated to a tree element)
+
+The specifier is in charge of providing services for the expressions above that can handle all tree elements.
+
+== Scenario
+
+Once this work is done, the Sirius Web Explorer should be using this new tree representation and popular applications based on Sirius Components should be able to define a new Explorer view more adapted to their needs.
```
sbegaudeau: As I said above, there's nothing wrong here but it's a bit too limited. The tree representation will also be used to create regular representations under any semantic element of the explorer and it will also be used within a form as a tree widget. `org.eclipse.sirius.components.view.form.TreeDescription` should be part of the discussion and thus it should not be forgotten.

As a result, there are multiple ways to start with the tree representation and it will impact how we think about it.

doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
+= Add tree representation in the view DSL
```
sbegaudeau, line 1: Since this is a shape, it is the support that will be used to discuss about the feature that will be developed. As a result, most of my comments below should not make you change the shape but instead there are here to give you or others who will read this PR some additional context on this work.

doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
...
+== Key Result
+
+* Specifiers must be able to create a new tree representation to describe how elements should be displayed and how they are organized each other in a tree structure.
+* Tree elements should be able to represent:
+** a meta model element (semantic element)
+** a virtual element which is not part of the meta model (e.g. virtual group, EMF Resource or Sirius Web representation)
+
+== Solution
+
+Define a Tree representation that will manage _all elements_ of the tree using the following specified AQL expressions:
```
sbegaudeau, line 16: I think that most of what's described below is good but by creating a shape for everything at once, you are hiding some details. It's not an issue but I need to remind all those who will read this PR why some things are defined this way.

The only issue with this approach is that one some topic, you are describing how to do things on top of the code that we have today while on others you are describing how things will be done on top of some code that will only exist in the future.

doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+* a way to compute a *label*: `labelExpression` (self is the object associated to a tree element),
```
sbegaudeau: This will not be usable just like that. Since the label will not be a string but instead a structure based on `StyledString`, you will need more than that or alternatively you can just mention a simple labelExpression and create a new ADR later to detail this issue.

doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
...
+* a way to compute an *icon* on both side of the label with possibly *overlays*: `beginIconExpression` and `endIconExpression` (self is a object associated to the tree element),
```
sbegaudeau: What will be the layout of this right icon with regards to the contextual menu? It feels a bit odd.

doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
...
+* a way to compute *top level objects* of the tree representation: `elementsExpression` (self is the editing context of the representation *#to be confirmed#*)
```
sbegaudeau: `self` will be the value provided by the person creating the representation. In the case of the explorer, I'm not even sure that `self` is relevant since it is not created on one specific elements, specifiers should rely on the `editingContext` variable.

In the case of a tree created as a regular representation, `self` will be the object on which the tree is created and in a form, self can be pretty much anything too.

doc/iterations/2024.9/add_tree_representation_view_model.adoc
```diff
@@ -0,0 +1,42 @@
...
+* a way to compute an *id* : `idExpression` (self is the object associated to a tree element),
+* a way to compute an *object* associated to a tree element: `objectExpression` (self is an id)
```
sbegaudeau: This will only be needed to let specifiers create representations on objects that are not treated as semantic elements by both `IIdentityService#getId` / `IObjectSearchService#getObject`. I'm not really sure that we would expose a random identifier as `self`, this is not what people expect when manipulating `self`. A variable named `id` would be more explicit.

sbegaudeau (changes requested):
I'll update a tiny detail or two, rebase it and merge it.

doc/iterations/2024.9/add_tree_representation_view_model.adoc
sbegaudeau: Done

sbegaudeau (approved):

PR #4076, opened 2024-10-08
[doc] Add details on the shape concerning undo redo

mcharfadi:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/iterations/2024.11/add_undo_redo_support.adoc
+++ b/doc/iterations/2024.11/add_undo_redo_support.adoc
@@ -12,36 +12,70 @@ The user can undo/redo an action he performed on a representation (he can't undo
 
 == Solution
 
-Pressing ctrl+z will undo the last action performed on a representation.
-Pressing ctrl+y will redo the action that was previously undone.
+Pressing [CMD]+z will undo the last action performed on a representation.
+Pressing [CMD]+y will redo the action that was previously undone.
 
 
-=== Scenario
+=== Scenario (semantic)
 
 1. Given only one user working on a representation, the user can undo/redo the actions performed
-- An user renames an element from "Paul" to "Bernard"
-- An user press ctrl+z
-- *Expected*: the element is renamed back to "Paul" and all elements that were impacted by the change also revert back to their previous state.
-- An user press ctrl+y
-- *Expected*: the element is renamed back to "Bernard" and all elements that were impacted by the change also revert back to their previous state.
+- A user renames an element from "Paul" to "Bernard"
+- A user press [CMD]+z
+- *Expected*: the element is renamed back to "Paul" and all elements that were impacted by the change also revert to their previous state.
+- A user press [CMD]+y
+- *Expected*: the element is renamed back to "Bernard" and all elements that were impacted by the change also revert to their previous state.
 
 
 2. Several users are working on the same representation
-- An user u1 renames an element from "Paul" to "Bernard"
-- An user u2 delete the element
-- If An user u1 press ctrl+z
+- A user u1 renames an element from "Paul" to "Bernard"
+- A user u2 delete the element
+- If A user u1 press [CMD]+z
 - *Expected*: nothing will change since it would create a conflict with the change from u2
 
 
-- If An user u2 press ctrl+z
-- *Expected*: the semantic elements are reverted back to when the element was "Bernard"
+- If A user u2 press [CMD]+z
+- *Expected*: the semantic elements are reverted to when the element was "Bernard"
 
 This should work regardless of the representation.
+
+=== Scenario (representation)
+
+Representation currently used in Sirius Web contains data that cannot be deducted from semantic data. (e.g. the layout in diagram or the sort order for table).
+
+On any representation holding a state, the same scenarios as above should work.
+
+1. Several users are editing the same Portal representation
+- The user _u1_ adds three other representations inside the portal, on at a time.
+- Another user _u2_ adds a fourth representation.
+- _u1_ undoes the addition of the third representation.
+- *Expected*: The Portal representation should display representations 1, 2 and 4.
+
+2. Several users are working on the same node
+- The user _u1_ moves the node from 0,0 to 10,10
+- After receiving the refresh, the user _u2_ moves this node from 10,10 to 20,20
+- After reveiving the refresh _u1_ presses [CMD]+z
+- *Expected*: Since it would create a conflict with the change from _u2_, nothing will change and an error is sent to _u1_
+
+3. Several users are working on the same node
+- The user _u1_ moves the node from 0,0 to 10,10
+- After receiving the refresh, the user _u2_ moves this node from 10,10 to 20,20
+- After receiving the refresh, _u2_ presses [CMD]+z
+- *Expected*: the node is moved back to 10,10
+- After receiving the refresh, _u1_ presses [CMD]+z
+- *Expected*: the node is moved back to 0,0
+
+4. Several users are working on the same node 3
+- The user _u1_ moves a node from 0,0 to 10,10
+- The user _u2_ moves this node from 10,10 to 20,20
+- _u2_ presses [CMD]+z to move back the node to 10,10
+- _u1_ moves the node from 10,10 to 0,0
+- _u2_ press [CMD]+y
+- *Expected*: Since it would create a conflict with the change from _u1_, nothing will change and an error is sent to _u2_
 
 === Cutting backs
 
 The undo redo will revert all semantic changes at first.
-Other changes likes layout on a diagram will be taken care of later.
+Other changes like the layout on a diagram will be taken care of later.
 The undo/redo stack is lost when a user closes the web page.
 
 
@@ -50,6 +84,6 @@ The undo/redo stack is lost when a user closes the web page.
 Nothing identified
 
 
-== No-gos
+== No-go
 
 Nothing identified


doc/iterations/2024.11/add_undo_redo_support.adoc
```diff
@@ -38,6 +38,34 @@ Pressing ctrl+y will redo the action that was previously undone.
 
 This should work regardless of the representation.
 
+=== Scenario (serialized representation)
+
+Serialized representation currently used in Sirius Web contains a layout part that is not contained in an EMF resource.
```
sbegaudeau: Some representations contain representation-specific data which may not be related at all with layout issues. For example, a table representation may let end users change the sort order of a column to sort the column name in the alphabetical order. This is not related to the layout but it may be persisted. Using Ctrl+Z, the end user should be able to reverse this change just like one could reverse a change adding a representation in a portal or changing the size or position of a node in a diagram.

It's not about layout vs EMF, it's about semantic data and now representation data. Our most popular example of a representation persisting representation data is the diagram with the layout AND presence of nodes and edges (think about unsynchronized node for example, it's about more than their layout but also their existence) but it's more than layout data.

It's not about serialized representations either, I could want to perform an undo in an unserialized representations, it just happens that in Sirius Web we do not have such use case but it could be achieved using Sirius Components in another product.

doc/iterations/2024.11/add_undo_redo_support.adoc
```diff
@@ -38,6 +38,34 @@ Pressing ctrl+y will redo the action that was previously undone.
...
+
+On an unsynchronized representation the same scenarios as above should work.
```
sbegaudeau: It's not about unsynchronized representations. It's about representation with a state, the only way one can create that in a diagram is with an unsynchronized node/edge description but it concerns every representation with a meaningful state (diagram with unsynchronized elements, portal, table)

doc/iterations/2024.11/add_undo_redo_support.adoc
```diff
@@ -38,6 +38,34 @@ Pressing ctrl+y will redo the action that was previously undone.
...
+
+1. Several users are working on the same node 1
+- An user u1 resize a node from 10*10 to 100*100
+- An user u2 add sub nodes to the same node
+- If An user u1 press ctrl+z
+- *Expected*: the node will be not be resize to 10*10 if the sub nodes added by u2 do not fit inside, it will be resized to a default size.
```
sbegaudeau: Yes but it's a very subtle example because we will try anyway to resize it to 10*10, it just happens that we won't succeed. It's a very good example to keep in mind but it's a complex one. Try to start with an example that does not involve diagrams to have a bird eye view of the problem. The solution cannot be specific to diagrams.

Consider this, one user create a portal, adds three forms inside. Another user adds a fourth forms and then the first users undo the addition of the third forms. There will only be the forms 1, 2 and 4.

doc/iterations/2024.11/add_undo_redo_support.adoc
```diff
@@ -38,6 +38,34 @@ Pressing ctrl+y will redo the action that was previously undone.
...
+
+2. Several users are working on the same node 2
+- An user u1 move a node from 0,0 to 10,10
+- An user u2 move this node from 10,10 to 20,20
+- If An user u1 press ctrl+z
+- *Expected*: nothing will change since it would create a conflict with the change from u2
+- If An user u2 press ctrl+z
+- *Expected*: the node is moved back to 10,10 and u1 can press ctrl-z to move it back to 0,0
```
sbegaudeau: So from this example, I conclude that you are advocating for a system in which:

- we would find the impact of a change done by a mutation in representation data
- we would store this change, for example, `move node X from 0,0 to 10,10`

and when we would try to undo it, we would:

- figure out the inverse of the change
- detect if it can be applied
- apply it

I agree with the approach, now let's consider some concrete use cases to challenge that:

### First use case

Let's consider that I have a portal, I execute a mutation to add a representation, after that the representation is added to the portal. We would need to keep somewhere that change. How would you compute the change done here? By analyzing the mutation? By analyzing the difference in the two versions of the portal? In this case, the input of the mutation contains everything relevant, it's easy.


### Second use case

Now let's consider that I have a class diagram with the following description (this is directly inspired by Eclipse SysON):

```
nodeDescription Class {
  domainType: "papaya::Class"
  semanticCandidatesExpression: "aql:self.types"
  synchronization: UNSYNCHRONIZED
  childNodeDescriptions: [AllAttributes]
}

nodeDescription AllAttributes {
  domainType: "papaya::Class"
  semanticCandidatesExpression: "aql:self"
  synchronization: UNSYNCHRONIZED
  childNodeDescriptions: [Attribute]
}

nodeDescription Attribute {
  domainType: "papaya::Attribute"
  semanticCandidatesExpression: "aql:self.attributes"
  synchronization: SYNCHRONIZED
}
```

Now that consider that I have a tool which will, thanks to a Java services and only a Java service:

- create a view for an existing class
- create an attribute in said class
- create a view for the attribute container AllAttributes

We have already seen how to record and revert the semantic changes, here the creation of an attribute. How would you detect the representation change that have been made? Here, there are three new nodes in the diagram:

- A Class unsynchronized node
- An AllAttribute unsynchronized node
- An Attribute node which is synchronized

Analyzing the input won't work since it will only tell you that the tool X has been called. On top of that, everything has been done in a Java service by the specifier so you don't know what happens.

You could navigate in the whole diagram data structure of the previous and new diagram and detect the differences (here you have three new nodes). From there, you would have to consider if those nodes are synchronized or not and if they are you would then keep in mind that some create view operation has been made. For this specific use case, you could also detect it thanks to the diagram event.

After this refresh, the diagram will be sent back to the frontend, which will compute the layout and send it back to the backend using a mutation. Let's ignore this part for now.

### Third use case

Now let's consider that a couple minutes later, I will move the newly created node. The frontend will send the full new layout to the backend and you will have to compute the difference. It can be done by analyzing the layout data of the previous and new diagram. The same issue would appear with portals but the mutation is straightforward so you could figure it from the input of the mutation alone.

How would you distinguish the layout operation from someone who has explicitly performed a layout change in the frontend from a layout update in response to the execution of a tool? I think it can be done but we would need to figure out how.

doc/iterations/2024.11/add_undo_redo_support.adoc
```diff
@@ -38,6 +38,34 @@ Pressing ctrl+y will redo the action that was previously undone.
...
+
+On an unsynchronized representation the same scenarios as above should work.
+
+1. Several users are working on the same node 1
+- An user u1 resize a node from 10*10 to 100*100
+- An user u2 add sub nodes to the same node
+- If An user u1 press ctrl+z
+- *Expected*: the node will be not be resize to 10*10 if the sub nodes added by u2 do not fit inside, it will be resized to a default size.
+
+2. Several users are working on the same node 2
```
mcharfadi: I'm not sure scenarios 2 and 3 are realistically doable.
For example 2, it would require knowing that the action performed was : moving a node from 0,0 to 10,10.
Meaning for one action having the initial state and the resulting state to compare to the actual state. The resulting state would be needed to know if another user has performed an action on the same node and the initial state to undo the action.

sbegaudeau (changes requested):
Given the size of the subject, I think that supporting the undo/redo in representation would need its own shape.

I also think that in conclusion of this first analysis from the shape, the first thing to consider would be a piece of code that would:

- Take the input, the previous state of a representation and the refreshed state of the representation
- Compute the change made and store the change

I suspect that this computation and the changes done will be highly specific to each representations. For now, table, portals and diagrams seems to be the only representations concerned by this issue.

I think that to move forward, on a simple example to evaluate our approach, we could create a piece of code for portals that can output some representation changes in response to the three or four possible changes that can be made in a portal:

- add a representation
- move the representation
- resize a representation (is that another change or is it in the move part?)
- remove the representation

We could introduce something like `IRepresentationChangeEvent` in `sirius-components-core` or `sirius-components-representations` with only an id along with a `IRepresentationChangeEventRecorder`. then in `sirius-components-collaborative-portals` have something like this:

```
public class PortalChangeEventRecorder implements IRepresentationChangeEventRecorder {
    public List<IRepresentationChangeEvent> getChanges(IInput input, IRepresentation previousRepresentation, IRepresentation newRepresentation) {
        // With the portal representation, we could do that just with the input
        // We could return some IPortalChangeEvent like RepresentationAddedChangeEvent, RepresentationResizedChangeEvent etc
    }
}
```

Technically recording what has been done is exactly the goal of the events (think `IDomainEvent`) so we could consider emitting those events to let people react to them too.

Everything that I have detailed here would be of course part of the ADR and not the shape but it gives us an idea of where we could go.

sbegaudeau:
@mcharfadi do you have any improvements on this PR?

mcharfadi:
I pushed and ADR on undo redo on Portal and another one for Diagrams.
These explains how to capture changes and handling them.
I pushed a PR for an implementation of the undo redo on Portal.

There are few specific changes to handle on a Portal representation since the component used to layout views guarantee that the layout cannot be broken (with a view on another one for example)

For the diagrams, I would like to finish the basic implementation (I'm almost done) and then list all possible issues during testing, I already have a list of some specifics situations but it's way easier to find when actually manipulating the undo redo.
I did list some implementations issues I faced for the Diagram in the ADR.

Issue #3056, opened 2024-02-05
Tree items only react to click/drag on the text area


pcdavid:
When hovering on a tree item the whole line is highlighted (at least from the ">"), but cliking in (or dragging from) the "empty" zone at the right of the "..." menu does nothing.
Same when clicking just left of the icon on leaf elements (where the ">" would be): the zone is highlighted on hover but not reactive to click.

It's not visible on the video, but in the first few seconds I click on the highlighted but empty area of the item, and nothing happens.

[Capture vidéo du 2024-02-05 12-05-44.webm](https://github.com/eclipse-sirius/sirius-web/assets/10608/96f29285-8fce-4169-a64a-5b36a5eb42e7)

PR #4084, opened 2024-10-09
[1062] Allow expand/collapse of tree items without selecting them

### LINKED ISSUE #1062: It should be possible to expand/collapse items in the explorer without impacting the selection
Currently clicking on the ">" marker of an item in the explorer to expand it and see its children automatically selects it.
This can be annoying as this changes the contents of the details view, and possibly  moves things around to focus on the newly selected element if it is visible on an open diagram.
In some contexts (e.g. when details are costly to compute) this can also have performance impact.

It should be possible to expand and collapse a tree item without selecting it if the user clicks on the icon (if navigating and expanding/collapsing using the keyboard, one has to already have the item selected to expand/collapse it, so it's not really possible in this case).

pcdavid:
This was originally fixed in 2022.3.0 via commit 48913d175c88a80d0dc10fd16f51a149bcaeee58, but broken again in 2024.3.0.

This reverts "[3056] Expand the clickable/draggable zone for tree items" (commit 190b91909afd65e678f736257fe65d1f1ca641e8) which caused the regression.

I'll try and see if I can get both #1062 and #3056 working at the same time, but in the meantime if we decide #1062 is more important than the small UX improvement in #3056 we could merge this.

Bug: https://github.com/eclipse-sirius/sirius-web/issues/1062

### DIFF
--- a/packages/trees/frontend/sirius-components-trees/src/treeitems/TreeItem.tsx
+++ b/packages/trees/frontend/sirius-components-trees/src/treeitems/TreeItem.tsx
@@ -22,14 +22,18 @@ import {
 import CropDinIcon from '@mui/icons-material/CropDin';
 import React, { useEffect, useRef, useState } from 'react';
 import { makeStyles } from 'tss-react/mui';
-import { TreeItemProps, TreeItemState, PartHovered } from './TreeItem.types';
+import { PartHovered, TreeItemProps, TreeItemState } from './TreeItem.types';
 import { TreeItemAction } from './TreeItemAction';
 import { TreeItemArrow } from './TreeItemArrow';
 import { TreeItemDirectEditInput } from './TreeItemDirectEditInput';
 import { isFilterCandidate } from './filterTreeItem';
 import { useDropTreeItem } from './useDropTreeItem';
 
-const useTreeItemStyle = makeStyles()((theme) => ({
+interface TreeItemStyleProps {
+  depth: number;
+}
+
+const useTreeItemStyle = makeStyles<TreeItemStyleProps>()((theme, { depth }) => ({
   treeItemBefore: {
     height: '2px',
   },
@@ -45,6 +49,7 @@ const useTreeItemStyle = makeStyles()((theme) => ({
       borderColor: 'black',
       borderStyle: 'dotted',
     },
+    paddingLeft: `${24 * (depth - 1)}px`,
   },
   treeItemHover: {
     backgroundColor: theme.palette.action.hover,
@@ -92,10 +97,7 @@ const useTreeItemStyle = makeStyles()((theme) => ({
     fontWeight: 'bold',
   },
   ul: {
-    marginLeft: theme.spacing(3),
-  },
-  highlight: {
-    backgroundColor: theme.palette.navigation.leftBackground,
+    marginLeft: 0,
   },
 }));
 
@@ -121,7 +123,7 @@ export const TreeItem = ({
   markedItemIds,
   treeItemActionRender,
 }: TreeItemProps) => {
-  const { classes } = useTreeItemStyle();
+  const { classes } = useTreeItemStyle({ depth });
 
   const initialState: TreeItemState = {
     editingMode: false,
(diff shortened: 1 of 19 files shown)


pcdavid:
I tried to have both behaviors, and a small UI improvement:
* expanding/collapsing an item from the arrow/chevron does *not* select it;
* when hovering on an item, the whole line is highlighted, witouth the blank gap on the left side of the arrow/chevron we had before;
* the whole line (except for the chrevron, see above) is clicable to select the item, not just the arrow/icon/label/menu part

https://github.com/user-attachments/assets/f73097e7-1015-41de-9328-dcc82b57799e

sbegaudeau (changes requested):
Could you add a Cypress test to validate this because we had so many regressions regarding the selection with the explorer recently that I'd like not to have more regressions in the future.

pcdavid:
Just a rebase for now, to check I did not mess up the conflict resolution with the work on DnD.

pcdavid:
> Could you add a Cypress test to validate this because we had so many regressions regarding the selection with the explorer recently that I'd like not to have more regressions in the future.

Done.

packages/trees/frontend/sirius-components-trees/src/treeitems/TreeItem.tsx
```diff
@@ -349,20 +351,21 @@ export const TreeItem = ({
...
           onMouseEnter={() => handleMouseEnter('item')}
-          onMouseLeave={handleMouseLeave}>
+          onMouseLeave={handleMouseLeave}
+          data-testid={`${label}-fullrow`}>
           <TreeItemArrow item={item} depth={depth} onExpand={onExpand} data-testid={`${label}-toggle`} />
           <div
             ref={refDom}
             tabIndex={0}
             onKeyDown={onBeginEditing}
+            draggable={true}
```
sbegaudeau: The `={true}` can be removed

packages/trees/frontend/sirius-components-trees/src/treeitems/TreeItemArrow.tsx
```diff
@@ -28,7 +28,10 @@ const useTreeItemArrowStyle = makeStyles()(() => ({
 export const TreeItemArrow = ({ item, depth, onExpand, 'data-testid': dataTestid }: TreeItemArrowProps) => {
   const { classes } = useTreeItemArrowStyle();
   if (item.hasChildren) {
-    const onClick = () => onExpand(item.id, depth);
+    const onClick: React.MouseEventHandler = (event) => {
+      event.stopPropagation();
```
sbegaudeau: It would be better to filter what is relevant where it is necessary. I suspect that the onClick in TreeItem is performing the selection, it should either be moved just like onDragStart etc or only consider the click as relevant if it has been done on the relevant element.

packages/trees/frontend/sirius-components-trees/src/treeitems/TreeItem.tsx
```diff
@@ -363,6 +360,10 @@ export const TreeItem = ({
...
+            onClick={onClick}
```
sbegaudeau: Do you need the onClick twice?

sbegaudeau (dismissed):

packages/trees/frontend/sirius-components-trees/src/treeitems/TreeItem.tsx
pcdavid: Done.

packages/trees/frontend/sirius-components-trees/src/treeitems/TreeItemArrow.tsx
pcdavid: Done (I think).

packages/trees/frontend/sirius-components-trees/src/treeitems/TreeItem.tsx
```diff
@@ -45,6 +49,7 @@ const useTreeItemStyle = makeStyles()((theme) => ({
       borderColor: 'black',
       borderStyle: 'dotted',
     },
+    paddingLeft: `${24 * (depth - 1)}px`,
```
mcharfadi, line 52: This does not look good but most of the style is not using theme values, so I don't think it's a big issue right now.

mcharfadi (approved):

packages/trees/frontend/sirius-components-trees/src/treeitems/TreeItemArrow.tsx
mcharfadi: I also think it's good, I'll approve the PR

PR #4085, opened 2024-10-10
[doc] Add shape to improve form property section layout

lfasani:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [X] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
+++ b/doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
@@ -1,0 +1,73 @@
+= Add support for form property section layout
+
+== Problem
+
+A property section in a form is made of a label, a help icon and a widget area.
+Currently, the layout is always vertical: label and help icon are in div and the widget area is below.
+
+== Key Result
+
+It will be possible as a specifier to choose if the layout is horizontal or vertical.
+
+== Solution
+
+The property section contains 
+
+* a "label" div which includes the label and the help icon.
+* a "widget"div
+
+The content of the property section is layouted with CSS Grid.
+New attributes are added on form widget styles to specify the layout.
+It corresponds to CSS Grid options:
+
+ * gridTemplateColumns
+ * gridTemplateRows
+ * labelGridColumn : the grid options that applies to the label area.
+ It is a shortHand for `grid-column-start ` and `grid-column-end`.
+ * labelGridRow : the grid options that applies to the label area.
+ It is a shortHand for `grid-row-start ` and `grid-row-end`.
+ * widgetGridColumn : the grid options that applies to the widget area.
+ It is a shortHand for `grid-column-start ` and `grid-column-end`.
+ * widgetGridRow : the grid options that applies to the widget area.
+ It is a shortHand for `grid-row-start ` and `grid-row-end`.
+ * gap
+
+Example:
+If all the property sections are defined with `gridTemplateColumns='15em auto'` or `gridTemplateColumns='30% 70%'` , the "label" div of all the property sections will be justified.
+
+=== Forms editor
+
+The form editor should display the property section accordingly to the widget styles`.
+
+The default layout in the form editor should be unchanged.
+
+=== Scope
+
+The change will be done to the following widgets:
+* text field
+* text area
+* select
+* multi select
+* date
+* checkbox
+* perhaps more
+
+=== Default details view
+
+The default `Details` must be displayed unchanged.
+In view model, when the user had a style on a widget, the default grid values are initialized so that the default behavior `Details` is set unchanged.
+
+=== Breadboarding
+
+NA
+
+=== Cutting backs
+
+NA
+
+== Rabbit holes
+
+NA
+
+== No-gos
+
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -10,6 +10,7 @@
 - Add support for drag and drop in the explorer
 - Add custom bending points to edges
 - Add support for draggable labels
+- Improve form property section layout
 
 
 === Architectural decision records


doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
```diff
@@ -0,0 +1,35 @@
+= Add support for form property section layout
+
+== Problem
+
+A property section in a form is made of a label, a help icon and a edition area.
```
frouene: Not necessarily an edition area, for exemple `LabelWidgetPropertySection`

doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+It will be possible as a specifier to choose if the layout is horizontal or vertical.
+
+== Solution
+
+A new attribute `flexDirection` is added on form widget styles. The value is among `[row, rowReverse, column, columnReverse]`.
+
+The label and the help icon are in the same "label" div. The `flexDirection` attribute aims to layout the "label" div and the edition div.
+If `row` and `row-reverse` the "label" div will take `1fr` and the edition div `2fr`.
+That will allow to align the "label" div of all the property sections.
```
sbegaudeau: If we uses flexbox, we won't use the `fr` unit which comes from `grid`. It's either flexbox or grid.

doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+The form editor should display the proprty sectiuon accordingly to the chosen `flexDirection`
```
sbegaudeau: section

doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
```diff
@@ -0,0 +1,35 @@
...
+A property section in a form is made of a label, a help icon and a edition area.
+Currently, the layout is always vertical: label and help icon are in div and the edition area is below.
+
+== Key Result
+
+It will be possible as a specifier to choose if the layout is horizontal or vertical.
+
+== Solution
+
+A new attribute `flexDirection` is added on form widget styles. The value is among `[row, rowReverse, column, columnReverse]`.
```
frouene: One line per sentence

sbegaudeau (changes requested):

doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
frouene: For the existing exemple CheckboxPropertySection, we didn't use the same naming
`labelPlacement: 'end' | 'start' | 'top' | 'bottom';`

doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+The label and the help icon are in the same "label" div. The `flexDirection` attribute aims to layout the "label" div and the edition div.
```
frouene: one line per sentence

doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
```diff
@@ -0,0 +1,35 @@
...
+If `row` and `row-reverse` the "label" div will take `1fr` and the edition div `2fr`.
```
frouene: Look for `CheckboxPropertySection` and usage of `FormControlLabel`

frouene (dismissed):

lfasani:
Eventually, grid is not suited because it does not provide the equivalent of row-reverse and column-reverse of Flexbox.
So, instead I will use FlexBox using
* labelFlex: the property `flex` apply to the label item of the property section
* labelValue: the property `flex` apply to the value item of the property section
Note that `flex` property is a short-hand for `flex-grow `, `flex-shrink` and  `flex-basis`

sbegaudeau (dismissed):

sbegaudeau:
> Eventually, grid is not suited because it does not provide the equivalent of row-reverse and column-reverse of Flexbox. So, instead I will use FlexBox using
> 
> * labelFlex: the property `flex` apply to the label item of the property section
> * labelValue: the property `flex` apply to the value item of the property section
>   Note that `flex` property is a short-hand for `flex-grow `, `flex-shrink` and  `flex-basis`

If you are considering css options for the label, you can assign the label to a specific grid row/column with grid-row and grid-column. If you think you have what you need with flexbox, I just want to see flexbox used from the view DSL to the frontend. Don't try to convert flexbox expressions in the view DSL to grid expressions on the frontend.

And if you want to align all the labels of the widgets on one side and then all the widgets, I don't see how you could even do something like that without CSS sub grid. I've removed my positive review and I'll wait for you to complete this shape once you have completely thought this through

lfasani:
Eventually, I opted for flexbox layout. 
it is well suited for `row-reverse` and `column-reverse` and the usage of Flexbox is homogeneous with Flexboxcontainer

I did not opt for subGrid because I think it is more complicated for a user that will be the specifier who is not an expert of CSS layout. Moreove, the scope of this ticket does not cover the full property section. If using subgrid, I think it would be defined at a layer global to property sections.

doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
```diff
@@ -0,0 +1,66 @@
...
+New attributes are added on form widget styles to specify the layout.
+It corresponds to CSS flexbox options:
+
+ * flexDirection
+ * labelFlex : the flex options which is a shortHand for for `flex-grow `, `flex-shrink` and  `flex-basis`
+ * valueFlex : the flex options which is a shortHand for for `flex-grow `, `flex-shrink` and  `flex-basis`
+ * gap
+
+Example:
+If many property sections are defined with `labelFlex='0 0 150px'`, and 'flexDirection=row', the "label" div of all the property sections will be justified.
```
sbegaudeau: I really don't understand. The expect key result is to choose a horizontal or vertical layout. Now you seem to be want to expose obscure flexbox options (flex-basis, flex-shrink) which are not relevant to this goal to provide a poor way to control the layout of the second dimension.

If you want a horizontal layout with `flexDirection=row` AND add some constraints on the vertical side too, use grid. Otherwise don't add poor capabilities to control the other dimension.

Either you need to control both dimensions and in this case provide use the dedicated API (grid) or you don't and then don't provide it. You need to clarify that from the key result part directly.


If you need to control one dimension, use flexbox. If you want both use grid, just choose what you want.

sbegaudeau (changes requested):

doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
lfasani: Effectively, I need both direction.
So I started using CSS grid.
The problem is I did not find elegant solution to have the equivalent of Flexbox `row-reverse` and `column-reverse` and I need it to be iso-functional with Checkbox. (Otherwise I would not care) which already have labelPlacement attribute.

The aim is having the upper part of the following screenshot
image

sbegaudeau (approved):

doc/iterations/2024.11/add_support_for_form_property_section_layout.adoc
```diff
@@ -0,0 +1,73 @@
...
+New attributes are added on form widget styles to specify the layout.
+It corresponds to CSS Grid options:
+
+ * gridTemplateColumns
+ * gridTemplateRows
+ * labelGridColumn : the grid options that applies to the label area.
+ It is a shortHand for `grid-column-start ` and `grid-column-end`.
+ * labelGridRow : the grid options that applies to the label area.
+ It is a shortHand for `grid-row-start ` and `grid-row-end`.
+ * valueGridColumn : the grid options that applies to the value area.
```
sbegaudeau: I'm not sure that value is the best name widget would be better because it may not represent a value at all

PR #4126, opened 2024-10-23
[doc] Add ADR on pagination for table representation

frouene:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -22,6 +22,7 @@
 - [ADR-163] Adapt the Diagram Palette API to the new palette design
 - [ADR-164] Add the ability to select newly created diagram elements
 - [ADR-165] Add support for actions in tree item menu
+- [ADR-166] Support Server-Side Pagination for Table Representation
 
 === Deprecation warning
 
--- a/doc/adrs/166_support-server-side-pagination-for-table.adoc
+++ b/doc/adrs/166_support-server-side-pagination-for-table.adoc
@@ -1,0 +1,196 @@
+= ADR-166 Support Server-Side Pagination for Table Representation
+
+== Context
+
+The default pagination features of Material React Table are client-side.
+This means you have to have all of your data fetched and stored in the table all at once.
+This may not be ideal for large datasets.
+
+== Decision
+
+To address the limitations of client-side pagination, we will implement server-side pagination for Material React Table.
+This approach will allow us to fetch and display data more efficiently, especially for large datasets.
+
+=== Cursor-Based vs. Page-Based Pagination
+
+When implementing server-side pagination, there are two primary approaches to consider: cursor-based and page-based pagination.
+
+==== Page-Based Pagination
+
+===== Advantages
+* Simplicity: Page-based pagination is simpler to implement and understand.
+* User Experience: It allows users to navigate directly to specific pages.
+
+===== Disadvantages
+* Performance: For large datasets, counting the total number of rows can be resource-intensive and slow down the response time.
+
+==== Cursor-Based Pagination
+
+===== Advantages
+* Efficiency: Cursor-based pagination is generally more efficient for large datasets as it avoids the need to count the total number of rows.
+* Consistency: It ensures consistency in the results, especially when the dataset is frequently updated.
+ The cursor acts as a pointer to the last fetched item, allowing for more accurate pagination.
+* Scalability: This approach scales better with large datasets and high-frequency updates.
+
+===== Disadvantages
+* Complexity: cursor-based pagination can be more complex to implement.
+* User Experience: It may not provide a straightforward way to jump to a specific page.
+* We cannot manage multiple `rowDescription` instances because we are unable to determine the number of rows returned for each description.
+
+==== Conclusion
+
+Performance considerations led us to choose the *cursor-based* pagination approach.
+The constraint of having only one instance of `rowDescription` is not inherently problematic.
+
+=== Material React Table
+
+To enable server-side pagination, we will set `manualPagination: true` in Material React Table.
+This will allow us to take control of pagination on the backend, ensuring that we can efficiently manage large datasets and accurately display the true total number of rows.
+The pagination must be managed within our own state, and listeners must be added to handle pagination state changes, allowing the backend to be called with the updated pagination data.
+
+[source, typescript]
+----
+const [pagination, setPagination] = useState({
+  cursor,
+  size,
+});
+----
+
+=== Frontend
+
+The pagination data will be added to the `representationId` of the subscription:
+
+[source, typescript]
+----
+`${tableId}?cursor=${cursor}&size=${size}`
+----
+
+=== GraphQL
+
+* Add pagination data in `Table` type
+
+[source, graphql]
+----
+type Table {
+
+  pagination: Pagination
+}
+
+type Pagination {
+  hasPreviousPage: Boolean!
+  hasNextPage: Boolean!
+  totalRowCount: Int
+}
+----
+
+NOTE: `totalRowCount` is optional for performance reasons.
+
+NOTE: the `rows` that is passed to the table already has had the pagination logic applied.
+
+=== Backend
+
+The following variables need to be passed to the row semantic candidat expression:
+* cursor (object retrieve with the cursor param from the front)
+* size
+
+== Customize row per page options
+
+Another issue to address is the ability to define the possible values for the number of rows per page.
+To address this, we will enhance the table description capabilities by adding a list of integers corresponding to the options that will be available on the frontend.
+
+* backend:
+
+[source, java]
+----
+public final class TableDescription {
+    private PaginationOptions paginationOptions;
+}
+
+public final class PaginationOptions {
+    private List<Integer> rowsPerPageOptions;
+}
+----
+
+Note: This customization may also be available at the view DSL level.
+
+* frontend:
+
+[source, typescript]
+----
+const table = useMaterialReactTable({
+  muiPaginationProps: {
+    rowsPerPageOptions: [5, 10, 20],// data from backend
+  },
+});
+----
+
+== Status
+
+Work in progress
+
+== Consequences
+
+* All filters must be applied on the backend to update the list of rows (by default, this is handled by Material React Table, client-side).
+* As we will be sharing information at both the synchronized (e.g., filters) and non-synchronized (pagination) levels,
+we will need to address the challenge of synchronizing all representations in memory simultaneously, regardless of their pagination state.
+
+=== New AQL services for navigation
+
+Support for cursor-based backend pagination cannot be properly done without new AQL services to navigate in the graph of EObjects.
+Those XPath related services will start their life in Sirius Web,
+but they should ultimately be upstreamed in the Acceleo project in `org.eclipse.acceleo.query.services.XPathServices` directly.
+The services we will need include:
+
+* `EObject#following(int size)`: Used to retrieve the next size objects following the given one
+* `EObject#hasFollowing()`
+* `EObject#preceding(int size)`: Used to retrieve the previous size objects preceding the given one
+* `EObject#hasPreceding()`
+* `EObject#collecting(String direction, int size)`: Used to call either following or preceding depending on the value of direction which could either be backward or forward.
+
+=== Support for paginated data
+
+We will need to retrieve pagination data on top of the raw data,
+for that the return type of the `semanticCandidatesExpression` will be a PaginatedData similar to Spring Data's Page.
+[source,java]
+----
+record PaginatedData  (
+  List  content,
+  boolean hasNextPage,
+  boolean hasPreviousPage,
+  int totalElementCount
+)
+----
+Contrary to what we might initially think,
+it will not be possible for us to apply our own filters such as `DomainType` or `preconditionExpression` after a paginated `semanticCandidatesExpression`.
+We will have to provide a service to let specifiers create paginated data from a collection themselves:
+
+* `Collection#toPaginatedData(boolean hasNextPage, boolean hasPreviousPage)`
+* `Collection#toPaginatedData(boolean hasNextPage, boolean hasPreviousPage, int totalElementCount)`
+
+NOTE: Those services and data structure should not be specific to tables and probably be in core parts of Sirius Components.
+
+
+=== Support for filter, sort, etc
+
+Additional features that will impact the content or the order of the elements will have to be given to specifiers as a variable in the `semanticCandidatesExpression`.
+To provide some basic support for those variables, we will have to provide better services to consider them, probably at least something like this:
+
+* `EObject#collecting(String direction, Filter filter, Sort sort, int size)`
+
+With potentially something like this to sort the results:
+
+[source, java]
+----
+record Sort(List  entries)
+
+record SortEntry(String featureName, String sortOrder)
+----
+To be able to indicate that we want to sort a collection of individual by "Family name ascending", then "First name ascending", then "age descending".
+The final API to sort / filter may not look like this,
+but it will be critical to provide all the relevant variables in the `semanticCandidatesExpression` and maintain XPath inspired AQL services to use the feature.
+With all that in mind, specifiers planning to use our support for table should be prepared to write something like this as a `semanticCandidatesExpression`:
+[source]
+----
+aql:cursor.collecting(direction, filter, sort, size)->toPaginatedData(cursor.hasPreceding(), cursor.hasFollowing())
+----
+None of those parameters will be ignorable to return a valid result.


doc/adrs/XXX_support-server-side-pagination-for-table.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+The default pagination features of Material React Table are client-side.
+This means you have to have all of your data fetched and stored in the table all at once.
+This may not be ideal for large datasets.
+
+== Decision
+
+To address the limitations of client-side pagination, we will implement server-side pagination for Material React Table.
+This approach will allow us to fetch and display data more efficiently, especially for large datasets.
+The server-side pagination method will only be applied to table representations.
```
sbegaudeau: I would tend to disagree and use the same strategy everywhere for both consistency, performance and user experience. In a form the real estate available for a table will be very limited and even displaying a "small" list would not provide a great user experience.

It would be more costly for us to maintain both and I don't see why we should prevent end users from supporting this.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
```diff
@@ -0,0 +1,108 @@
...
+For the use of tables as form widgets, client-side pagination will continue to be used.
+
+=== Material React Table
+
+To enable server-side pagination, we will set `manualPagination: true` in Material React Table.
+This will allow us to take control of pagination on the backend, ensuring that we can efficiently manage large datasets and accurately display the true total number of rows.
+The pagination must be managed within our own state, and listeners must be added to handle pagination state changes, allowing the backend to be called with the updated pagination data.
+
+[source, typescript]
```
sbegaudeau: I really wanted to have this discussion with you. There are multiple ways to define the pagination each with specific tradeoffs. We should at least talk about our choice.

The two main strategies:

- page based (what you have described)
- cursor based

With a page based strategy I think it's easier to manipulate the API but it provides an inferior user experience. On top of that it offers inferior performances for some use cases. Let's consider that I'm iterating on everything starting from an object (which can even be the whole resource set).

´´´
aql:self.eAllContents()->subSequence(pageIndex * pageSize, pageSize)
´´´

In this case, we would still pretty much have to do an eAllContents() anyway, we wouldn't have a choice. I think that for this use case it would be faster with to start from an object with a cursor based pagination:

´´´
aql:self.eResource().getResourceSet().getEObject(cursor).nextElements(pageSize)
´´´

With something like that we would retrieve the current position in the iteration very quickly thanks to the cursor and then we would just have to find the ´pageSize´ next elements from there.

I don't see how we could let specifiers provide a performance efficient solution on this use case with a page based pagination.

On the other hand, because it's all about tradeoffs, if I want to return something like more complex, the dumb way to do it would be a less efficient with a cursor based pagination:

´´´
aql:self.friends->union(self.colleagues)->subSequence(pageIndex * pageSize, pageSize)
´´´

vs

´´´
aql:self.friends->union(self.colleagues)->subSequence(
  self.friends->union(self.colleagues)->indexOf(self.eResource().getResourceSet().getEObject(cursor)),
  pageSize
)
´´´

But we need to consider that in Java we would not have to compute the union twice, we would just keep it in a variable (which is not doable in raw aql). It would "just" require an additional ´indexOf´ which should not be that time consuming.

On the other hand, if one was able to write their expression as a computation with a starting point, it would be possible to do something like that:

´´´
aql: self.eResource().getResourceSet().getEObject(cursor).nextElementInTheUnionOfFriendsAndColleagues()
´´´

So if a specifier has the ability to compute not everything but instead all the next elements from a given point it should lower the amount of ´eAllContents´ performed because this will destroy the performances. I think one can write something like that because we could start an iterator at a given position with solething like this:

´´´
arrayList.listIterator(index)
´´´

I suspect that any serious Java service trying to support the pagination would not have to compute a collection but instead an iterator which would start at ´cursor´ or ´pageIndex * pageSize´ and return the next ´pageSize´ results. The question would now be what is more efficient: creating an iterator from a position or from the identifier of an object.

We definitely need to try to create an example of such an iterator and I have a good example in mind. Let's consider that I want to display a list of all the Papaya types with pagination. Thanks to a custom iterator I could very easily skip iterating on attributes, methods, parameters, and all their content. We need to try writing just the two versions of the tree iterator to compare the ease of use and performance.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
```diff
@@ -0,0 +1,108 @@
...
+The pagination data will be added to the `representationId` of the subscription:
+
+[source, typescript]
+----
+`${tableId}?pageIndex=${pageIndex}&pageSize=${pageSize}`
+----
+
+=== GraphQL
+
+* Add pagination data in `Table` type
```
sbegaudeau: I'm not sure we need to have this information on the table. What would be the need to this? The frontend already have this information.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+----
+
+=== Frontend
+
+The pagination data will be added to the `representationId` of the subscription:
+
+[source, typescript]
+----
+`${tableId}?pageIndex=${pageIndex}&pageSize=${pageSize}`
```
sbegaudeau: I agree but we need to keep in mind that we will end up with multiple versions of the table in memory and then when the end users will want to perform some synchronized operation like activating some filter for all users we would have to consider this input (or the change description) in multiple representations. This will have an impact in quite a unique way for future development.

We never had to handle something like this in trees for example because changes are either semantic or local to one instance.

sbegaudeau (changes requested):

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
sbegaudeau: I'm not event sure that the code of the iterator would change, it's just the code of the method used to return the tree iterator which would have to change for page or cusor based pagination.

We would also need to think about other use cases from our community to see if we are not forgetting some use cases.

frouene: I see another use case to take into account, the fact that we want to display the total number of results available.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
frouene: The frontend doesn't have the rowCount (total available result)

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
frouene: Yes, I think the pagination is the only data that we don't want to be synchronized between all the open representation.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
frouene: ok, that was to avoid addressing the problem of unsynchronized data between form representations in a first time.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
sbegaudeau: It's a very good concern to take into account, we may have to iterate on all results anyway.

Here is an example of what I can come up with for the pagination with both strategies: https://gist.github.com/sbegaudeau/04dd722cd934a496e3d2dea9ea2a64a6

sbegaudeau: But if we want to return the total count, we may be required to iterate on all potential results anyway which would make the difficulty in retrieving the starting point with a page based approach moot since we would have to use the custom iterator for the whole resource set anyway.

We could this way end up with pretty much requiring everyone to create a custom iterator (if they want good performances) to skip useless results but to always run the iterator on the whole resource set to have all results and thus compute the number of results.

It can be efficient (for example, in the case of my quick example, I can skip everything which is not a type and contained types) but it can't be perfect in the sense that we will need to iterate on more than just what we will return.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
sbegaudeau: Yes indeed, that's a good reason for this. The table representation needs to compute pagination data because it can never be sure to return the "whole table", it's just computing a subset visible by the frontend depending on the requested range.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
sbegaudeau: It's doable to handle all the other concerns but the synchronization of the filtering state for example across all instances of a specific table with various instances depending on the pagination state of each subscription will be tricky. We will need to talk about this in a future dedicated ADR. We should keep that in mind as a consequence of this ADR.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
sbegaudeau: So if we want to return the total count, we will have to iterate on everything to compute the length of the data returned. If we have to iterate on everything, we can keep having multiple rowDescriptions and execute all the `semanticCandidatesExpression` and we can simply compute both the `subList(start, end)` and the `.size()` by ourselves.

So in the end, and don't hesitate to correct me if you think I'm wrong, but I suspect that if we want to display the `totalCount`, we can just ask users for a list of elements in a semanticCandidatesExpression and we will do the rest because they won't be able to do better than us. It would be a simple API but it would condemn us to iterate on all considered elements at least once.

If I have 1 000 000 model elements and I want to display 100 000 of them in a table and the end user wants to see 10 of those at a time, during each refresh we would need at best to iterate on 100 000 elements and then execute all the other expression like a label expression etc on 10 elements.

I think we can live with this, what do you think?

We would have a simple API but not a perfect one in terms of performance.

frouene: Okay, so to sum up, we can keep a simpler API page based .
We don't expose the variables linked to pagination, and we doing the split ourselves to send only the sub list that responds to pagination.
To do this, we use the complete list retrieved via rowDescription (which gives us the total number of rows).

It's ok for me.

sbegaudeau: On the other hand if we want to maximuze performances, we would need the cursor based API with variables like cursor, direction (previous or next) and the size.

We would probably need additional aql services like EObject#preceding(int),  EObject#following(int) and EObject#next(direction, int) which would delegate yo preceding or following and some service to compute the a page object to be able write something like this

aql:self.eResource().getResourceSet().getEObject(cursor).next(direction, size)->asPage(...)

In both cases we need to find out the information needed in a page:

- size of the page?
- total count?
- hasNextPage
- hasPreviousPage

Some parameters could be optional like the total count for a maximum amount of performance.

sbegaudeau: If we can get a good version of this ´toPage´ service, I think we could go for this approach.

Does MaterialReactTable supports undefined pagination? If we don't know the total number of pages for example but we know if there is a previous / next page.

Since we can customize the user interface for the pagination, we may be able to do it.

We could also make sure that cursor is an EObject if we can resolve it, it would greatly simplify queries.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+=== Material React Table
+
+To enable server-side pagination, we will set `manualPagination: true` in Material React Table.
+This will allow us to take control of pagination on the backend, ensuring that we can efficiently manage large datasets and accurately display the true total number of rows.
+The pagination must be managed within our own state, and listeners must be added to handle pagination state changes, allowing the backend to be called with the updated pagination data.
+
+[source, typescript]
+----
+const [pagination, setPagination] = useState({
```
sbegaudeau: Since we will use cursor based pagination, could you update this code to use `cursor` and `size` instead. I suspect that this will not match the expected TypeScript API of Material React Table to you could also remove the code sample entirely since we will have to see how the API will look like when building it.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+----
+
+=== Frontend
+
+The pagination data will be added to the `representationId` of the subscription:
+
+[source, typescript]
+----
+`${tableId}?cursor=${cursor}&pageSize=${pageSize}`
```
sbegaudeau: Could you use `size` instead of `pageSize` please? It's not much but it will help clarify that we are not using page based pagination a tiny bit more.

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
```diff
@@ -0,0 +1,146 @@
...
+----
+type Table {
+
+  pagination: Pagination
+}
+
+type Pagination {
+  hasPreviousPage: Boolean!
+  hasNextPage: Boolean!
+  rowCount: Int
```
sbegaudeau: Could you use `totalRowCount` to clarify that we are not talking about the number of rows actually returned by the query. Same comment on the line below

doc/adrs/XXX_support-server-side-pagination-for-table.adoc
```diff
@@ -0,0 +1,146 @@
...
+    rowsPerPageOptions: [5, 10, 20],// data from backend
+  },
+});
+----
+
+== Status
+
+Work in progress
+
+== Consequences
```
sbegaudeau: This new version of the ADR is great, it captures the complexity of the problem but I just want to clarify a couple of things that will happen as a consequence of this ADR. It will have a lot of impact on all users of tables and on the development of future features so I want to make it very explicit.

Could you had, as consequences, something a bit like this:

=== New AQL services for the navigation

Support for a cursor-based backend pagination cannot be properly done without new AQL services to navigate in the graph of EObjects. Those [XPath](https://www.w3.org/TR/1999/REC-xpath-19991116/) related services will start their life in Sirius Web but they should ultimately be upstreamed in the Acceleo project in `org.eclipse.acceleo.query.services.XPathServices` directly. Among the services we will need, we will have:

- `EObject#following(int size)`: Used to retrieve the next `size` objects following the given one
- `EObject#hasFollowing()`
- `EObject#preceding(int size)`: Used to retrieve the previous `size` objects preceding the given one
- `EObject#hasPreceding()`
- `EObject#collecting(String direction, int size)`: Used to call either `following` or `preceding` depending on the value of `direction` which could either be `backward` or `forward`.


=== Support for paginated data

We will need to retrieve pagination data on top of the raw data, for that the return type of the `semanticCandidatesExpression` will be a `PaginatedData` not unlike Spring Data's `Page`.

```
record PaginatedData<T> (
  List<T> content,
  boolean hasNextPage,
  boolean hasPreviousPage,
  int totalElementCount
)
```

Contrary to what I told you on the other ADR for the view DSL, it will not be possible for us to apply our own filters like `DomainType` or `preconditionExpression` after a paginated `semanticCandidatesExpression`. We will have to provide a service to let specifiers create paginated data from a collection themselves:

- `Collection#toPaginatedData(boolean hasNextPage, boolean hasPreviousPage)`
- `Collection#toPaginatedData(boolean hasNextPage, boolean hasPreviousPage, int totalElementCount)`

Those services and data structure should not be specific to tables and probably be in core parts of Sirius Components.

=== Support for filter, sort, etc

Additional features that will impact the content or the order of the elements will have to be given to specifiers as a variable in the `semanticCandidatesExpression`. In order to provide some basic support for those variables, we will have to provide better services to consider them, probably at least something like this:

- `EObject#collecting(String direction, Filter filter, Sort sort, int size)`

With potentially something like this to sort the results:

```
record Sort(List<SortEntry> entries)

record SortEntry(String featureName, String sortOrder)
```

To be able to indicate that we want to sort a collection of individual by "Family name ascending", then "First name ascending", then "age descending". The final API to sort / filter may not look like this but it will be critical to provide all the relevant variables in the `semanticCandidatesExpression` and maintain XPath inspired AQL services to use the feature. With all that in mind, specifiers planning to use our support for table should be prepared to write something like this as a `semanticCandidatesExpression`:

```
aql:cursor.collecting(direction, filter, sort, size)->toPaginatedData(cursor.hasPreceding(), cursor.hasFollowing())
```

None of those parameters will be ignorable to return a valid result.

sbegaudeau (changes requested):
Just one last set of improvements and it should be good to merge :+1:

sbegaudeau (approved):

PR #4175, opened 2024-11-05
[doc] Add ADR on support for forking table description

mcharfadi:


doc/adrs/168_add_support_for_forking_table_description_from_representation.adoc
```diff
@@ -0,0 +1,48 @@
+= ADR-166 Add support for forking table description from representation
+
+== Context
+
+On a table representation, we want to be able to modify the specification used as description of this table and then reload the table with this new specification.
+
+== Decision
+
+We will allow users to open a new studio from an opened table representation. This new studio will be a fork of the one used to instantiate the table. After modifying this new studio, the user will be able to re open the previous table representation that will use the updated studio as description.
```
sbegaudeau: - one sentence per line please :)
- "This new studio will be a fork of the one used to instantiate the table."  I would say that the new studio will contain a copy of the table description used to instantiate the table. Technically, in the first use case, the original table description may be contributed programmatically and created with the view builder API. It may not come from a "studio"
- "that will use the updated studio as description" -> that will use the updated table description

doc/adrs/168_add_support_for_forking_table_description_from_representation.adoc
```diff
@@ -0,0 +1,48 @@
...
+}
+```
+
+The data fetcher will then call `this.createForkedStudioService.create(editingContextId, representationId)`
+
+This method will :
+
+* Dispose the current editing context
+* Create a new Studio that will be a fork of the description used by the representation
+* Update the representations in the database with the new descriptionId
```
sbegaudeau: update only the representation with the `representationId` given in the input

doc/adrs/168_add_support_for_forking_table_description_from_representation.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+We will add an extension point to contribute to the settings menu item.
+One contribution will add a menu item in order to send the createForkedStudio mutation.
+
+== Advantages
+* This solution does not require deep modification on how we reload representation.
+
+== Disadvantages
+* User Experience: The user that wants to make a change will need to reopen his representation after the changes.
+* User Experience: Saved changes in the database from the previous representation will be irrelevant and deleted.
```
sbegaudeau: Why? I don't see why anything should be irrelevant and deleted

sbegaudeau (changes requested):

doc/adrs/168_add_support_for_forking_table_description_from_representation.adoc
mcharfadi: I'm not sure what data we're going to store for a table and wanted to avoid possible issues.
I thought about the case where if we store the current width of a column and that column does not exist anymore after some changes on the forked studio.

doc/adrs/168_add_support_for_forking_table_description_from_representation.adoc
mcharfadi: In that case the datafetcher will not have enought informations to do everything.
I need an event handler that have access to the current editing context in order to get the original table description.

sbegaudeau (approved):

PR #4198, opened 2024-11-14
[3980] Re-enable selection after the execution of a tool on diagrams

### LINKED ISSUE #3980: Add the ability to select newly created node
- by default a newly created instance in a tool using the view DSL should be selected
- an API should be available to specify what should be selected instead

pcdavid:
This reverts commits 5fc00f3c2c322d8d6c5160271ef82dd24087f1b8 and 14f4133753af01571ef091d39641bc5029c11827 which themselves reverted the initial frontend side of the feature, and provides an updated version of the DiagramRenderer.tsx hook which should fix the issue.

I'm keeping it as a draft for the moment, it looks like there might still be issues/regressions (but it's not clear yet, some of them are sometimes reproducible on master without this change).

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -90,6 +90,9 @@ Specifiers are also encouraged to implement their own `IRestDataVersionPayloadSe
 - https://github.com/eclipse-sirius/sirius-web/issues/4210[#4210] [table] Add the ability to fork the studio used by a table representation
 - https://github.com/eclipse-sirius/sirius-web/issues/4273[#4273] [table] Add support to column filtering in table
 
+- https://github.com/eclipse-sirius/sirius-web/issues/3980[#3980] Add the ability to select newly created nodes.
+The backend part (the ability to define an _Elements to Select Expression_ on diagram tools) was added in Sirius Web 2024.11.0 but the frontend did not apply the requested selection.
+This is now fixed.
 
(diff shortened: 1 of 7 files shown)


pcdavid:
Removing the draft status, I can reproduce the "glitches" I noticed earlier on `master` (so without this PR). It seems specific to Chrome (can't reproduce on Firefox) and maybe related to focus handling.

mcharfadi (dismissed):
I found an issue when selecting a node after creating one

https://github.com/user-attachments/assets/54b7ffd7-c748-495a-b01a-33e4dfbe639c

The view model is the one from a new studio with another node with `sutherland::Entity1` as domain type

mcharfadi:
https://github.com/user-attachments/assets/f0af88c9-bd70-4b53-b39b-5dcf1561fe93

I found an issue when selecting an element after another one was created

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/useRefreshedDiagram.ts
pcdavid, line 1: This one was committed by mistake. It was an unfinished attempt to extract the large useEffect into a custom hook, but it's not functional and not actually used. I'll remove it.

AxelRICHARD (approved):

PR #4256, opened 2024-11-29
[4254] Make explorer services reusable

### LINKED ISSUE #4254: Make default explorer services reusable at the view level

### Expected behavior

The default explorer representation is hardcoded in Sirius Web at the _description_ level.
The explorer now supports multiple tree representations, which are provided as `TreeDescriptions`, potentially defined at the _view_ level.
Services used by the default explorer take `VariableManager` as parameter, which make them practically unusable from trees defined at the _view_ level.

Extracting these services would let custom explorers reuse default explorer behavior easily.
A new API could be created to expose these services:
```java
public interface IExplorerServices {
    boolean hasChildren(Object self, IEditingContext editingContext);
    List<Object> getChildren(Object self, IEditingContext editingContext, List<String> expandedIds, List<String> activeFilterIds);
   // Other services used by the explorer
}
```

Existing default explorer services would rely on this API to compute their actual result.

### Cuttings backs

Nothing identified.

### Potential side effects

The `IExplorerTreeItemAlteredContentProvider` interface (which applies filters on the explorer) requires a `VariableManager` as parameter.
Since this interface may be implemented by downstream applications, its signature won't be updated and the associated filtering behavior will stay in the existing services.
This implies that custom explorer implementations will have to handle filters explicitly.

Migrating the default explorer as a _view_ (instead of a _description_) would be another way to tackle the issue, but it is more complicated. The proposed solution is a first step in this direction.

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/4254

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?
- Existing default explorer should have the same behavior as before.
- The "Domain Explorer by DSL" explorer should have the same behavior as before.

### DIFF
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/explorer/services/ExplorerServices.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/explorer/services/ExplorerServices.java
@@ -1,0 +1,260 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.web.application.views.explorer.services;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.emf.ecore.resource.ResourceSet;
+import org.eclipse.emf.edit.domain.EditingDomain;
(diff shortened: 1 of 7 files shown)


packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/explorer/services/ExplorerServices.java
```diff
@@ -0,0 +1,254 @@
...
+import org.springframework.data.jdbc.core.mapping.AggregateReference;
+import org.springframework.stereotype.Service;
+
+/**
+ * Services used to perform operations in the explorer.
+ *
+ * @author gdaniel
+ */
+@Service
+public class ExplorerServices implements IExplorerServices {
```
AxelRICHARD, line 48: It seems this service class could also be useful for every TreeRepresentation, not only the Explorer tree.
So, may be it could be named TreeServices instead of ExplorerServices? (and also rename the associated interface)
@gdaniel @sbegaudeau What do you think about?

sbegaudeau: I think `IExplorerService` is good since it really is for those who want to clone the behavior of the explorer. There are other trees out there that do not have this behavior.

AxelRICHARD (approved):

PR #4284, opened 2024-12-11
[doc] Contribute a shape for model publication

gdaniel:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -10,11 +10,14 @@
 - Add support to dynamically update representation description
 - Add loading indicators for long actions
 - Add impact analysis before tool execution
+- Add support for artifact publication from a project
 
 
 === Architectural decision records
 
 - [ADR-178] Add support of actions in table row menu
+- [ADR-179] Add support for artifact publication from a project
+
 
 === Deprecation warning
 
--- a/doc/adrs/179_add_support_for_artifact_publication_from_a_project.adoc
+++ b/doc/adrs/179_add_support_for_artifact_publication_from_a_project.adoc
@@ -1,0 +1,72 @@
+= [ADR-179] Add support for artifact publication from a project
+
+== Context
+
+We need to be able to find all the available workflows and execute one.
+
+
+== Decision
+
+Sirius Web should provide record and interfaces similar to the examples below:
+
+```java
+public record ProjectWorkflow(
+  @NotNull String id,
+  @NotNull String label
+) {}
+
+public interface IProjectWorkflowProvider {
+  List<ProjectWorkflow> getProjectWorkflows(UUID projectId);
+}
+
+public interface IProjectWorkflowHandler {
+    /*
+     * Can use the nature of the project to figure out if the workflow is relevant
+     */
+    boolean canHandle(String projectWorkflowId);
+
+     /*
+      * I'm not providing the editing context here on purpose to let the specifier load the snapshot of the
+      * data from the database in order NOT to block the editing context event processor.
+      */
+    void handle(ICause cause, String projectWorkflowId, UUID projectId);
+}
+```
+
+It will allow Sirius Web to provide the following GraphQL API to retrieve and execute workflows:
+
+```
+type Project {
+  workflows: ProjectsWorkflowsConnection!
+}
+
+type ProjectsWorkflowsConnection {
+  edges: [ProjectWorkflowEdge!]!
+  pageInfo: PageInfo!
+}
+
+type ProjectWorkflowEdge {
+  node: ProjectWorkflow!
+}
+
+type ProjectWorkflow {
+  id: ID!
+  label: String!
+}
+
+type Mutation {
+  executeProjectWorkflow(input: ExecuteProjectWorkflowInput!): ExcecuteProjectWorkflowPayload!
+}
+
+union ExecuteProjectWorkflowPayload = ErrorPayload | SuccessPayload
+
+input ExecuteProjectWorkflowInput {
+  id: ID!
+  projectId: ID!
+  projectWorkflowId: ID!
+}
+```
+
+== Status
+
+Work in progress
--- a/doc/iterations/2025.2/add_support_for_artifact_publication_from_a_project.adoc
+++ b/doc/iterations/2025.2/add_support_for_artifact_publication_from_a_project.adoc
@@ -1,0 +1,83 @@
+= Add support for artifact publication from a project
+
+== Problem
+
+Sirius Web doesn't provide an **in-application** way to publish artifacts from a project.
+The current import/export mechanisms rely on generating an interchange format file (Json, XMI, SysML), and uploading such file in a project to import it.
+
+When exported models are used as shared reusable components, it is the responsibility of the stakeholders to:
+
+- Store the shared models in an external location
+- Ensure that every user is working on the same version of the shared model
+- Coordinate to export a new version when needed, and import it in all the dependent projects
+
+These processes are error-prone, and could be more controlled if they all happened inside the application.
+
+== Key Result
+
+Sirius Web should provide a way for downstream application to 
+
+- Define their publication _workflow_
+- Store published artifacts in a provided data structure
+
+Note that the actual implementation of the publication workflow (i.e. what to publish from a given project) should be delegated to downstream applications. 
+We already have downstream projects that implement or want to implement different publication mechanisms:
+
+- Publish fragments of a model
+- Publish entire editing contexts
+- Publish multiple artifacts from a single project
+
+This shape doesn't address the question of importing published artifacts in a project. 
+A solution to this issue will be proposed in a complementary shape.
+
+=== Acceptance criteria
+
+- The support for studio should provide a workflow publishing all the domains and representations descriptions from a studio project.
+Each domain and each representation description should be published as a dedicated artifact.
+After the execution of the workflow of a studio, all the domains and representation descriptions should be visible in the artifact page.
+- Backend tests should be used to validate that this publication mecanism is working as expected
+
+
+== Solution
+
+Sirius Web should provide a menu to view the various workflows and run one.
+A new button will be availble in the navigation bar to open a dialog containing all the workflows of the project.
+The user will be able to click on one workflow to run it.
+For that a GraphQL query will be available to retrieve all the workflows of the project.
+
+Sirius Web should also provide a GraphQL mutation to trigger the execution of a project workflow. 
+This mutation should ensure that the execution of the workflow does not block the editing context event processor (some workflows may take a long time to complete).
+
+Note that this workflow support is not restricted to artifact publication.
+They could be leveraged by downstream applications to perform any kind of behavior on a project.
+
+Downstream applications should be able to trigger the execution of a workflow from their own user interface (from the contextual menu in the explorer, from a custom tool in the diagram palette, etc).
+
+
+Sirius Web should provide some database table(s) to store published artifacts. 
+Those tables should at least contain the following pieces of information for each published artifact:
+
+- identifier
+- serialized content
+- dependencies
+- version
+- description
+- date of creation
+- date of last modification
+
+We should keep in mind that one artifact may be composed of multiple documents and that multiple artifacts could be published from one document.
+Some artifacts will also have a dependency to other artifacts, we should be able to capture this information provided by the specifier from day one.
+Documents cannot be considered as the compulsory granularity used by an artifact.
+
+Finally, Sirius Web should provide a new page to let user see the artifacts that have been published by a project. 
+This page will be accessible at `projects/projectId/artifacts`, and it will contain a table with the information contained in the database, filtered for the selected project.
+
+=== Cutting backs
+
+- (Nice to have) downstream application can add custom properties on the published artifacts.
+
+== Rabbit holes
+
+Nothing identified.
+
+== No-gos


doc/iterations/2025.1/add_support_for_model_publication.adoc
```diff
@@ -0,0 +1,80 @@
...
+- Coordinate to export a new version when needed, and import it in all the dependent projects
+
+These processes are error-prone, and could be more controlled if they all happened inside the application.
+
+== Key Result
+
+Sirius Web users should be able to:
+
+- Publish a model with a given version number
+- Import a published with a given version number
```
AxelRICHARD: a published with => a published model with

doc/iterations/2025.1/add_support_for_model_publication.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+== Solution
+
+The contextual menu of _Model_ elements in the explorer contains a new _Publish_ tool.
```
sbegaudeau: How could you be sure that a single model is self contained and does not have references to other models in the same editing context? What would happen if a model A has references to objects in a model B? It cannot be acceptable to publish a broken model with proxies. I don't see how we could realistically support the publication of anything other than the entire editing context in Sirius Web.

As a result, this menu should probably be on the project itself. If we go one step further, I don't see how we could consider this feature as "model publication" instead of "project dependencies"

doc/iterations/2025.1/add_support_for_model_publication.adoc
```diff
@@ -0,0 +1,80 @@
...
+This tool opens a pop-up that asks the user for a version number and comments.
+
+**Note**: Sirius Web will not allow to publish a model with the a version number that matches an already published version of the model. 
+User will have to update the version of the model if they want to re-publish it.
```
sbegaudeau: So users should be able to view the list of versions already published in a dedicated page and (probably later) open them in read only to view what has been published.

doc/iterations/2025.1/add_support_for_model_publication.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+The _Upload model_ tool in the explorer is renamed to _Import model_, and the associated popup contains 2 tabs:
+- The _Upload_ tab that performs the same action as it does now (upload an external file as a model)
+- The _Published Models_ tab presents a list of published models (with their version number) the user can select to import in their project.
```
sbegaudeau: Keep both features separated because they really don't have the same behavior or lifecycle (and it's less expensive) and they will grow further apart.

doc/iterations/2025.1/add_support_for_model_publication.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+=== Scenario
+
+1. User publishes a model
```
sbegaudeau: The user should see what has been published in the past too and probably the list of projects that are using each version that has been published.

doc/iterations/2025.1/add_support_for_model_publication.adoc
```diff
@@ -0,0 +1,80 @@
...
+- The user fills requested information in the popup (version number, comment, etc) and presses _Publish_
+- A message is displayed indicating that the publication was successful
+
+2. User imports a published model
+- The user creates a new project
+- The user clicks on the _Import Model_ tool in the explorer toolbar
+- The user selects the _Published Models_ tab in the popup
+- The published models are displayed in the tab, with their version number
+- The user selects a published model and presses _Import_
+- The model is imported at the root of the project
```
sbegaudeau: What does imported mean?

- Copied and free to be modified? 
- Referenced and immutable?

We will need to let the end users should decide and only keep a link to the project using what has been published only in case of a reference.

doc/iterations/2025.1/add_support_for_model_publication.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+== Rabbit holes
+
+Nothing identified.
+
+== No-gos
+
+This shape doesn't address the issue of upgrading/downgrading and imported model with a new version.
+In the proposed solution, importing two different versions of the same model will create two models in the project (one for each version). 
+This behavior matches the current behavior of Sirius Web when importing the same external file two times.
```
sbegaudeau: And I don't think that this is what we want for most use cases. If I create a library, my first use case would be to let people include this library in their project without letting them manipulate it.

doc/iterations/2025.1/add_support_for_model_publication.adoc
```diff
@@ -0,0 +1,80 @@
+= Add support for model publication
+
+== Problem
```
sbegaudeau: We definitely need to talk about studios in this PRs. This feature was THE next step in the management of studio in Sirius Web. We worked for weeks on this topic before deactivating everything and putting the code aside. We will need to be able to publish studio projects and let end users use only the published view/domain instead of all the view/domain found in the database.

doc/iterations/2025.1/add_support_for_model_publication.adoc
```diff
@@ -0,0 +1,80 @@
...
+=== Breadboarding
+
+Publish model action in the _Model_ contextual menu and _Publish_ popup.
+
+image::images/publish-model.png[]
+
+Import model popup, displayed when clicking on the _Import Model_ tool in the explorer toolbar. 
+The _Upload_ tab corresponds to the current _Upload_ popup in Sirius Web.
+
+image::images/import-model.png[]
```
sbegaudeau: We can't realistically display all versions of everything that has been published on the server. We should probably list projects first and versions after that. This interface will have tens of thousands of entries if not more. A table with pagination is the bare minimum.

sbegaudeau (changes requested):

doc/iterations/2025.1/add_support_for_model_publication.adoc
sbegaudeau: If we want to publish parts of a project or even better if we want to publish multiple artifacts for a project, then I'm not sure that Sirius Web can decide how to do it. Think about how we are exporting X maven artifacts (that may be preprocessed by the way) from one repository. It then creates a dependency tree that is used by the one using the dependency.

The only thing I think we can do is to export all models in a project because that's the only coherent unit of semantic data that we have.

sbegaudeau: After having thought about this a bit more, given the fact that we have multiple downstream projects (at least 3 I think) that have implemented something along those lines, I suspect that we can't decide on our own what publishing means (for example a validation step before the publication, some processing to transform the data, etc).

I suspect that we should be able to register "workflows" (think Github Workflows for Github Actions) which let specifiers define how to publish things for a project.

```
public interface IProjectWorkflow {
    /*
     * Can us the nature of the project to figure out if the workflow is relevant
     */
    boolean canHandle(UUID projectId);

     /*
      * I'm not providing the editing context here on purpose to let the specifier load the snapshot of the
      * data from the database in order NOT to block the editing context event processor.
      */
    void handle(UUID projectId, ...);
}
```

A project could have multiple workflows (like we do in our Github repository) and the end user could just select the workflow to run for the publication. We have to keep in mind that a workflow could take some time (a few milliseconds to a couple of hours). The workflow would explicitly decide to perform a publication (it may not publish anything at all). One could make a workflow to publish everything, or another one which would for example only publish the `papaya::Component` from a Papaya project and not other artifacts.

If you want to run a workflow using some variables as input (here a version, or a specific object) you may contribute a dedicated tool to execute the mutation triggering the workflow in your own application. Some downstream projects already have dedicated user interface accessible conditionally from specific objects in specific projects to start this. In such use case, all downstream projects would "just" share a common workflow execution system and publication area (think Github Packages).

sbegaudeau: So there are probably two shapes for this topic:

- How to publish artifacts
- How to consume artifacts (project dependencies, reference vs copy etc)

gdaniel: I see, if Sirius Web forces specifiers to publish entire editing contexts, which is, as you said, the only thing that makes sense in a vacuum, this may conflict with existing publication implementations. For example, I know some downstream application have a publication mechanism where only a part of a model is published, because the application ensures that this sub model does not depend on anything.

If we go for the workflow approach, should Sirius Web provide a specific table for that? I am not sure this is what you mean by publication area. If we have such table in Sirius Web we need to let downstream applications add information next to the published model (e.g. version number, date, etc). This could be their responsibility, but in this case they would have to do kind of a lot of work every time they want to re-use the publication mechanism of Sirius Web. Maybe we can provide some default publication data that can be easily reused?

I agree this could be split into two shapes.

doc/iterations/2025.1/add_support_for_model_publication.adoc
gdaniel: The initial mockup we did for this popup displays the current version number and the latest one. I agree this could be improved to let the user see the list of all the published versions, potentially with the date of publication.

In any case, with the workflow approach, I think this is not a Sirius Web question anymore: the implementation of the workflow and the UI components that let the user provide the necessary variables to handle the publication will be defined in downstream applications. This is still a question for Studio publication, but I discuss it on the corresponding comment.

doc/iterations/2025.1/add_support_for_model_publication.adoc
gdaniel: Ok, this is not a Sirius Web question anymore with the workflow approach. But I keep it in mind.

doc/iterations/2025.1/add_support_for_model_publication.adoc
gdaniel: By upgrading I meant importing a newer version of an imported model, and assume that Sirius Web will take care of replacing the older version with the newer one (including some semantic changes to make sure the models are still somewhat consistent).

I didn't want to address this issue in this shape, because the "somewhat consistent" part is tricky, and overriding an imported model with a newer/older version can be done in a different shape later.

doc/iterations/2025.1/add_support_for_model_publication.adoc
gdaniel: Based on the previous comments, I assume this would be an implementation of a workflow tailored to Studios. To me this whole feature looks like a different shape if we go for a solution where Sirius Web only provides the workflow execution mechanism.

doc/iterations/2025.1/add_support_for_model_publication.adoc
gdaniel: Right, maybe a paginated table with one row/model and a dropdown to select the version would be better. We'll rework the mockups for this feature.

doc/iterations/2025.1/add_support_for_model_publication.adoc
sbegaudeau: > If we go for the workflow approach, should Sirius Web provide a specific table for that? I am not sure this is what you mean by publication area.

I think that there be should somewhere (probably another page) where users could see what has been published by a project. It would just list the artifacts published and when they have been published. Using a simple "static" table in a dedicated page (something like `/projects/:projectId/artifacts` or something.

> If we have such table in Sirius Web we need to let downstream applications add information next to the published model (e.g. version number, date, etc). This could be their responsibility, but in this case they would have to do kind of a lot of work every time they want to re-use the publication mechanism of Sirius Web. Maybe we can provide some default publication data that can be easily reused?

I was thinking about that and for example in Papyrus Web (since I worked on this one a tiny bit), they have the following properties:

- version kind (development, release, major or custom)
- version (can be computed from version kind)
- date
- author
- comments
- copyright
- message

Some fields are computed from others or from the backend for example they can check the latest artifact published to figure out the previous version and offer to create a new upgraded version (using a new release version kind to automatically go from 1.2.3 to 1.3.0 for example).

I don't think that every downstream projects could use the same properties and even if all artifacts on an instance of Sirius Web could use the same properties. They would still be free to provide a custom user interface to publish models but we could provide a way for specifiers to add some properties and tell us how to display them.

Something like this:

```
package org.eclipse.sirius.web.domain.boundedcontexts.publication;

@Table("artifact")
public class Artifact extends AbstractValidatingAggregateRoot<Artifact> implements Persistable<UUID> {
    @Transient
    private boolean isNew;

    @Id
    private UUID id;

    private String version;

    private String description;

    @MappedCollection(idColumn = "artifact_id", keyColumn = "index")
    private List<ArtifactProperty> properties;

    private Instant createdOn;

    private Instant lastModifiedOn;
}
```

with something like this:

```
@Table("artifact_property")
public record ArtifactProperty(
    String name,
    String value
) {}
```

This way, if a downstream application wants to provide a custom user interface or just a custom workflow which would capture additional properties to be attached to the artifact published, they could tell us. On the frontend, we would only, for now display the properties in additional columns of the table. The field version and description would be mandatory (the description could be blank but at least it would be handled natively by Sirius Web) and additional fields could be contributed if a downstream project wants it. This part regarding custom properties of the artifact published would definitely be a cutting back that could be implemented later.

doc/iterations/2025.1/add_support_for_model_publication.adoc
sbegaudeau: Yes I agree, I still think Sirius Web should have a page listing artifacts published but a specific downstream application will propose its end users with the integration of those artifacts using a specific user interface.

gdaniel:
@sbegaudeau I reworked the shape to integrate what we discussed in the comments. The shape now focuses on artifact publication, importing published artifacts will be in a different shape

sbegaudeau:
> @sbegaudeau I reworked the shape to integrate what we discussed in the comments. The shape now focuses on artifact publication, importing published artifacts will be in a different shape

I'll have another look at the shape today in order to get it merged this week

doc/iterations/2025.1/add_support_for_model_publication.adoc
gdaniel: Fixed (this part is not in the current version of the shape).

doc/iterations/2025.1/add_support_for_artifact_publication_from_a_project.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+When exported models are used as shared reusable components, it is the responsibility of the stakeholders to:
+
+- Store the shared models in an external location
+- Ensure that every user is working on the same version of the shared model
+- Coordinate to export a new version when needed, and import it in all the dependent projects
+
+These processes are error-prone, and could be more controlled if they all happened inside the application.
+
+== Key Result
```
sbegaudeau: I'll just add an acceptance criteria section which is now required for all new shapes.

doc/iterations/2025.1/add_support_for_artifact_publication_from_a_project.adoc
```diff
@@ -0,0 +1,82 @@
...
+- Publish multiple artifacts from a single project
+
+This shape doesn't address the question of importing published artifacts in a project. 
+A solution to this issue will be proposed in a complementary shape.
+
+== Solution
+
+Sirius Web should provide a project workflow interface similar to the example below
+```java
+public interface IProjectWorkflow {
```
sbegaudeau: Given the fact that we will want to let end users select the workflow to run, we should probably adopt the same pattern as project templates with both a provider and a handler. These technical parts should be better in a dedicated ADR. I'll initialize one with some additional details to keep the shape light.

Don't hesitate to improve the ADR thanks to another PR later.

doc/iterations/2025.1/add_support_for_artifact_publication_from_a_project.adoc
```diff
@@ -0,0 +1,82 @@
...
+}
+```
+
+Sirius Web should also provide a GraphQL mutation to trigger a project workflow. 
+This mutation should ensure that the execution of the workflow does not block the editing context event processor (some workflows may take a long time to complete).
+
+Note that this interface (and the GraphQL mutation) are not restricted to artifact publication. 
+They could be leveraged by downstream applications to define any kind of workflow on a project.
+
+Downstream applications contributing implementations of this interface could then trigger the workflow from their own tools (from the contextual menu in the explorer, from a tool in the diagram palette, etc).
```
sbegaudeau: Yes downstream application should be able to provide a custom UI to trigger a workflow but I've been asked to ensure that every downstream projects will not have to re-invent the same wheel, so we will have to provide a common user interface for everybody. It may have to be customizable in the future but a first version will be necessary. I'll add some details about that.

Anyway, we can't have an unreachable feature like that in Sirius Web so we will have to provide something at least for studios (and maybe learn from this to build something even more generic).

doc/iterations/2025.1/add_support_for_artifact_publication_from_a_project.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+Sirius Web should provide a database table to store published artifacts. 
+This table should contain the following information for each published artifact:
```
sbegaudeau: I think that it will require more than that, we will have to discover it when we will implement it but some artifacts will depend on others and we should not lose this piece of information. We will have to see how it behaves when we will try to publish view and domain models.

doc/iterations/2025.1/add_support_for_artifact_publication_from_a_project.adoc
```diff
@@ -0,0 +1,82 @@
...
+
+- (Nice to have) downstream application can add custom properties on the published artifacts.
+
+== Rabbit holes
+
+Nothing identified.
+
+== No-gos
+
+The publication of Sirius Web artifacts (e.g. studios) is out of the scope of this shape. The proposed mechanism could be leveraged in the future to implement these new features.
```
sbegaudeau: I'm not sure what was said on this topic before but I will disagree on this part because I don't see what other kind of example you would propose to test that this feature will work. You cannot work blindly on this feature without ever running the code.

I need an example and its either Studio or Papaya and given the fact that there's not really anything relevant in Papaya for this feature (even if I can create something to make it relevant) I don't see how this feature could end up as anything other than a large piece of dead code without the support for uploading studios.

A real example is also needed in order to start facing real questions from day one.

sbegaudeau (commented):
I'll update the shape and merge it

sbegaudeau (approved):

PR #4289, opened 2024-12-12
[fix] Fix palette size on Firefox

mcharfadi:
image


pcdavid (approved):
Seems OK, but it'd be nice if we found a proper fix to get the smaller palette on all browsers, so maybe we should wait a little and investigate further.

PR #4344, opened 2024-12-23
[4343] Keep filters when opening a table

### LINKED ISSUE #4311: Typos in Table variables
* [x] **I have checked that this bug has not yet been reported by someone else**
* [x] **I have checked that this bug appears on Chrome**
* [x] **I have specified the version** :
- latest
* [x] **I have specified my environment** : 
- All

There are two typos in Table variables defined in `org.eclipse.sirius.components.tables.descriptions.ColumnDescription`:

```java
    public static final String COLUMN_TARGET_OBJECT = "columTargetObject";

    public static final String COLUMN_TARGET_OBJECT_ID = "columTargetObjectId";
```

should be:

```java
    public static final String COLUMN_TARGET_OBJECT = "columnTargetObject";

    public static final String COLUMN_TARGET_OBJECT_ID = "columnTargetObjectId";
```

### LINKED ISSUE #4343: Table global filter and column filters are reset at the opening
When a filter is set, the value is saved in the database, but the next time the table is opened, the filter returns to empty.

frouene:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/4343

# Pull request template

## General purpose
- [x] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/tables/backend/sirius-components-collaborative-tables/src/main/java/org/eclipse/sirius/components/collaborative/tables/TableEventProcessorFactory.java
+++ b/packages/tables/backend/sirius-components-collaborative-tables/src/main/java/org/eclipse/sirius/components/collaborative/tables/TableEventProcessorFactory.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024 CEA LIST.
+ * Copyright (c) 2024, 2025 CEA LIST.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
@@ -49,7 +49,7 @@ public class TableEventProcessorFactory implements IRepresentationEventProcessor
     private static final String SIZE = "size";
     private static final String GLOBAL_FILTER = "globalFilter";
     private static final String COLUMN_FILTERS = "columnFilters";
-    
+
     private final IRepresentationSearchService representationSearchService;
 
     private final IRepresentationDescriptionSearchService representationDescriptionSearchService;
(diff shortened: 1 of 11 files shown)


packages/tables/backend/sirius-components-collaborative-tables/src/main/java/org/eclipse/sirius/components/collaborative/tables/TableEventProcessorFactory.java
```diff
@@ -150,7 +150,7 @@ private CursorBasedPaginationData getCursorBasedPaginationData(IEditingContext e
     }
 
     private String getGlobalFilter(String representationId) {
-        String globalFilter = "";
+        String globalFilter = null;
```
sbegaudeau: I don't understand why `TableCreationParameters.Builder#globalFilter` or `TableCreationParameters.Builder#columnFilters` has to be nullable nor why it has to be nullable in `org.eclipse.sirius.components.tables.components.TableComponentProps`either and everywhere else for that matter and why you are trying to propagate even more `null` values everywhere.

You are doing this and deep down in `TableComponent`, you now have to do:

```
        var globalFilter = this.props.globalFilter();
        if (this.props.globalFilter() == null) {
            globalFilter = optionalPreviousTable.map(Table::getGlobalFilter).orElse("");
            variableManager.put(TableRenderer.GLOBAL_FILTER_DATA, globalFilter);
        }

        List<ColumnFilter> columnsFilters;
        if (this.props.columnFilters() == null) {
            columnsFilters = optionalPreviousTable.map(Table::getColumnFilters).orElse(List.of());
        } else {
            columnsFilters = this.props.columnFilters();
        }
```

Why aren't those values initialized properly in `TableEventProcessorFactory` given that you already have access to the previous version of the table? I don't see why more `null` solves anything given that you have to fallback to a proper value in the frontend

```
  const [globalFilter, setGlobalFilter] = useState<string>(table.globalFilter ?? '');

  useEffect(() => {
    if (enableGlobalFilter) {
      changeGlobalFilterValue(globalFilter ?? '');
      onGlobalFilterChange(globalFilter ?? '');
    }
  }, [globalFilter]);

  useEffect(() => {
    setGlobalFilter(table.globalFilter ?? '');
  }, [table.globalFilter]);
```

What state of the table is `null` supposed to represent?

sbegaudeau (changes requested):

packages/tables/backend/sirius-components-collaborative-tables/src/main/java/org/eclipse/sirius/components/collaborative/tables/TableEventProcessorFactory.java
frouene: When we open an existing table, we are unaware of the persisted value of the global filter on the front end (the same applies to the column filter). However, we do not want to send an empty value to the backend, as this would override the previous value (similar to when a user clicks the clear button in the filter section). Therefore, we need to recognize that we are in a situation where the table's global filter is not yet known, which is why I used the null value.

frouene: When we open an existing table, we are unaware of the persisted value of the global filter on the front end (the same applies to the column filter). However, we do not want to send an empty value to the backend, as this would override the previous value (similar to when a user clicks the clear button in the filter section). Therefore, we need to recognize that we are in a situation where the table's global filter is not yet known, which is why I used the null value.

sbegaudeau: I understand why the frontend can send a lack of value to the back in the URL of the subscription sure but here we are opening the representation and we have the table from the database so we both know that:

- The frontend can't really send a meaningful value
- The source of truth is what has been persisted in the backend the last time

So why give `null` to everybody if the real answer is just next to us? Am I mistaken in thinking that the only possible value here is the real value coming the previous table loaded line 88?

frouene: Yes, you're right, I didn't understand your comment at first. I can indeed directly retrieve the persisted value without needing to handle the null case.
I've updated the PR to remove all nulls for globalSearch and columnFilter

sbegaudeau (approved):
I'll merge it after the build

PR #4448, opened 2025-01-21
[4362] Add new navigationBarCenterContributionExtensionPoint

### LINKED ISSUE #4362: Add frontend extension point to wrap/augment the NavigationBar children
The `NavigationBar` component already supports several extension points.

`navigationBarLeftContributionExtensionPoint` (resp. `navigationBarRightContributionExtensionPoint`), but elements added to this are "packed" to the left side (resp. right side) of the header.

`editProjectNavbarSubtitleExtensionPoint` (used only in `EditProjectNavbar`) adds support for a subtitle displayed _below_ the project name:

Capture d’écran du 2025-01-02 18-05-26

There is no way to add elements around the project name at the center, to replace it or add some context (for example to be used in custom `ProjectContextMenuItem` elements).

It should be possible to support all of these use cases, and more, by adding a single extension point around the `<div>{children}</div>` part of `NavigationBar`.

pcdavid:
With something like:

```js
defaultExtensionRegistry.addComponent(navigationBarCenterContributionExtensionPoint, {
  identifier: `siriusWeb_${navigationBarCenterContributionExtensionPoint.identifier}_test`,
  Component: ({ children }: NavigationBarProps) => {
    if (children) {
      return (
        <>
          <Typography>LEFT</Typography>
          <div>{children}</div>
          <Typography>RIGHT</Typography>
        </>
      );
    } else {
      return null;
    }
  },
});
```

it renders:

image


packages/sirius-web/frontend/sirius-web-application/src/navigationBar/NavigationBar.tsx
```diff
@@ -38,19 +39,22 @@ const useNavigationBarStyles = makeStyles()((theme) => ({
   },
   toolbar: {
     display: 'grid',
-    gridTemplateColumns: '1fr min-content 1fr',
+    gridTemplateColumns: '1fr 1fr 1fr',
```
sbegaudeau: Why change that? It creates a massive space for the center area

packages/sirius-web/frontend/sirius-web-application/src/navigationBar/NavigationBar.tsx
```diff
@@ -38,19 +39,22 @@ const useNavigationBarStyles = makeStyles()((theme) => ({
...
   },
   left: {
     display: 'flex',
     flexDirection: 'row',
     alignItems: 'center',
+    justifyContent: 'flex-start',
     gap: theme.spacing(1),
   },
-  link: {
+  center: {
```
sbegaudeau: Why choose to align things this way, let consumers of the extension point do what they want. The may not want to be aligned vertically, they may not want any gap etc.

sbegaudeau (changes requested):

packages/sirius-web/frontend/sirius-web-application/src/navigationBar/NavigationBar.tsx
pcdavid: Done.

packages/sirius-web/frontend/sirius-web-application/src/navigationBar/NavigationBar.tsx
pcdavid: Done.

sbegaudeau (approved):

sbegaudeau (approved):

PR #4519, opened 2025-02-07
[doc] Add an ADR to support project dependencies

gdaniel:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -27,6 +27,7 @@
 - [ADR-181] Improve the customization of the navigation bar
 - [ADR-182] Add additional capabilities to synchronize some code with the deletion of a project
 - [ADR-183] Add support of row hierarchy in tables
+- [ADR-184] Add support for project dependencies
 
 === Deprecation warning
 
--- a/doc/adrs/184_add_support_for_project_dependencies.adoc
+++ b/doc/adrs/184_add_support_for_project_dependencies.adoc
@@ -1,0 +1,29 @@
+= [ADR-184] Add support for project dependencies
+
+
+== Context
+
+https://github.com/eclipse-sirius/sirius-web/blob/master/doc/iterations/2025.2/add_support_for_the_import_of_libraries.adoc[This shape] describes how project can import libraries, but does not define how the project-to-library dependency is stored.
+
+We need to capture this information to implement future library import/upgrade strategies.
+For example, we need to know whether a part of a project comes from a library (to prevent users from editing it), or what are the libraries imported by a project that can be upgraded to a newer version.
+
+
+== Decision
+
+A new table `semantic_data_dependency` will be introduced with the following columns:
+
+- `semantic_data_id`
+- `library_id`
+
+An `IEditingContextProcessor` implementation should be provided to load the content of imported libraries.
+This solution reuses the approach described in https://github.com/eclipse-sirius/sirius-web/blob/master/doc/adrs/124_add_support_for_libraries.adoc[ADR 124].
+The loaded resources should have a dedicated URI scheme: `sirius:/library/`, and should not be persisted with the editing context of the project that imports them.
+
+A new `IReadOnlyObjectPredicate` should be created, as well as a `ComposedReadOnlyObjectPredicate` to let downstream application implement their own read-only policy.
+Sirius Web should provide a default implementation that returns `true` if the object is contained in a resource with the library scheme.
+
+
+== Status
+
+Work in progress


doc/adrs/183_add_support_for_project_dependencies.adoc
```diff
@@ -0,0 +1,40 @@
...
+We need to capture this information to implement future library import/upgrade strategies.
+For example: we need to know whether a part of a project comes from a read-only library (to prevent users from editing it), or what are the libraries imported by a project that can be upgraded to a newer version.
+
+== Decision
+
+A new table `project_dependency` with the following columns should be created
+
+- `projectId`
+- `libraryId`
+- `readOnly`: a boolean indicating whether the imported library is read-only or not
```
sbegaudeau: Why would that be necessary? If it's not readonly then it should be part of the semantic data and thus modifiable. How could we have a dependency (so something that is not under our responsibility) that we can modify? Either something is in our semantic data (we can modify it, it's ours) or it's in a library (we can't modify it)

doc/adrs/183_add_support_for_project_dependencies.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+An `IEditingContextProcessor` implementation should be provided to copy imported read-only libraries into studio projects when loading them.
+This solution reuses the approach described in https://github.com/eclipse-sirius/sirius-web/blob/master/doc/adrs/124_add_support_for_libraries.adoc[ADR 124].
+Nothing is done to handle imported non read-only libraries into studio projects: the default behavior is to assume they are already in the editing context (e.g. they have been compied into it on import).
+
+Downstream applications need to implement their own `IEditingContextProcessor` to handle libraries they publish.
+
+The `StudioEditingContextPersistenceFilter` should prevent the persistence of imported read-only libraries in studio project.
```
sbegaudeau: The imported library should probably have a dedicated schema to not be persisted at all. Something like `sirius:/library/` or something. It should not be done for one specific use case but instead be a feature.

doc/adrs/183_add_support_for_project_dependencies.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+- `projectId`
+- `libraryId`
+- `readOnly`: a boolean indicating whether the imported library is read-only or not
+
+An `IEditingContextProcessor` implementation should be provided to copy imported read-only libraries into studio projects when loading them.
+This solution reuses the approach described in https://github.com/eclipse-sirius/sirius-web/blob/master/doc/adrs/124_add_support_for_libraries.adoc[ADR 124].
+Nothing is done to handle imported non read-only libraries into studio projects: the default behavior is to assume they are already in the editing context (e.g. they have been compied into it on import).
+
+Downstream applications need to implement their own `IEditingContextProcessor` to handle libraries they publish.
```
sbegaudeau: Why? Why can't we load published libraries on our own?

sbegaudeau (changes requested):

doc/adrs/183_add_support_for_project_dependencies.adoc
gdaniel: Because we don't know how to do that.
A library can be a resource or a set of resources (in this case we would know how to load it), but it can be something else  (individual elements, textual serialization of a model, etc).
We can have a dedicated library loading mechanism if you want, and not use the `IEditingContextProcessor`, but in any case the actual implementation that changes the content of the editing context needs to be in downstream applications.
For example, if we decide in SysON that the content of published libraries is stored as textual SysML, we need to define in SysON a way to load them, I don't see how this could be done in Sirius Web. 
That's my rationale to propose an implementation only for studios: because we know what libraries they publish and how to load them.

Btw, this makes me think that we should probably add something like a `nature` column in the library table to ease the implementation of library loaders. We could even have a default implementation that loads libraries that contain Sirius Web resources, that would be more powerful than the studio solution, while still extensible to support other loading mechanisms.

doc/adrs/183_add_support_for_project_dependencies.adoc
gdaniel: Agreed, I've added it

doc/adrs/183_add_support_for_project_dependencies.adoc
gdaniel: You are right, the idea was to have a way to track libraries that were copied in the project, but as you said, these libraries are in our semantic data, and they are now our responsibility, so there is no reason to track this information for them. I have removed the boolean

doc/adrs/183_add_support_for_project_dependencies.adoc
```diff
@@ -0,0 +1,40 @@
...
+== Context
+
+https://github.com/eclipse-sirius/sirius-web/blob/master/doc/iterations/2025.2/add_support_for_the_import_of_libraries.adoc[This shape] describes how project can import libraries, but does not define how the project-to-library dependency is stored.
+
+We need to capture this information to implement future library import/upgrade strategies.
+For example: we need to know whether a part of a project comes from a library (to prevent users from editing it), or what are the libraries imported by a project that can be upgraded to a newer version.
+
+== Decision
+
+A new table `project_dependency` with the following columns should be created
```
sbegaudeau: Since the coupling between project and semantic data has evolved in the past few weeks, this link will be necessary between semantic data and library

doc/adrs/183_add_support_for_project_dependencies.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+- `projectId`
+- `libraryId`
+
+The `library` table should be updated with a `nature` column to give hints on the nature of the content stored in the library.
```
sbegaudeau: The library has semantic data which have the list of domains used associated. There's no need for library natures to know how to load them for now.

doc/adrs/183_add_support_for_project_dependencies.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+An `IEditingContextProcessor` implementation should be provided to load the content of imported libraries with the `Sirius` nature as Sirius Web resources in the editing context.
+This solution reuses the approach described in https://github.com/eclipse-sirius/sirius-web/blob/master/doc/adrs/124_add_support_for_libraries.adoc[ADR 124].
+The loaded resources should have a dedicated URI schema: `sirius:/library/`, and should not be persisted with the editing context of the project that imports them.
+
+Downstream applications need to implement their own `IEditingContextProcessor` to handle libraries with a different nature.
+
+The `IObjectService` should have a method to check whether an object is read-only.
```
sbegaudeau: No need to modify the object service, we will probably move to its deletion than to couple it with additional concepts. Just keep the other one.

sbegaudeau (changes requested):
I'll update the ADR and merge it

sbegaudeau (approved):

PR #4574, opened 2025-02-17
[4573] Adapt the palette size to its content

mcharfadi:


mcharfadi:
This change does not work on Firefox because of the use of gridAutoRows: `fit-content(100%)`, in `Palette.tsx`

fit-content() is not supported by Firefox browser yet.
https://developer.mozilla.org/en-US/docs/Web/CSS/max-height

image

mcharfadi:
This has been changed in another PR https://github.com/eclipse-sirius/sirius-web/pull/5221

PR #4598, opened 2025-02-20
[4597] Add a command to publish libraries from studios

gdaniel:
# Pull request template

Based on #4587 

## General purpose
What is the main goal of this pull request?
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management
- [ ] Has the pull request been added to the relevant project and milestone? (Only if you know that your work is part of a specific iteration such as the current one)
- [ ] Have the `priority:` and `pr:` labels been added to the pull request? (In case of doubt, start with the labels `priority: low` and `pr: to review later`)
- [ ] Have the relevant issues been added to the pull request?
- [ ] Have the relevant labels been added to the issues? (`area:`, `difficulty:`, `type:`)
- [ ] Have the relevant issues been added to the same project and milestone as the pull request?
- [ ] Has the `CHANGELOG.adoc` been updated to reference the relevant issues?
- [ ] Have the relevant API breaks been described in the `CHANGELOG.adoc`? (Including changes in the GraphQL API)
- [ ] In case of a change with a visual impact, are there any screenshots in the `CHANGELOG.adoc`? For example in `doc/screenshots/2022.5.0-my-new-feature.png`

## Architectural decision records (ADR)
- [ ] Does the title of the commit contributing the ADR start with `[doc]`?
- [ ] Are the ADRs mentioned in the relevant section of the  `CHANGELOG.adoc`?

## Dependencies
- [ ] Are the new / upgraded dependencies mentioned in the relevant section of the `CHANGELOG.adoc`?
- [ ] Are the new dependencies justified in the `CHANGELOG.adoc`?

## Frontend

This section is not relevant if your contribution does not come with changes to the frontend.

### General purpose
- [ ] Is the code properly tested? (Plain old JavaScript tests for business code and tests based on React Testing Library for the components)

### Typing
We need to improve the typing of our code, as such, we require every contribution to come with proper TypeScript typing for both changes contributing new files and those modifying existing files.
Please ensure that the following statements are true for each file created or modified (this may require you to improve code outside of your contribution).

- [ ] Variables have a proper type
- [ ] Functions’ arguments have a proper type
- [ ] Functions’ return type are specified
- Hooks are properly typed:
	- [ ] `useMutation<DATA_TYPE, VARIABLE_TYPE>(…)`
	- [ ] `useQuery<DATA_TYPE, VARIABLE_TYPE>(…)`
[...]

PR #4653, opened 2025-02-28
[4652] Add support for loading editing contexts with dependencies

### LINKED ISSUE #4652: Add support for project to library dependencies

### Expected behavior

Details of this ticket are available in [ADR 184](https://github.com/eclipse-sirius/sirius-web/blob/master/doc/adrs/184_add_support_for_project_dependencies.adoc).

The main concepts are:
- Allow to add dependencies to a project (e.g. libraries this project needs)
- Load the dependencies of a project when loading its editing context
- Ensure that loaded dependencies are read-only

### LINKED ISSUE #4668: Identify the content of semantic data dependencies as read only

gdaniel:
# Pull request template

Based on https://github.com/eclipse-sirius/sirius-web/pull/4598

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextDependencyLoader.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextDependencyLoader.java
@@ -1,0 +1,74 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.web.application.editingcontext.services;
+
+import java.util.List;
+import java.util.Objects;
+
+import org.eclipse.sirius.components.core.api.IEditingContext;
+import org.eclipse.sirius.components.emf.services.api.IEMFEditingContext;
+import org.eclipse.sirius.web.application.UUIDParser;
+import org.eclipse.sirius.web.application.editingcontext.services.api.IEditingContextDependencyLoader;
+import org.eclipse.sirius.web.application.editingcontext.services.api.IEditingContextMigrationParticipantPredicate;
+import org.eclipse.sirius.web.application.editingcontext.services.api.IResourceLoader;
+import org.eclipse.sirius.web.application.library.services.LibraryMetadataAdapter;
+import org.eclipse.sirius.web.domain.boundedcontexts.library.services.api.ILibrarySearchService;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.SemanticData;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.services.api.ISemanticDataSearchService;
+import org.springframework.data.jdbc.core.mapping.AggregateReference;
+import org.springframework.stereotype.Service;
+
+/**
+ * Loads dependencies into the editing context.
+ *
+ * @author gdaniel
+ */
+@Service
+public class EditingContextDependencyLoader implements IEditingContextDependencyLoader {
+
+    private final ISemanticDataSearchService semanticDataSearchService;
+
+    private final IResourceLoader resourceLoader;
+
+    private final List<IEditingContextMigrationParticipantPredicate> migrationParticipantPredicates;
+
+    private final ILibrarySearchService librarySearchService;
+
+    public EditingContextDependencyLoader(ISemanticDataSearchService semanticDataSearchService, IResourceLoader resourceLoader, List<IEditingContextMigrationParticipantPredicate> migrationParticipantPredicates, ILibrarySearchService librarySearchService) {
+        this.semanticDataSearchService = Objects.requireNonNull(semanticDataSearchService);
+        this.resourceLoader = Objects.requireNonNull(resourceLoader);
+        this.migrationParticipantPredicates = Objects.requireNonNull(migrationParticipantPredicates);
+        this.librarySearchService = Objects.requireNonNull(librarySearchService);
+    }
+
+    @Override
+    public void loadDependencies(IEditingContext editingContext) {
+        if (editingContext instanceof IEMFEditingContext emfEditingContext) {
+            var dependenciesSemanticData = new UUIDParser().parse(emfEditingContext.getId())
+                    .map(this.semanticDataSearchService::findAllDependenciesRecursivelyById)
+                    .orElse(List.of());
+            for (SemanticData semanticData : dependenciesSemanticData) {
+                var optionalLibrary = this.librarySearchService.findBySemanticData(AggregateReference.to(semanticData.getId()));
+                semanticData.getDocuments().forEach(document -> {
+                    var applyMigrationParticipants = this.migrationParticipantPredicates.stream().anyMatch(predicate -> predicate.test(editingContext.getId()));
+                    this.resourceLoader.toResource(emfEditingContext.getDomain().getResourceSet(), document.getId().toString(), document.getName(), document.getContent(), applyMigrationParticipants).ifPresent(resource -> {
+                        if (optionalLibrary.isPresent()) {
+                            var library = optionalLibrary.get();
+                            resource.eAdapters().add(new LibraryMetadataAdapter(library.getNamespace(), library.getName(), library.getVersion()));
+                        }
+                    });
+                });
+            }
+        }
+    }
+}
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextLoader.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextLoader.java
@@ -21,12 +21,16 @@ import org.eclipse.sirius.components.core.api.IEditingContextRepresentationDescr
 import org.eclipse.sirius.components.emf.services.EditingContextCrossReferenceAdapter;
 import org.eclipse.sirius.emfjson.resource.JsonResource;
 import org.eclipse.sirius.web.application.editingcontext.EditingContext;
+import org.eclipse.sirius.web.application.editingcontext.services.api.IEditingContextDependencyLoader;
 import org.eclipse.sirius.web.application.editingcontext.services.api.IEditingContextLoader;
 import org.eclipse.sirius.web.application.editingcontext.services.api.IEditingContextMigrationParticipantPredicate;
 import org.eclipse.sirius.web.application.editingcontext.services.api.IResourceLoader;
+import org.eclipse.sirius.web.application.library.services.LibraryMetadataAdapter;
+import org.eclipse.sirius.web.domain.boundedcontexts.library.services.api.ILibrarySearchService;
 import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.SemanticData;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.data.jdbc.core.mapping.AggregateReference;
 import org.springframework.stereotype.Service;
 
 /**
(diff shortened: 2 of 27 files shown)


packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/EditingContextLibraryLoader.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public EditingContextLibraryLoader(ISemanticDataSearchService semanticDataSearchService, ILibrarySearchService librarySearchService, IResourceLoader resourceLoader, List<IEditingContextMigrationParticipantPredicate> migrationParticipantPredicates) {
+        this.semanticDataSearchService = Objects.requireNonNull(semanticDataSearchService);
+        this.librarySearchService = Objects.requireNonNull(librarySearchService);
+        this.resourceLoader = Objects.requireNonNull(resourceLoader);
+        this.migrationParticipantPredicates = Objects.requireNonNull(migrationParticipantPredicates);
+    }
+
+    @Override
+    public void loadLibraries(IEditingContext editingContext) {
```
gdaniel, line 60: If the provided `editingContext` is built from the semantic data of a library, this class won't load its dependencies correctly. There is a discussion on [this PR](https://github.com/eclipse-sirius/sirius-web/pull/4598) to know if we should merge the library dependencies and the semantic data dependencies. I'll update this PR based on the conclusion.

packages/sirius-web/backend/sirius-web-infrastructure/src/main/resources/db/changelog/2025.4/02-add-semantic-data-dependency-table.xml
```diff
@@ -0,0 +1,35 @@
...
+  ~
+  ~ Contributors:
+  ~     Obeo - initial API and implementation
+  -->
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.9.xsd">
+
+    <changeSet id="02-add-semantic-data-dependency-table" author="gdaniel">
```
gdaniel, line 19: As we discussed @sbegaudeau, I think this table could be used for both project-to-library dependencies, and library-to-library dependencies. This way we would have a single table to store dependencies to libraries.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/services/SemanticDataUpdateService.java
```diff
@@ -45,4 +52,38 @@ public void updateDocuments(ICause cause, AggregateReference<SemanticData, UUID>
             this.semanticDataRepository.save(semanticData);
         });
     }
+
+    @Override
+    public IResult<Void> addDependency(ICause cause, UUID semanticDataId, UUID dependencySemanticDataId) {
```
gdaniel: Since we have merged library-to-library dependencies and project-to-library dependencies into semantic data dependencies, we now need to provide update capabilities to be able to add dependencies to a project (e.g. import a library).
This is dangerous, because it also allows to update the dependencies of a library, although libraries should be immutable. I don't know where we should ensure this.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/SemanticData.java
```diff
@@ -151,6 +174,20 @@ public static Builder newSemanticData() {
         return new Builder();
     }
 
+    @Override
+    public boolean equals(Object obj) {
```
gdaniel: Having `equals` and `hashCode` in `SemanticData` eases my life when computing transitive dependencies. This could have some consequences I do not anticipate though. If this is the case, I can rework the code so that we don't need it.

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/libraries/LibraryControllerIntegrationTests.java
```diff
@@ -63,9 +63,6 @@ public class LibraryControllerIntegrationTests extends AbstractIntegrationTests
     @Autowired
     private ISemanticDataSearchService semanticDataSearchService;
 
-    @Autowired
-    private IStandardLibrarySemanticDataInitializer standardLibrarySemanticDataInitializer;
```
gdaniel, line 67: This is not used in the test, I removed it

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/services/EditingContextLifecycleTests.java
```diff
@@ -115,4 +138,34 @@ public void givenEditingContextProperlyLoadedWhenItIsModifiedAndPersistedThenThe
...
         }
     }
+
+    @Test
+    @GivenSiriusWebServer
+    @DisplayName("Given semantic data with a dependency, when the loading is performed, then the semantic data are available in the editing context")
+    public void givenSemanticDataWithDependencyWhenLoadingIsPerformedThenSemanticDataAvailableInEditingContext() {
+        Optional<Library> library = this.librarySearchService.findByNamespaceAndNameAndVersion("java", "Java Standard Library", "17.0.0");
+        assertThat(library).isPresent();
+        var result = this.semanticDataUpdateService.addDependency(null, PapayaIdentifiers.PAPAYA_EDITING_CONTEXT_ID, library.get().getSemanticData().getId());
```
gdaniel: I manually added the dependency to setup my test. This could be simplified by moving the dependency in the initialization script (this would imply to move the Java Standard Library in the script too, but since it hasn't been done in the previous PR I didn't do it here).

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextDependencyLoader.java
```diff
@@ -0,0 +1,73 @@
...
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.services.api.ISemanticDataSearchService;
+import org.springframework.stereotype.Service;
+
+/**
+ * Loads dependencies into the editing context.
+ *
+ * @author gdaniel
+ */
+@Service
+public class EditingContextDependencyLoader implements IEditingContextDependencyLoader {
```
gdaniel, line 37: This is an external service because some import commands (future PR) will need to reload the dependencies to show the new imported resources to the end user

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextLoader.java
```diff
@@ -41,22 +42,27 @@ public class EditingContextLoader implements IEditingContextLoader {
 
     private final IResourceLoader resourceLoader;
 
+    private final IEditingContextDependencyLoader editingContextDependencyLoader;
```
sbegaudeau, line 48: Try not to add APIs this way since they can have only one implementation. Most of the time it's better with multiple implementations so here a `List<IEditingContextDependencyLoader> editingContextDependencyLoaders`

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/api/IEditingContextDependencyLoader.java
```diff
@@ -0,0 +1,28 @@
...
+import org.eclipse.sirius.components.core.api.IEditingContext;
+
+/**
+ * Loads dependencies into the editing context.
+ *
+ * @author gdaniel
+ */
+public interface IEditingContextDependencyLoader {
+
+    String DEPENDENCY_SCHEME = "dependency";
```
sbegaudeau: We wanted to use something like `siriusweb:/dependency/` just like EMF uses `platform:/{plugin|resource}/` for example.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/services/SemanticDataSearchService.java
```diff
@@ -50,4 +54,29 @@ public Optional<SemanticData> findById(UUID id) {
...
+                .map(AggregateReference::getId)
+                .map(this::findById)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .forEach(dependencySemanticData -> {
+                    collectedSemanticData.add(dependencySemanticData);
+                    this.collectTransitiveSemanticData(dependencySemanticData, collectedSemanticData);
+                });
+        }
+    }
```
sbegaudeau: This could load a ton of data one request at a time from the database, It would be better to compute first the list of all the ids to load and then load them and all that in two requests instead of "n". I suspect that you would not need to add equals / hashcode in SemanticData this way.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/services/SemanticDataUpdateService.java
sbegaudeau: The application service should carry the business behavior indicating that a library cannot be updated. The semantic data bounded context does not really care about that. We will have to see this in some application service when the time will come.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/services/api/ISemanticDataSearchService.java
```diff
@@ -30,4 +31,6 @@ public interface ISemanticDataSearchService {
     Optional<SemanticData> findById(UUID id);
 
     boolean existsById(UUID id);
+
+    Set<SemanticData> findAllTransitiveSemanticDataById(UUID id);
```
sbegaudeau: `findAllDependenciesRecursivelyById` would be a bit more explicit

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/services/api/ISemanticDataUpdateService.java
```diff
@@ -27,4 +28,9 @@
  */
 public interface ISemanticDataUpdateService {
     void updateDocuments(ICause cause, AggregateReference<SemanticData, UUID> semanticData, Set<Document> documents, Set<String> domainUris);
+
+    IResult<Void> addDependency(ICause cause, UUID semanticDataId, UUID dependencySemanticDataId);
+
+    IResult<Void> removeDependency(ICause cause, UUID semanticDataID, UUID dependencySemanticDataId);
```
sbegaudeau: Just a side note but in order to support loading editing contexts with dependencies, you don't need to have either of those methods. You just need to build a better example for example in the SQL files or by making the Papaya example more complex with a dependency. Here you have two methods with two events that nobody needs right know.

You are adding stuff because you think we might need it not because we do need it (tests are not relevants for such addition, we can test what you are adding without any of those methods). What if in the next PR, we discover that we want to add multiple dependencies at once? You will be encouraged to just use this `addDependency` method multiple times which would trigger multiple calls the the database while the proper requirements would have been a method `addDependencies` with a list of dependencies.

Do not add dead code because you think we will need it.

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/services/EditingContextLifecycleTests.java
sbegaudeau: The standard library is in an SQL script and it could also be done in the Papaya example.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/services/api/ISemanticDataUpdateService.java
sbegaudeau: Given the shape of the API that you are suggesting in your next PR:

```
input ImportLibrariesInput {
  id: ID!
  editingContextId: ID!
  type: String!
  libraryIds: [String!]!
}
```

I suspect that my feeling was not completely wrong. I thus suspect that we would need a method like `SemanticData#addDependency(ICause, List<UUID>)` or even a method like `SemanticData#updateDependencies` and no add/remove methods at all just like we did it in `SemanticData#updateDocuments` and such a method would be very different and it may event offer a completely different event.

There's at least one thing I know, it's that I do not know what we will need and neither you nor I can guess it for the moment.

sbegaudeau (changes requested):

sbegaudeau:
I'll update the PR and ensure that it gets merged.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/api/IEditingContextDependencyLoader.java
sbegaudeau: Now that I can play with the code, I'm not sure that it was a good idea. On one hand, it would help us identify easily resources that have been created by libraries but on the other hand:

- It implies that the granularity of the import of a library is a resource, I'm not sure that it will apply in all use cases. One could want to load a dependency with a smaller granularity.
- It is also not enough for what we want given that we would like to attach some additional metadata (namespace, name, version).

Yet it is enough to prevent Sirius Web from persisting the resource. But if we can attach some metadata to the resource, we can also detect said metadata to prevent the save of the resource. I think that the metadata will be both necessary and more powerful while the dependency scheme will not be that relevant once we have the appropriate metadata.

sbegaudeau (commented):
I'll wait for what should be the final build and merge it just after.

sbegaudeau (approved):

PR #4655, opened 2025-02-28
[4509] Add a command to import libraries into a studio

### LINKED ISSUE #4509: Provide a command to import libraries into a studio

### Expected behavior
The behavior is described in [this shape](https://github.com/eclipse-sirius/sirius-web/blob/master/doc/iterations/2025.2/add_support_for_the_import_of_libraries.adoc)

The main goal is to have a new command in the palette that let users import published libraries. 
This command opens a dialog with a table containing the published libraries, and provides actions the user can execute on selected libraries (import as a copy, import in read-only mode, etc).

### Cuttings backs

We will only provide a default action for the first version of the import (import as a copy).

This work is based on the work in the following pull requests:

- https://github.com/eclipse-sirius/sirius-web/pull/4463
- https://github.com/eclipse-sirius/sirius-web/pull/4492

gdaniel:
# Pull request template

Based on https://github.com/eclipse-sirius/sirius-web/pull/4653

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextDependencyLoader.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextDependencyLoader.java
@@ -23,6 +23,7 @@ import org.eclipse.sirius.web.application.editingcontext.services.api.IEditingCo
 import org.eclipse.sirius.web.application.editingcontext.services.api.IResourceLoader;
 import org.eclipse.sirius.web.application.library.services.LibraryMetadataAdapter;
 import org.eclipse.sirius.web.domain.boundedcontexts.library.services.api.ILibrarySearchService;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.Document;
 import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.SemanticData;
 import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.services.api.ISemanticDataSearchService;
 import org.springframework.data.jdbc.core.mapping.AggregateReference;
@@ -59,16 +60,33 @@ public class EditingContextDependencyLoader implements IEditingContextDependency
                     .orElse(List.of());
             for (SemanticData semanticData : dependenciesSemanticData) {
                 var optionalLibrary = this.librarySearchService.findBySemanticData(AggregateReference.to(semanticData.getId()));
-                semanticData.getDocuments().forEach(document -> {
-                    var applyMigrationParticipants = this.migrationParticipantPredicates.stream().anyMatch(predicate -> predicate.test(editingContext.getId()));
-                    this.resourceLoader.toResource(emfEditingContext.getDomain().getResourceSet(), document.getId().toString(), document.getName(), document.getContent(), applyMigrationParticipants).ifPresent(resource -> {
-                        if (optionalLibrary.isPresent()) {
-                            var library = optionalLibrary.get();
-                            resource.eAdapters().add(new LibraryMetadataAdapter(library.getNamespace(), library.getName(), library.getVersion()));
-                        }
-                    });
-                });
+                semanticData.getDocuments().stream()
+                        .filter(document -> !this.isAlreadyLoaded(emfEditingContext, document))
+                        .forEach(document -> {
+                            var applyMigrationParticipants = this.migrationParticipantPredicates.stream().anyMatch(predicate -> predicate.test(editingContext.getId()));
+                            this.resourceLoader.toResource(emfEditingContext.getDomain().getResourceSet(), document.getId().toString(), document.getName(), document.getContent(), applyMigrationParticipants).ifPresent(resource -> {
+                                if (optionalLibrary.isPresent()) {
+                                    var library = optionalLibrary.get();
+                                    resource.eAdapters().add(new LibraryMetadataAdapter(library.getNamespace(), library.getName(), library.getVersion()));
+                                }
+                            });
+                        });
             }
         }
+    }
+
+    /**
+     * Uses to ensure that we are not loading twice a resource with the same identity.
+     *
+     * This could easily happen while trying to load two dependencies that both share a dependency to another one.
+     * We don't want to load the common dependency twice.
+     *
+     * @param editingContext The editing context
+     * @param document The document being loaded
+     * @return true if the document is matching an existing resource of the resource set, false otherwise
+     */
+    private boolean isAlreadyLoaded(IEMFEditingContext editingContext, Document document) {
+        return editingContext.getDomain().getResourceSet().getResources().stream()
+                .anyMatch(resource -> resource.getURI().path().substring(1).equals(document.getId().toString()));
     }
 }
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/CopyLibraryImportHandler.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/CopyLibraryImportHandler.java
@@ -1,0 +1,131 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.web.application.studio.services.library;
+
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.sirius.components.collaborative.api.ChangeDescription;
+import org.eclipse.sirius.components.collaborative.api.ChangeKind;
+import org.eclipse.sirius.components.collaborative.api.IEditingContextEventHandler;
+import org.eclipse.sirius.components.collaborative.api.Monitoring;
+import org.eclipse.sirius.components.collaborative.messages.ICollaborativeMessageService;
+import org.eclipse.sirius.components.core.api.ErrorPayload;
+import org.eclipse.sirius.components.core.api.IEditingContext;
+import org.eclipse.sirius.components.core.api.IInput;
+import org.eclipse.sirius.components.core.api.IPayload;
+import org.eclipse.sirius.components.core.api.SuccessPayload;
+import org.eclipse.sirius.components.emf.services.JSONResourceFactory;
+import org.eclipse.sirius.components.representations.Message;
+import org.eclipse.sirius.components.representations.MessageLevel;
+import org.eclipse.sirius.web.application.editingcontext.EditingContext;
+import org.eclipse.sirius.web.application.editingcontext.services.api.IEditingContextMigrationParticipantPredicate;
+import org.eclipse.sirius.web.application.editingcontext.services.api.IResourceLoader;
+import org.eclipse.sirius.web.application.library.dto.ImportLibrariesInput;
+import org.eclipse.sirius.web.domain.boundedcontexts.library.Library;
+import org.eclipse.sirius.web.domain.boundedcontexts.library.services.api.ILibrarySearchService;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.Document;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.SemanticData;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.services.api.ISemanticDataSearchService;
+import org.springframework.stereotype.Service;
+
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import reactor.core.publisher.Sinks.Many;
+import reactor.core.publisher.Sinks.One;
+
+/**
+ * Handles the import of a library as a copy.
+ *
+ * @author gdaniel
+ */
+@Service
+public class CopyLibraryImportHandler implements IEditingContextEventHandler {
+
+    private final ILibrarySearchService librarySearchService;
+
+    private final ISemanticDataSearchService semanticDataSearchService;
+
+    private final IResourceLoader resourceLoader;
+
+    private final List<IEditingContextMigrationParticipantPredicate> migrationParticipantPredicates;
+
+    private final ICollaborativeMessageService messageService;
+
+    private final Counter counter;
+
+    public CopyLibraryImportHandler(ILibrarySearchService librarySearchService, ISemanticDataSearchService semanticDataSearchService, IResourceLoader resourceLoader, List<IEditingContextMigrationParticipantPredicate> migrationParticipantPredicates, ICollaborativeMessageService messageService, MeterRegistry meterRegistry) {
+        this.librarySearchService = Objects.requireNonNull(librarySearchService);
+        this.semanticDataSearchService = Objects.requireNonNull(semanticDataSearchService);
+        this.resourceLoader = Objects.requireNonNull(resourceLoader);
+        this.migrationParticipantPredicates = Objects.requireNonNull(migrationParticipantPredicates);
+        this.messageService = Objects.requireNonNull(messageService);
+        this.counter = Counter.builder(Monitoring.EVENT_HANDLER)
+                .tag(Monitoring.NAME, this.getClass().getSimpleName())
+                .register(meterRegistry);
+    }
+
+    @Override
+    public boolean canHandle(IEditingContext editingContext, IInput input) {
+        return input instanceof ImportLibrariesInput importLibrariesInput
+                && Objects.equals(importLibrariesInput.type(), "copy");
+    }
+
+    @Override
+    public void handle(One<IPayload> payloadSink, Many<ChangeDescription> changeDescriptionSink, IEditingContext editingContext, IInput input) {
+        this.counter.increment();
+
+        String message = this.messageService.invalidInput(input.getClass().getSimpleName(), ImportLibrariesInput.class.getSimpleName());
+        IPayload payload = new ErrorPayload(input.id(), message);
+        ChangeDescription changeDescription = new ChangeDescription(ChangeKind.NOTHING, editingContext.getId(), input);
+
+        if (input instanceof ImportLibrariesInput importLibrariesInput && Objects.equals(importLibrariesInput.type(), "copy")) {
+            if (editingContext instanceof EditingContext siriusWebEditingContext) {
+                List<Library> libraries = this.librarySearchService.findAllById(importLibrariesInput.libraryIds().stream().map(UUID::fromString).toList());
+
+                Set<SemanticData> semanticDataToCopy = new LinkedHashSet<>();
+                for (Library library : libraries) {
+                    this.semanticDataSearchService.findById(library.getSemanticData().getId())
+                        .ifPresent(semanticData -> {
+                            semanticDataToCopy.add(semanticData);
+                            semanticDataToCopy.addAll(this.semanticDataSearchService.findAllDependenciesRecursivelyById(semanticData.getId()));
+                        });
+                }
+
+                for (SemanticData semanticData: semanticDataToCopy) {
+                    for (Document document : semanticData.getDocuments()) {
+                        Optional<Resource> optionalResource = this.resourceLoader.toResource(siriusWebEditingContext.getDomain().getResourceSet(), document.getId().toString(), document.getName(), document.getContent(),
+                                this.migrationParticipantPredicates.stream().anyMatch(predicate -> predicate.test(editingContext.getId())));
+                        optionalResource.ifPresent(resource -> {
+                            URI newResourceURI = new JSONResourceFactory().createResourceURI(UUID.randomUUID().toString());
+                            URI oldResourceURI = resource.getURI();
+                            siriusWebEditingContext.getDomain().getResourceSet().getURIConverter().getURIMap().put(oldResourceURI, newResourceURI);
+                            resource.setURI(newResourceURI);
+                        });
+                    }
+                }
+
+                payload = new SuccessPayload(input.id(), List.of(new Message("Libraries imported", MessageLevel.SUCCESS)));
+                changeDescription = new ChangeDescription(ChangeKind.SEMANTIC_CHANGE, editingContext.getId(), importLibrariesInput);
+            }
+        }
+        payloadSink.tryEmitValue(payload);
+        changeDescriptionSink.tryEmitNext(changeDescription);
+    }
+}
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/ReferenceLibraryImportHandler.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/ReferenceLibraryImportHandler.java
@@ -1,0 +1,136 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.web.application.studio.services.library;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+
+import org.eclipse.sirius.components.collaborative.api.ChangeDescription;
+import org.eclipse.sirius.components.collaborative.api.ChangeKind;
+import org.eclipse.sirius.components.collaborative.api.IEditingContextEventHandler;
+import org.eclipse.sirius.components.collaborative.api.Monitoring;
+import org.eclipse.sirius.components.collaborative.messages.ICollaborativeMessageService;
+import org.eclipse.sirius.components.core.api.ErrorPayload;
+import org.eclipse.sirius.components.core.api.IEditingContext;
+import org.eclipse.sirius.components.core.api.IInput;
+import org.eclipse.sirius.components.core.api.IPayload;
+import org.eclipse.sirius.components.core.api.SuccessPayload;
+import org.eclipse.sirius.components.representations.Message;
+import org.eclipse.sirius.components.representations.MessageLevel;
+import org.eclipse.sirius.web.application.UUIDParser;
+import org.eclipse.sirius.web.application.editingcontext.EditingContext;
+import org.eclipse.sirius.web.application.editingcontext.services.api.IEditingContextDependencyLoader;
+import org.eclipse.sirius.web.application.library.dto.ImportLibrariesInput;
+import org.eclipse.sirius.web.domain.boundedcontexts.library.Library;
+import org.eclipse.sirius.web.domain.boundedcontexts.library.services.api.ILibrarySearchService;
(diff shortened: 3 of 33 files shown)


packages/sirius-web/backend/sirius-web-application/src/main/resources/schema/libraries.graphqls
```diff
@@ -13,6 +13,7 @@ type ViewerLibrariesEdge {
 }
 
 type Library {
+  id: ID!
```
gdaniel, line 16: I put back the ID to ease my life with the library selection table.
If necessary I can remove it and use the `namespace:name:version` instead

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/api/ILibraryImportHandler.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.eclipse.sirius.components.core.api.IPayload;
+import org.eclipse.sirius.web.application.library.dto.ImportLibrariesInput;
+
+/**
+ * Handles the import of libraries.
+ *
+ * @author gdaniel
+ */
+public interface ILibraryImportHandler {
```
gdaniel: Remove this class, it is not used, this was part of a previous solution

packages/sirius-web/backend/sirius-web-application/src/main/resources/schema/libraries.graphqls
sbegaudeau: I haven't reviewed the whole thing but I've just stumble upon this comment. Prefer the namespace, name and version indeed. I'm not fan of exposing the concept of libraryId in our APIs, this is not how people should think about the identity of a library.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/library/services/LibrarySearchService.java
```diff
@@ -61,6 +62,16 @@ public Optional<Library> findById(UUID id) {
...
 
+    @Override
+    public List<Library> findAllById(Iterable<UUID> ids) {
+        return this.libraryRepository.findAllById(ids);
+    }
+
+    @Override
+    public Optional<Library> findByDocumentId(UUID documentId) {
+        return this.libraryRepository.findByDocumentId(documentId);
+    }
```
sbegaudeau: - The concept of document does not exists in the bounded context of the library, this highlight a lifecycle issue
- Even if it did, you cannot find the root of an aggregate using the identifier of one of its child element. In this case, the id of a document is not unique to a semantic data and thus to a library. Multiple libraries WILL have documents with the same id.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/api/ILibraryImportHandler.java
gdaniel: Done

packages/sirius-web/backend/sirius-web-application/src/main/resources/schema/libraries.graphqls
sbegaudeau: Let's keep that for now if it creates some additional complexity. I'll see later what we can do on that topic

gdaniel: It requires a bit of work on the table in the frontend to merge the `namespace:name:version` into a valid Material React Table row ID, and split it before sending the GraphQL query to the backend.
Since the mutation should allow to import multiple libraries, its input should provide a list of (`namespace`, `name`, `version`). I don't know GraphQL enough to know the best way to do that, it could maybe be something like
```
input PublishLibrariesInput {
    libraries: [LibraryMetadata!]!
}

type LibraryMetadata {
    namespace: String!
    name: String!
    version: String!
}
```

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/library/services/LibrarySearchService.java
gdaniel: This method doesn't exist anymore

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextDependencyLoader.java
```diff
@@ -60,13 +60,16 @@ public void loadDependencies(IEditingContext editingContext) {
...
                 semanticData.getDocuments().forEach(document -> {
-                    var applyMigrationParticipants = this.migrationParticipantPredicates.stream().anyMatch(predicate -> predicate.test(editingContext.getId()));
-                    this.resourceLoader.toResource(emfEditingContext.getDomain().getResourceSet(), document.getId().toString(), document.getName(), document.getContent(), applyMigrationParticipants).ifPresent(resource -> {
-                        if (optionalLibrary.isPresent()) {
-                            var library = optionalLibrary.get();
-                            resource.eAdapters().add(new LibraryMetadataAdapter(library.getNamespace(), library.getName(), library.getVersion()));
-                        }
-                    });
+                    if (emfEditingContext.getDomain().getResourceSet().getResources().stream().noneMatch(resource -> resource.getURI().path().substring(1).equals(document.getId().toString()))) {
+                        // The resource isn't already in the ResourceSet
```
gdaniel: I don't know if I should handle the case of a resource already in the resource set here or create a dedicated method to load some dependencies into the resource set.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/StudioLibraryPublicationHandler.java
```diff
@@ -69,14 +72,17 @@ public class StudioLibraryPublicationHandler implements ILibraryPublicationHandl
 
     private final IProjectSemanticDataSearchService projectSemanticDataSearchService;
 
+    private final ILibrarySearchService librarySearchService;
+
     private final IMessageService messageService;
 
-    public StudioLibraryPublicationHandler(IEditingContextSearchService editingContextSearchService, ISemanticDataCreationService semanticDataCreationService, IResourceToDocumentService resourceToDocumentService, IStudioLibraryDependencyCollector studioLibraryDependencyCollector, IProjectSemanticDataSearchService projectSemanticDataSearchService, IMessageService messageService) {
+    public StudioLibraryPublicationHandler(IEditingContextSearchService editingContextSearchService, ISemanticDataCreationService semanticDataCreationService, IResourceToDocumentService resourceToDocumentService, IStudioLibraryDependencyCollector studioLibraryDependencyCollector, IProjectSemanticDataSearchService projectSemanticDataSearchService, ILibrarySearchService librarySearchService, IMessageService messageService) {
```
gdaniel, line 79: I updated this class to take into account imported libraries in the publication process (we don't want to publish them, we want to create dependencies to them)

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/services/SemanticDataUpdateService.java
```diff
@@ -45,4 +46,12 @@ public void updateDocuments(ICause cause, AggregateReference<SemanticData, UUID>
             this.semanticDataRepository.save(semanticData);
         });
     }
+
+    @Override
+    public void addDependencies(ICause cause, AggregateReference<SemanticData, UUID> semanticDataId, List<AggregateReference<SemanticData, UUID>> dependencySemanticDataIds) {
```
gdaniel, line 51: I don't really know if the list of dependencies should be a `List<UUID>` or a `List<AggregateReference<SemanticData, UUID>>`

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/SemanticData.java
```diff
@@ -131,6 +131,15 @@ public void updateDocuments(ICause cause, Set<Document> newDocuments, Set<String
         }
     }
 
+    public void addDependencies(ICause cause, List<AggregateReference<SemanticData, UUID>> dependencySemanticDataIds) {
+        List<SemanticDataDependency> newDependencies = dependencySemanticDataIds.stream()
+            .map(SemanticDataDependency::new)
+            .toList();
+        this.dependencies.addAll(newDependencies);
+        this.lastModifiedOn = Instant.now();
+        this.registerEvent(new SemanticDataUpdatedEvent(UUID.randomUUID(), this.lastModifiedOn, cause, this));
```
gdaniel: I reused this event because right now we don't really need a dedicated event

packages/sirius-web/frontend/sirius-web-application/src/omnibox/ImportLibraryDialog.tsx
```diff
@@ -0,0 +1,191 @@
...
+  ImportStudioSplitButtonState,
+} from './ImportLibraryDialog.types';
+import { useCurrentProject } from '../views/edit-project/useCurrentProject';
+
+export const ImportLibraryDialog = ({ open, title, onClose }: ImportLibraryDialogProps) => {
+  const [state, setState] = useState<ImportLibraryDialogState>({
+    selectedLibraries: [],
+    actions: [
+      { id: 'import', label: 'Import' },
+      { id: 'copy', label: 'Copy' },
```
gdaniel, line 44: As we discussed, for now this is hardcoded, but could be requested from the backend in the future

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextDependencyLoader.java
sbegaudeau: I'm not fan of defensive programming so I don't think that this needs to be changed. If someone changes the data of the resource set, then it is responsible for creating a resource set that will work. In this case, how one could load a dependency to something that already exists? By adding two versions of the same library in the same editing context? Then the identity of the documents in the library published must become unique. This is not an issue that needs to be handled by the core parts of Sirius Web.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/CopyLibraryImportHandler.java
```diff
@@ -0,0 +1,131 @@
...
+                    this.semanticDataSearchService.findById(library.getSemanticData().getId())
+                        .ifPresent(semanticData -> {
+                            semanticDataToCopy.add(semanticData);
+                            semanticDataToCopy.addAll(this.semanticDataSearchService.findAllDependenciesRecursivelyById(semanticData.getId()));
+                        });
+                }
+
+                for (SemanticData semanticData: semanticDataToCopy) {
+                    for (Document document : semanticData.getDocuments()) {
+                        Optional<Resource> optResource = this.resourceLoader.toResource(siriusWebEditingContext.getDomain().getResourceSet(), document.getId().toString(), document.getName(), document.getContent(),
```
sbegaudeau: I think that this can be simplified, I'm not convinced that we need to mess with the URI converter for example.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/ReferenceLibraryImportHandler.java
```diff
@@ -0,0 +1,124 @@
...
+                        if (optionalSemanticData.get().getDependencies().stream().anyMatch(dependency -> dependency.dependencySemanticDataId().getId().equals(library.getSemanticData().getId()))) {
+                            this.logger.warn("Cannot add the dependency to library " + library.getNamespace() + ":" + library.getName() + ":" + library.getVersion() + ": the dependency already exists");
+                        }
+                        else {
+                            dependencies.add(AggregateReference.to(library.getSemanticData().getId()));
+                        }
+                    }
+                }
+                this.semanticDataUpdateService.addDependencies(input, AggregateReference.to(optionalSemanticData.get().getId()), dependencies);
+                this.editingContextDependencyLoader.loadDependencies(editingContext);
```
sbegaudeau, line 126: Wouldn't that trigger a reloading of all the dependencies of the editing context? (Which may explain why you need to protect yourself against duplicates)

packages/sirius-web/backend/sirius-web-application/src/main/resources/schema/libraries.graphqls
sbegaudeau: I suspect that in the end, the id of a library could literally become `namespace:name:version` since none of those property is mutable.

packages/sirius-web/backend/sirius-web-application/src/main/resources/schema/libraries.graphqls
```diff
@@ -34,3 +37,12 @@ input PublishLibrariesInput {
 }
 
 union PublishLibrariesPayload = ErrorPayload | SuccessPayload
+
+input ImportLibrariesInput {
+  id: ID!
+  editingContextId: ID!
+  type: String!
+  libraryIds: [String!]!
```
sbegaudeau: [ID!]!

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/SemanticData.java
sbegaudeau: When someone will ask for it, we will introduce events with a smaller granularity

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/SemanticData.java
```diff
@@ -131,6 +131,15 @@ public void updateDocuments(ICause cause, Set<Document> newDocuments, Set<String
         }
     }
 
+    public void addDependencies(ICause cause, List<AggregateReference<SemanticData, UUID>> dependencySemanticDataIds) {
+        List<SemanticDataDependency> newDependencies = dependencySemanticDataIds.stream()
```
sbegaudeau: We should probably check that we have not duplicates dependencies

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/services/SemanticDataUpdateService.java
sbegaudeau: `List<AggregateReference<SemanticData, UUID>>` is fine

packages/sirius-web/frontend/sirius-web-application/src/omnibox/ImportLibraryCommand.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <>
+      <ListItemButton key={command.id} data-testid={command.label} onClick={handleClick} onKeyDown={onKeyDown}>
+        <ListItemIcon>
+          <IconOverlay iconURL={command.iconURLs} alt={command.label} />
+        </ListItemIcon>
+        <ListItemText sx={{ whiteSpace: 'nowrap', textOverflow: 'ellipsis' }}>{command.label}</ListItemText>
+      </ListItemButton>
+      {state.open && <ImportLibraryDialog open={state.open} title={'Import libraries'} onClose={onClose} />}
```
sbegaudeau: no need for `={'`

packages/sirius-web/frontend/sirius-web-application/src/omnibox/LibrariesImportTable.tsx
```diff
@@ -0,0 +1,120 @@
...
+} from 'material-react-table';
+import { useState, useEffect, useMemo } from 'react';
+import { useLibraries } from '../views/library-browser/useLibraries';
+import { GQLLibrary } from '../views/library-browser/useLibraries.types';
+import { LibrariesImportTableProps, LibrariesImportTableState } from './LibrariesImportTable.types';
+
+export const LibrariesImportTable = ({ onSelectionChange }: LibrariesImportTableProps) => {
+  const [state, setState] = useState<LibrariesImportTableState>({
+    data: null,
+  });
```
sbegaudeau: You don't need to store the result of a query in a state

sbegaudeau (changes requested):
I'll update the PR, test it manually and ensure that it gets merged.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/CopyLibraryImportHandler.java
gdaniel: Ok, here is my reasoning in case it helps you find a better solution:
- The documents of the library may contain proxies to other documents of the library (or dependencies of the library). We don't load the dependencies in any particular order, so there will be some proxies.
- If I load all the documents into resources with an URI based on the document's ID, I am sure the proxies can be resolved
- I can't keep the same ID because of the unicity of `document#id` in the database, so I have to change the ID of the resource to be able to persist it.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/ReferenceLibraryImportHandler.java
gdaniel: I need to protect myself against duplicates in this context:
- I imported some libraries in my editing context
- I import some additional libraries in my editing context, and these libraries (or the one they depend on) are already in my editing context.
I don't want to have these libraries twice in my editing context (for example, if I import 2 descriptions one after the other, I'll have 2 domains and 2 shared_components). This is maybe the point we do not agree on: based on your comment above, I understand this may not be a Sirius Web concern.

In any case, `ILibrarySearchService#findAllDependenciesRecursivelyById` does not return duplicates, so there won't be duplicated libraries when loading an editing context, so it seems reasonable to me to ensure that this won't be the case on individual imports.

This being said, we could have a method to load some dependencies in the editing context, and not reload everything. In this case, (and assuming we don't want to have the same library twice in the editing context), we should still deal with duplicates, because the new dependency may create resources that already exist (the new dependency itself but also the transitive ones). In addition, there is nothing for now in Sirius Web that prevents you from importing the same library twice. If we don't check for duplicates we will have them twice in the editing context, but when we reload the editing context we'll have them once (because of the search service that filters duplicates), this is not consistent.

This is why I went for the easy way of reloading everything. This is not efficient, but it relies on the search service to ensure there are no duplicates in the transitive dependencies, and if there are some resources already loaded we do not reload them.

Long story short: if we want to be consistent, we should either allow duplicates at every point of the lifecycle, or forbid them. There are probably more efficient way to do what I did, but I hope this clarifies why I did it this way.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/SemanticData.java
gdaniel: True, this is not visible at the moment because `ReferenceLibraryImportHandler` checks that the new dependency is not already in the dependencies of the editing context, but it should be checked in `SemanticData`.
Btw, I think we should also return early if the list of dependencies is empty, and not update the `lastModifiedOn` field nor send the event.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/CopyLibraryImportHandler.java
sbegaudeau: I understand more what you are trying to do but what if I have a dependency to a library A and I'm trying to copy the very same library in my resource set? I suspect that by messing with the URIConverter, you will have an issue with the dependency to the library.

I think we should load the library in a different resource set, change the identifiers there (all of them including the object ones not just the one from the document) and then copy the content into the new resource set. Otherwise you will have multiple objects with the same identifier in the resource set (which is invalid).

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/ReferenceLibraryImportHandler.java
sbegaudeau: I did not take into account diamond dependencies when I read this part of the code with things like a project `P` using a library `A` which depends on `C`:

```
   P
 /    
A      
 \    
   C
```

And then adding a new dependency to the library `B` which also depends on `C`

```
   P
 /   \
A     B
 \   /
   C
```

So indeed, we will have to protect ourselves in some way against duplicated dependencies. It will be interesting to see if we can have two versions of the same library in the same editing context.

- One one hand, two versions of the same library should have the same identity (documentId, objectId, etc) otherwise those are not the same and it would be complicated to upgrade from one to the other. The publication or import mechanism would be more complex because we would have to re-create tons of identifiers. I don't think we want that.
- One the other hand, if they have the same identity, how could we allow importing two versions of the same library because we wouldn't be able to resolve the links properly with `IObjectSearchService#getObject`.

We will have to think about that when we will test real upgrade use cases because I know for a fact that we will both want:

- to upgrade easily
- to have multiple versions of the same dependency in an editing context (but maybe not right now in the very first version)

So we will have to think about that in an ADR later.

packages/sirius-web/backend/sirius-web-application/src/main/resources/schema/libraries.graphqls
sbegaudeau: Done

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/SemanticData.java
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/omnibox/ImportLibraryCommand.tsx
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/omnibox/LibrariesImportTable.tsx
sbegaudeau: Done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextDependencyLoader.java
sbegaudeau: This part is necessary in the end for diamond dependency trees, I'll just document it

sbegaudeau: hmmm.... I'm torned on this issue because as discussed on the other comment, the current path of the code should ensure that this will not happen but we need something else to load additional dependencies...I'll test a couple of things to see how it behaves. It may not be that necessary in the end.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/ReferenceLibraryImportHandler.java
sbegaudeau: So I've looked into it and tried to only load the new dependencies but if we consider the following use case. We have loaded a project P with the following dependency tree.

- Project P
  - Library A
    - Library C
  - Library B
    - Library C

We have the content of P, A, C and B in our resource set (with C loaded once only). Now we want to add the following libraries:

- Library D
  - Library A
    - Library C
- Library E
  - Library C

If we want to only consider the following libraries, we need to find recursively their semantic data, remove any duplicates (to only consider D, A, C, E) and try to load only those in the editing context. Sine it already contains A and C, we need to detect those duplicate and load only D and E. At the end of the day, that's exactly the same code as `IEditingContextDependencyLoader` but we may load less dependencies in some use cases but we won't have the same guarantees in term of consistency... So using `IEditingContextDependencyLoader` may really be the smartest choice in the end. So we will need to ensure that we haven't loaded the same document twice in there.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/CopyLibraryImportHandler.java
```diff
@@ -0,0 +1,131 @@
...
+            if (editingContext instanceof EditingContext siriusWebEditingContext) {
+                List<Library> libraries = this.librarySearchService.findAllById(importLibrariesInput.libraryIds().stream().map(UUID::fromString).toList());
+
+                Set<SemanticData> semanticDataToCopy = new LinkedHashSet<>();
+                for (Library library : libraries) {
+                    this.semanticDataSearchService.findById(library.getSemanticData().getId())
+                        .ifPresent(semanticData -> {
+                            semanticDataToCopy.add(semanticData);
+                            semanticDataToCopy.addAll(this.semanticDataSearchService.findAllDependenciesRecursivelyById(semanticData.getId()));
+                        });
```
sbegaudeau, line 108: What if the library being copied and the current editing context are sharing a common dependency, wouldn't that dependency be copied as well instead of being reused? I don't think I'll block the PR to handle this but we will need to think clearly in a dedicated shape to all the use cases that we will have to consider and create dedicated test for them.

sbegaudeau: So the question is if I have a project depending on B and I'm copying a library A which depend on B in my project, should I copy B in the project or use the existing reference to B? What if instead I copy a library A depending on B and my project does not depend on B, what should be do? Add a dependency to B or copy B.

Technically, I could open a SysON workspace in my IDE and open the code of Papyrus Web in the same IDE and both would reference the code of Sirius Web using maven. Or I could both import the code of Papyrus Web and Sirius Web in my IDE and then all the code of three projects SysON, Papyrus Web and Sirius Web would reference using maven EMFJson. Both are valid use cases and we will need to think about the consequences... not today, since we need to start with a first version but copying may not be as simple as copy everything.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/CopyLibraryImportHandler.java
sbegaudeau: Another fun question, for both copying and referencing a library, what if the library does not use the same EPackages as my editing context? Right now we assume that everything is working nicely because everyone is using the same metamodels but we can't really make this assumption. We may need to load the relevant metamodels in `IEditingContextDependencyLoader#loadDependencies`.

sbegaudeau (commented):
SO this first version seems to be working but there are still tons of minor issues to be ironed out. For example, it's clearly not visible in the workbench that something is a library and should be treated as read only. We will need a better support for `IReadOnlyObjectPredicate` because it took me a couple of minutes to figure out that it was indeed working ^^

I'll update the PR and merge it after the build. I'll work on a more complex Papaya example to let us challenge the current state of our library support.

sbegaudeau (approved):

Issue #4590, opened 2025-02-18
It should be easier to quickly invoke the same tool multiple times in a row


pcdavid:
Currently, even with the palette remembering the last tool invoked, this requires several clicks/mouse move for each repeated invocation.

PR #4688, opened 2025-03-11
[cleanup] Extract hooks from usePalette

pcdavid:
This is preparatory work to make the changes needed for #4590 more manageable.

- [x] Extract useCollapseExpand hook
- [x] Extract useDelete hook
- [x] Extract usePaletteContents
- [x] Extract useSingleClickTool hook

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [x] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.tsx
+++ b/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.tsx
@@ -1,0 +1,106 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+import { gql, useQuery } from '@apollo/client';
+import { useContext, useEffect } from 'react';
+
+import { useMultiToast } from '@eclipse-sirius/sirius-components-core';
+import { DiagramContext } from '../../contexts/DiagramContext';
+import { DiagramContextValue } from '../../contexts/DiagramContext.types';
+import { GQLPalette } from './Palette.types';
+import {
+  GQLDiagramDescription,
+  GQLGetToolSectionsData,
+  GQLGetToolSectionsVariables,
+  GQLRepresentationDescription,
+  UsePaletteContentValue,
+} from './usePaletteContents.types';
+
+export const getPaletteQuery = gql`
+  fragment ToolFields on Tool {
+    __typename
+    id
+    label
+    iconURL
+    ... on SingleClickOnDiagramElementTool {
+      targetDescriptions {
+        id
+      }
+      appliesToDiagramRoot
+      dialogDescriptionId
+      withImpactAnalysis
+    }
+  }
+  query getPalette($editingContextId: ID!, $diagramId: ID!, $diagramElementId: ID!) {
+    viewer {
+      editingContext(editingContextId: $editingContextId) {
+        representation(representationId: $diagramId) {
+          description {
+            ... on DiagramDescription {
+              palette(diagramElementId: $diagramElementId) {
+                id
+                quickAccessTools {
+                  ...ToolFields
+                }
+                paletteEntries {
+                  ...ToolFields
+                  ... on ToolSection {
+                    id
+                    label
+                    iconURL
+                    tools {
+                      ...ToolFields
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+const isDiagramDescription = (
+  representationDescription: GQLRepresentationDescription
+): representationDescription is GQLDiagramDescription => representationDescription.__typename === 'DiagramDescription';
+
+export const usePaletteContents = (diagramElementId: string): UsePaletteContentValue => {
+  const { diagramId, editingContextId } = useContext<DiagramContextValue>(DiagramContext);
+  const { addErrorMessage } = useMultiToast();
+
+  const {
+    data: paletteData,
+    loading,
+    error: paletteError,
+  } = useQuery<GQLGetToolSectionsData, GQLGetToolSectionsVariables>(getPaletteQuery, {
+    variables: {
+      editingContextId,
+      diagramId,
+      diagramElementId,
+    },
+  });
+
+  const description: GQLRepresentationDescription | undefined =
+    paletteData?.viewer.editingContext.representation.description;
+  const palette: GQLPalette | null = description && isDiagramDescription(description) ? description.palette : null;
+
+  useEffect(() => {
+    if (paletteError) {
+      addErrorMessage('An unexpected error has occurred, please refresh the page');
+    }
+  }, [paletteError]);
+
+  return { palette, loading };
+};
--- a/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.types.ts
+++ b/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.types.ts
@@ -1,0 +1,49 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+import { GQLPalette } from './Palette.types';
+
+export interface UsePaletteContentValue {
+  palette: GQLPalette | null;
+  loading: boolean;
+}
+
+export interface GQLDiagramDescription extends GQLRepresentationDescription {
+  palette: GQLPalette;
+}
+
+export interface GQLRepresentationDescription {
+  __typename: string;
+}
+
+export interface GQLGetToolSectionsVariables {
+  editingContextId: string;
+  diagramId: string;
+  diagramElementId: string;
+}
+
+export interface GQLGetToolSectionsData {
+  viewer: GQLViewer;
+}
+
+export interface GQLViewer {
+  editingContext: GQLEditingContext;
+}
+
+export interface GQLEditingContext {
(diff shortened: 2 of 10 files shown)


packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useDelete.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+`;
+
+export const useDelete = (editingContextId: string, diagramId: string): UseDeleteValue => {
+  const [rawDeleteFromDiagram, rawDeleteFromDiagramResult] = useMutation<
+    GQLDeleteFromDiagramData,
+    GQLDeleteFromDiagramVariables
+  >(deleteFromDiagramMutation);
+
+  const deleteDiagramElements = useCallback(
```
sbegaudeau: I think we could get rid of those useCallback if we just asked for everything we need when it's relevant and not in two different places (signature of the hook and in the function here) at once for now real reason. The same comment could be applied elsewhere.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.tsx
```diff
@@ -0,0 +1,103 @@
...
+      }
+    }
+  }
+`;
+
+const isDiagramDescription = (
+  representationDescription: GQLRepresentationDescription
+): representationDescription is GQLDiagramDescription => representationDescription.__typename === 'DiagramDescription';
+
+export const usePaletteContents = (diagramElementId: string): GQLPalette | null => {
```
sbegaudeau: From experience, the signature of those hooks tends to evolve so please encapsulate the result in an object to prepare it for future needs (if we need to track the loading for example)

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.types.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export interface GQLEditingContext {
+  representation: GQLRepresentationMetadata;
+}
+
+export interface GQLRepresentationMetadata {
+  id: string;
+  label: string;
+  kind: string;
```
sbegaudeau: You don't have access to those fields in your query

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.types.ts
```diff
@@ -0,0 +1,48 @@
...
+ *******************************************************************************/
+
+import { GQLPalette } from './Palette.types';
+
+export interface GQLDiagramDescription extends GQLRepresentationDescription {
+  palette: GQLPalette;
+}
+
+export interface GQLRepresentationDescription {
+  id: string;
```
sbegaudeau: This field is not yet available and thus not requested in your query

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useSingleClickTool.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const { addMessages } = useMultiToast();
+  const { setSelection } = useSelection();
+  const { showDialog } = useDialog();
+
+  const [invokeSingleClickOnDiagramElementTool] = useMutation<
+    GQLInvokeSingleClickOnDiagramElementToolData,
+    GQLInvokeSingleClickOnDiagramElementToolVariables
+  >(invokeSingleClickOnDiagramElementToolMutation);
+
+  const invokeSingleClickTool = useCallback(
```
sbegaudeau: Same comment, we should get rid of the parameter in the signature of the hook and ask instead for the parameters in the signature of the invokeSingleClickTool. We would be able to remove those useless useCallback without any impact on the performance.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useSingleClickTool.tsx
```diff
@@ -0,0 +1,136 @@
...
+    async (tool: GQLTool, diagramElementId: string, variables: GQLToolVariable[], x: number, y: number) => {
```
sbegaudeau: We should also use the same pattern as usual here, the use of async/await here is trying to hide a lifecycle issue in the way those tools are called. Technically, we are asking for the palette to be closed before calling the tool and thus the component is unmounted and we can't do anything afterward. All this code here is executed on an unmounted component and it is thus very brittle.

This improvement can be done in another contribution but just so you know, nothing here in the way this code is written is normal. This is the reason why I need our mutation hooks to expose `data` because otherwise we will unmount the component before having an answer from the backend.

The main culprit here is:

```
const handleToolClick = (tool: GQLTool) => {
    closeAllPalettes();

   // Call some mutation
}
```

which should be instead:

```
const handleToolClick = (tool: GQLTool) => {
   // Call some mutation
}

useEffect(() => {
  if(isSuccess(mutationData)) {
    closeAllPalettes();
  }
}, [mutationData])
```

And since we are on this subject, the fact that usePalette is depending on `closeAllPalettes` is among the next fight against this `usePalette` hook. A child component, the palette with its hook, should not know and thus manipulate a hook from its parent component. It should instead have a props like `onClose` and each parent component should manipulate its own hook.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useCollapseExpand.tsx
```diff
@@ -0,0 +1,70 @@
...
+        collapsingState,
+      };
+      rawCollapseExpand({ variables: { input } });
+    },
+    [editingContextId, diagramId, rawCollapseExpand]
+  );
+
+  useReporting(rawCollapseExpandResult, (data: GQLUpdateCollapsingStateData) => data.collapseExpandDiagramElement);
+
+  return { collapseExpandElement };
```
sbegaudeau: Please expose `loading` and the resulting `data: GQLUpdateCollapsingStateData | null` as usual. I know, no one is using it right know but it's a bug

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useDelete.tsx
```diff
@@ -0,0 +1,68 @@
...
+        deletionPolicy,
+      };
+      rawDeleteFromDiagram({ variables: { input } });
+    },
+    [editingContextId, diagramId, rawDeleteFromDiagram]
+  );
+
+  useReporting(rawDeleteFromDiagramResult, (data: GQLDeleteFromDiagramData) => data.deleteFromDiagram);
+
+  return { deleteDiagramElements };
```
sbegaudeau: Same comment regarding `loading` and `data: GQLDeleteFromDiagramData | null`

sbegaudeau (changes requested):

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.types.ts
pcdavid: It was there [before](https://github.com/eclipse-sirius/sirius-web/blob/master/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePalette.types.ts#L163), I only moved it. But OK, removed.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.types.ts
pcdavid: Other leftovers from [before](https://github.com/eclipse-sirius/sirius-web/blob/master/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePalette.types.ts#L157-L159). Cleaned up.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.tsx
pcdavid: Done.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useCollapseExpand.tsx
pcdavid: Done.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useDelete.tsx
pcdavid: Done.

pcdavid: Done.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useSingleClickTool.tsx
pcdavid: Done.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useSingleClickTool.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const { setSelection } = useSelection();
+  const { showDialog } = useDialog();
+  const { showImpactAnalysisDialog } = useImpactAnalysisDialog();
+
+  const [invokeSingleClickOnDiagramElementTool] = useMutation<
+    GQLInvokeSingleClickOnDiagramElementToolData,
+    GQLInvokeSingleClickOnDiagramElementToolVariables
+  >(invokeSingleClickOnDiagramElementToolMutation);
+
+  const invokeSingleClickTool = async (
```
sbegaudeau: Does this still need to be `async`?

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useSingleClickTool.tsx
```diff
@@ -0,0 +1,173 @@
...
+    editingContextId: string,
+    diagramId: string,
+    tool: GQLTool,
```
sbegaudeau: I think that all the code paths that lead to this, have a tool with the type `GQLSingleClickOnDiagramElementTool` so the check line 84 is not necessary anymore.

sbegaudeau (changes requested):
I'll look into my last two comments and rebase the PR to merge it as soon as possible.

sbegaudeau (approved):

PR #4702, opened 2025-03-14
[4694] Add a contextual menu action to update imported libraries

### LINKED ISSUE #4694: Provide a contextual menu action to update an imported library

### Expected behavior
A contextual menu action "Upgrade Library" is available in the explorer, on `TreeItem` elements that correspond to libraries.
This action opens a dialog containing a table with the different versions of the library, and an _upgrade_ button. 
The user can select **one** version of the library and click on the upgrade button.
The initial library is replaced with the new version, and references to elements in the initial library are updated to their newer version when possible.

### Cuttings backs
The dialog does not open an impact analysis dialog for the moment, since this feature hasn't been merged in Sirius Web.
The initial implementation will be based on Sirius Web identifiers to retrieve elements across versions of the library (same ID = same element).

We discarded the option of having a command to perform the upgrade. This would imply a two-step process where the user first selects the library they want to upgrade, then select the version to upgrade to.

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/4694

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


packages/core/backend/sirius-components-collaborative/src/main/resources/schema/core.graphqls
```diff
@@ -26,6 +26,7 @@ type EditingContext {
   queryBasedObject(query: String!, variableName: String): Object
   queryBasedObjects(query: String!, variableName: String): [Object]
   actions: EditingContextActionsConnection!
+  object(objectId: ID!): Object
```
gdaniel, line 29: As we discussed, this is needed by the frontend to retrieve the library containing an object

packages/emf/backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/handlers/EditingContextObjectEventHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    public boolean canHandle(IEditingContext editingContext, IInput input) {
+        return input instanceof EditingContextObjectInput;
+    }
+
+    @Override
+    public void handle(Sinks.One<IPayload> payloadSink, Sinks.Many<ChangeDescription> changeDescriptionSink, IEditingContext editingContext, IInput input) {
+        Optional<Object> object = Optional.empty();
+        if (input instanceof EditingContextObjectInput objectInput) {
+            object = this.objectService.getObject(editingContext, objectInput.objectId());
+            if (object.isEmpty() && editingContext instanceof IEMFEditingContext emfEditingContext) {
```
gdaniel: `IObjectService` does not return resources, but we want to get them if the object is a resource.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/UpdateLibraryHandler.java
```diff
@@ -0,0 +1,249 @@
...
+
+                        for (Resource oldLibraryResource : oldLibraryResources) {
+                            siriusWebEditingContext.getDomain().getResourceSet().getResources().remove(oldLibraryResource);
+                            oldLibraryResource.unload();
+                        }
+                        this.removeLibraryDependency(updateLibraryInput, siriusWebEditingContext, oldLibrary);
+                        payload = new SuccessPayload(input.id(), List.of(new Message("Library " + oldLibrary.getName() + " updated to version " + newLibrary.getVersion(), MessageLevel.SUCCESS)));
+                        changeDescription = new ChangeDescription(ChangeKind.SEMANTIC_CHANGE, editingContext.getId(), input);
+                    } else {
+                        // We do not support update of transitive libraries for the moment.
```
gdaniel: Updating a transitive dependency is a different animal: since we can't change the dependencies of an imported library, we have to first import the transitive dependency we want to update, then update it.
I am not even sure this is something we want, especially right now since we don't have a strategy to handle multiple versions of the same library in the workspace (which is necessary to make sure reloading the project doesn't load the original transitive dependency AND the imported one).
I feel this can be added later, and should probably be carefully specified first.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/UpdateLibraryHandler.java
```diff
@@ -0,0 +1,249 @@
...
+                Optional<Library> optNewLibrary = this.librarySearchService.findById(UUID.fromString(updateLibraryInput.libraryId()));
+                if (optNewLibrary.isPresent()) {
+                    Library newLibrary = optNewLibrary.get();
+                    // Find the version of the library already in the dependencies.
+                    Optional<Library> optOldLibrary = this.getLibraryDependencyWithDifferentVersion(siriusWebEditingContext, newLibrary);
+                    if (optOldLibrary.isPresent()) {
+                        Library oldLibrary = optOldLibrary.get();
+                        List<Resource> oldLibraryResources = this.getAllResourcesFromLibraryRecursiveDependencies(siriusWebEditingContext, oldLibrary);
+                        this.addLibraryDependency(updateLibraryInput, siriusWebEditingContext, newLibrary);
+                        this.editingContextDependencyLoader.loadDependencies(siriusWebEditingContext);
```
gdaniel: As we discussed in the import discussion, we need to reload all the dependencies to make sure we don't have duplicates

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/UpdateLibraryHandler.java
```diff
@@ -0,0 +1,249 @@
...
+                this.semanticDataUpdateService.addDependencies(updateLibraryInput, AggregateReference.to(optionalSemanticData.get().getId()), List.of(library.getSemanticData()));
+            }
+        }
+    }
+
+    private void removeLibraryDependency(UpdateLibraryInput updateLibraryInput, EditingContext editingContext, Library library) {
+        new UUIDParser().parse(editingContext.getId())
+            .flatMap(this.semanticDataSearchService::findById)
+            .ifPresent(semanticData -> {
+                this.semanticDataUpdateService.removeDependencies(updateLibraryInput, AggregateReference.to(semanticData.getId()), List.of(library.getSemanticData()));
```
gdaniel: This was easier for me, but I suspect we should have an `updateDependencies` method instead (as you suggested some PRs ago).

packages/sirius-web/frontend/sirius-web-application/src/views/library-browser/useLibraries.ts
```diff
@@ -43,11 +43,8 @@ export const useLibraries = (page: number, limit: number): UseLibrariesValue =>
     }
   }, [error]);
 
-  const refreshLibraries = (page: number, limit: number) => refetch({ page, limit });
-
   return {
     data: data ?? null,
     loading,
-    refreshLibraries,
```
gdaniel, line 51: Cleanup, this was part of a previous implementation of the library-browser.

packages/core/backend/sirius-components-collaborative/src/main/java/org/eclipse/sirius/components/collaborative/dto/EditingContextObjectPayload.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.UUID;
+
+import org.eclipse.sirius.components.core.api.IPayload;
+
+/**
+ * The payload for the query field object.
+ *
+ * @author gdaniel
+ */
+public record EditingContextObjectPayload(UUID id, Optional<Object> object) implements IPayload {
```
sbegaudeau: Don't use optionals in records like that, especially to require that the optional should not be null, that's very odd

packages/emf/backend/sirius-components-emf/src/main/java/org/eclipse/sirius/components/emf/handlers/EditingContextObjectEventHandler.java
sbegaudeau: This should be modified in the core services instead of adding yet another exception. It will create other problems since the other core services are not aligned.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/controllers/ObjectLibraryDataFetcher.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public LibraryDTO get(DataFetchingEnvironment environment) throws Exception {
+        Object object = environment.getSource();
+        Resource resource = null;
+        if (object instanceof Resource objectResource) {
+            resource = objectResource;
+        } else if (object instanceof EObject eObject) {
+            resource = eObject.eResource();
+        }
```
sbegaudeau: Why, we should just look for the adapter where it is being asked. An object could be a library without its containing resource being one.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/dto/UpdateLibraryInput.java
```diff
@@ -0,0 +1,28 @@
...
+import org.eclipse.sirius.components.core.api.IInput;
+
+import jakarta.validation.constraints.NotNull;
+
+/**
+ * Input used to update a library.
+ *
+ * @author gdaniel
+ */
+public record UpdateLibraryInput(@NotNull UUID id, @NotNull String editingContextId, @NotNull String libraryId) implements IInput {
```
sbegaudeau: Do you have all the information you need to do that?

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/UpdateLibraryHandler.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    private Optional<Library> getLibraryDependencyWithDifferentVersion(EditingContext editingContext, Library library) {
+        return new UUIDParser().parse(editingContext.getId())
+            .flatMap(this.semanticDataSearchService::findById)
+            .map(SemanticData::getDependencies)
+            .orElse(List.of())
+            .stream()
+            .map(SemanticDataDependency::dependencySemanticDataId)
+            .map(this.librarySearchService::findBySemanticData)
```
sbegaudeau: Are you loading the content of all the dependencies from scratch from the database in order to only find one? This is very expensive.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/UpdateLibraryHandler.java
```diff
@@ -0,0 +1,249 @@
...
+                        }
+                    }
+                }
+            });
+        }
+    }
+
+    private Optional<EObject> getMatchingObject(EObject eObject, List<Resource> resources) {
+        String eObjectId = this.identityService.getId(eObject);
+        // Do not use IObjectService here: we are in a state where multiple objects have the same ID in the editing context.
```
sbegaudeau: This should never happen. There's something odd in the lifecycle of the feature if it does.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/explorer/services/ExplorerTreeItemContextMenuEntryProvider.java
```diff
@@ -0,0 +1,109 @@
...
+    public List<ITreeItemContextMenuEntry> getTreeItemContextMenuEntries(IEditingContext editingContext, TreeDescription treeDescription, Tree tree, TreeItem treeItem) {
+        List<ITreeItemContextMenuEntry> result = new ArrayList<>();
+        this.getUpdateLibraryEntry(editingContext, treeItem).ifPresent(result::add);
+        return result;
+    }
+
+    private Optional<ITreeItemContextMenuEntry> getUpdateLibraryEntry(IEditingContext editingContext, TreeItem treeItem) {
+        Optional<ITreeItemContextMenuEntry> result = Optional.empty();
+        if (editingContext instanceof IEMFEditingContext emfEditingContext) {
+            Optional<Resource> optResource = emfEditingContext.getDomain().getResourceSet().getResources().stream()
```
sbegaudeau: A library may not be a resource

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/explorer/services/ExplorerTreeItemContextMenuEntryProvider.java
```diff
@@ -0,0 +1,109 @@
...
+                        // We do not support the update of a transitive dependency for the moment.
+                        result = Optional.of(new SingleClickTreeItemContextMenuEntry("updateLibrary", "", List.of()));
+                    }
+                }
+            }
+        }
+        return result;
+    }
+
+    private boolean isDirectDependency(LibraryMetadataAdapter libraryMetadata, IEMFEditingContext emfEditingContext) {
```
sbegaudeau: I'll have to look into this but that seems quite expensive.

packages/sirius-web/backend/sirius-web-application/src/main/resources/schema/libraries.graphqls
```diff
@@ -17,6 +17,7 @@ type Library {
   namespace: String!
   name: String!
   version: String!
+  versions(page: Int!, limit: Int!): ViewerLibrariesConnection!
```
sbegaudeau: `ViewerLibrariesConnection`??? This has nothing to do with the GraphQL field `Viewer#libraries` at all. It should be `LibraryVersionsConnection`

packages/sirius-web/backend/sirius-web-application/src/main/resources/schema/libraries.graphqls
```diff
@@ -25,9 +26,14 @@ type Library {
 extend type Mutation {
   publishLibraries(input: PublishLibrariesInput!): PublishLibrariesPayload!
   importLibraries(input: ImportLibrariesInput!): ImportLibrariesPayload!
+  updateLibrary(input: UpdateLibraryInput!): UpdateLibraryPayload!
   
 }
 
+extend type Object {
+	library: Library
```
sbegaudeau: formatting

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/SemanticData.java
```diff
@@ -146,6 +146,12 @@ public void addDependencies(ICause cause, List<AggregateReference<SemanticData,
         }
     }
 
+    public void removeDependencies(ICause cause, List<AggregateReference<SemanticData, UUID>> dependencySemanticDataIds) {
+        this.dependencies.removeIf(dependency -> dependencySemanticDataIds.contains(dependency.dependencySemanticDataId()));
+        this.lastModifiedOn = Instant.now();
+        this.registerEvent(new SemanticDataUpdatedEvent(UUID.randomUUID(), this.lastModifiedOn, cause, this));
+    }
```
sbegaudeau, line 153: - Why removeDependencies? You cannot guarantee the consistency of the data with a removal of a dependency.
- You can update the last modified on date and fire an event even if nothing has been modified.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/semanticdata/services/SemanticDataUpdateService.java
```diff
@@ -54,4 +54,12 @@ public void addDependencies(ICause cause, AggregateReference<SemanticData, UUID>
...
         });
     }
+
+    @Override
+    public void removeDependencies(ICause cause, AggregateReference<SemanticData, UUID> semanticDataId, List<AggregateReference<SemanticData, UUID>> dependencySemanticDataIds) {
+        this.semanticDataRepository.findById(semanticDataId.getId()).ifPresent(semanticData -> {
+            semanticData.removeDependencies(cause, dependencySemanticDataIds);
+            this.semanticDataRepository.save(semanticData);
+        });
+    }
```
sbegaudeau, line 64: Same question, can you guarantee that you will have a consistent state after this operation? I'm really not sure.

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/trees/TreeItemContextMenuControllerTests.java
```diff
@@ -183,6 +185,46 @@ public void givenAStudioWhenTheContextMenuActionsAreRequestedOnAnEntityThenTheCo
...
+                    "representationId", treeId.get(),
+                    "treeItemId", PapayaIdentifiers.PAPAYA_SIRIUS_WEB_TESTS_DATA_DOCUMENT.toString()
+            );
+            var result = this.treeItemContextMenuQueryRunner.run(variables);
+
+            Object document = Configuration.defaultConfiguration().jsonProvider().parse(result);
+
+            List<String> actionIds = JsonPath.read(document, "$.data.viewer.editingContext.representation.description.contextMenu[*].id");
+            assertThat(actionIds).isNotEmpty();
+            // The update action doesn't have a label, it only contains an ID used by the frontend.
```
sbegaudeau: It could still have a label, it's does not create any issue.

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const rows: GQLLibraryVersion[] = state.data?.viewer.library.versions.edges.map((edge) => edge.node) || [];
+  const count: number = state.data?.viewer.library.versions.pageInfo.count ?? 0;
+
+  const { data } = useLibraryVersions(namespace, name, version, pagination.pageIndex, pagination.pageSize);
+
+  useEffect(() => {
+    if (data) {
+      setState((prevState) => ({ ...prevState, data }));
+    }
+  }, [data]);
```
sbegaudeau: Like I said in your previous PR, you do not need to store the result of a query/mutation in a state

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { useState, useEffect, useMemo } from 'react';
+import { GQLLibrary } from '../../views/library-browser/useLibraries.types';
+import { UpdateLibraryTableProps, UpdateLibraryTableState } from './UpdateLibraryTable.types';
+import { useLibraryVersions } from './useLibraryVersions';
+import { GQLLibraryVersion } from './useLibraryVersions.types';
+
+export const UpdateLibraryTable = ({ namespace, name, version, onSelectionChange }: UpdateLibraryTableProps) => {
+  const [state, setState] = useState<UpdateLibraryTableState>({
+    data: null,
+  });
```
sbegaudeau: Can be removed

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+      ...prevState,
+      selectedLibraries: selection,
+    }));
+  };
+
+  const { project } = useCurrentProject();
+
+  const handleUpdateLibrary = () => {
+    if (state.selectedLibraries.length > 0) {
+      updateLibrary(project.currentEditingContext.id, state.selectedLibraries[0]);
```
sbegaudeau: Why pick the first one if everything lets you select multiple ones?

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useCurrentProject } from '../../views/edit-project/useCurrentProject';
+
+export const UpdateLibraryModal = ({ open, title, namespace, name, version, onClose }: UpdateLibraryModalProps) => {
+  const [state, setState] = useState<UpdateLibraryModalState>({
+    selectedLibraries: [],
+  });
+
+  const { updateLibrary } = useUpdateLibrary();
+
+  const onSelectionChange = (selection) => {
```
sbegaudeau: Let's not use the concept of selection in the naming convention to reference anything else than the workbench selection.

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryTable.types.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface UpdateLibraryTableProps {
+  namespace: string;
+  name: string;
+  version: string;
+  onSelectionChange(selection: string[]);
+}
+
+export interface UpdateLibraryTableState {
+  data: GQLGetLibraryVersionsQueryData | null;
+}
```
sbegaudeau: Can be removed

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/useLibraryVersions.types.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export interface GQLViewer {
+  library: GQLLibrary | null;
+}
+
+export interface GQLLibrary {
+  namespace: string;
+  name: string;
+  version: string;
+  versions: GQLLibraryVersionConnection;
```
sbegaudeau: `LibraryVersionsConnection` funnily enough, you have not kept the (invalid) name from the backend GraphQL type

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/UpdateLibraryTreeItemContextMenuContribution.tsx
```diff
@@ -0,0 +1,68 @@
...
+    ref: React.ForwardedRef<HTMLLIElement>
+  ) => {
+    const [state, setState] = useState({
+      open: false,
+    });
+
+    let fragment = null;
+
+    const { data, loading } = useObject(editingContextId, item.id);
+    if (!loading) {
```
sbegaudeau: can be removed, you don't need to test both loading and data if you are only using data

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/UpdateLibraryTreeItemContextMenuContribution.tsx
```diff
@@ -0,0 +1,68 @@
...
+                  <RefreshIcon fontSize="small" />
+                </ListItemIcon>
+                <ListItemText primary="Update Library" />
+              </MenuItem>
+              <UpdateLibraryModal
+                open={state.open}
+                namespace={library.namespace}
+                name={library.name}
+                version={library.version}
+                title={`Update ${item.label.styledStringFragments.map((fragment) => fragment.text).join('')}`}
```
sbegaudeau: ?

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/UpdateLibraryTreeItemContextMenuContribution.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { TreeItemContextMenuComponentProps } from '@eclipse-sirius/sirius-components-trees';
+import ListItemIcon from '@mui/material/ListItemIcon';
+import ListItemText from '@mui/material/ListItemText';
+import MenuItem from '@mui/material/MenuItem';
+import { Fragment, forwardRef, useState } from 'react';
+import RefreshIcon from '@mui/icons-material/Refresh';
+import { UpdateLibraryModal } from '../../modals/update-library/UpdateLibraryModal';
+import { useObject } from './useObject';
+
+export const UpdateLibraryTreeItemContextMenuContribution = forwardRef(
```
sbegaudeau: The context menu contribution have changed location

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/useObject.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+import { gql, useQuery } from '@apollo/client';
+import { useMultiToast } from '@eclipse-sirius/sirius-components-core';
+import { useEffect } from 'react';
+import { GQLGetObjectQueryData, GQLGetObjectQueryVariables, UseObjectValue } from './useObject.types';
+
+const getObjectLibrary = gql`
```
sbegaudeau: Can be moved with the context menu contribution

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/useObject.types.ts
```diff
@@ -0,0 +1,45 @@
...
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+export interface UseObjectValue {
```
sbegaudeau: Same comment

sbegaudeau (changes requested):
I'll rebase the PR, fix most comments and ensure that it gets merged. It would be easier with smaller contributions.

CHANGELOG.adoc
```diff
@@ -104,6 +104,9 @@ The publication of a studio that references elements from imported libraries now
 +
 image:doc/screenshots/duplicationDialog.png[DuplicationDialog, 70%]
 - https://github.com/eclipse-sirius/sirius-web/issues/4383[#4383] [sirius-web] Add support for representation duplication from explorer
+https://github.com/eclipse-sirius/sirius-web/issues/4694[#4694] [sirius-web] Add a contextual menu action to update imported libraries
+The action is visible on libraries that are first-level dependencies of the project (it is not visible on transitive dependencies).
+The implementation of the library update mechanism is generic and works on any project containing EMF resources.
```
sbegaudeau, line 124: I'll have to look into this but it seems odd for me that we are deciding how to perform the upgrade.

packages/core/backend/sirius-components-collaborative/src/main/java/org/eclipse/sirius/components/collaborative/dto/EditingContextObjectPayload.java
sbegaudeau: Done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/controllers/ObjectLibraryDataFetcher.java
sbegaudeau: Done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/explorer/services/ExplorerTreeItemContextMenuEntryProvider.java
sbegaudeau: Done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/views/explorer/services/ExplorerTreeItemContextMenuEntryProvider.java
sbegaudeau: Done, I've reduce both the number of request and the data loaded

packages/sirius-web/backend/sirius-web-application/src/main/resources/schema/libraries.graphqls
sbegaudeau: Done

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/trees/TreeItemContextMenuControllerTests.java
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/useLibraryVersions.types.ts
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryTable.types.ts
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryTable.tsx
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryTable.tsx
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+import Chip from '@mui/material/Chip';
+import {
+  MaterialReactTable,
+  MRT_ColumnDef,
+  MRT_PaginationState,
+  MRT_RowSelectionState,
+  useMaterialReactTable,
+} from 'material-react-table';
+import { useState, useEffect, useMemo } from 'react';
+import { GQLLibrary } from '../../views/library-browser/useLibraries.types';
```
sbegaudeau: Don't mix types from various hooks

sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/useObject.types.ts
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/useObject.ts
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/UpdateLibraryTreeItemContextMenuContribution.tsx
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/UpdateLibraryTreeItemContextMenuContribution.tsx
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/useUpdateLibrary.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export const useUpdateLibrary = (): UseUpdateLibraryValue => {
+  const { addErrorMessage, addMessages } = useMultiToast();
+  const [performUpdateLibrary, { loading, data }] = useMutation<
+    GQLUpdateLibraryMutationData,
+    GQLUpdateLibraryMutationVariables
+  >(updateLibraryMutation, {
+    // We have to use onComplete/onError here because the modal that triggered the update
+    // could have been closed by the refresh of the explorer (the element on which the contextual
+    // menu was opened does not exist anymore)
```
sbegaudeau: Updating a library one by one on the explorer feels really brittle. Given the constraints of transitive dependencies, I don't see this ever working for anything more than a very simple example. Even in terms of lifecycle it feels odd.

Let's say that I have a project A depending on B and C which are both depending on D. How could I ever upgrade anything with an approach where we have to consider upgrades one by one?

To be ever more funny, let's now consider the following versions with their compatible dependencies:

| D | B  | C  |
|---|---|---|
| 1.0.0  | 1.0.0  | 1.0.0  |
| 1.0.1  |   | 1.0.1  |
| 1.0.2  | 1.0.1  |   |
| 1.0.3  | 1.0.2  |   |
| 1.0.4  | 1.0.3  | 1.0.2  |
| 1.0.5  |   | 1.0.3  |
| 1.0.6  | 1.0.4  | 1.0.4  |

You can't upgrade to the version 1.0.1 or 1.0.2 of B because you don't have a matching version of C. You can't either upgrade to the 1.0.1 or 1.0.3 of C because you don't have a matching version of B and you can't upgrade anything individually at all. You can only switch between those three steps:

- B@1.0.0 and C@1.0.0 and D@1.0.0
- B@1.0.3 and C@1.0.2 and D@1.0.4
- B@1.0.4 and C@1.0.4 and D@1.0.6

And I'm only considering two dependencies... Given the massive state of the art with projects like maven, npm, or p2, there are very serious gaps in capabilities in this whole library lifecycle.

sbegaudeau: So while the comment indicates that we have to do this because of an issue in the lifecycle of the application, the real issue is that the lifecycle of the feature (where one would update dependencies one by one from the tree items of the explorer) cannot possibly work in a real use case.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/dto/UpdateLibraryInput.java
sbegaudeau: So you have all the information you need for you behavior but it cannot work beyond very basic examples.

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryModal.tsx
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/UpdateLibraryModal.tsx
sbegaudeau: Done

packages/sirius-web/frontend/sirius-web-application/src/modals/update-library/useUpdateLibrary.ts
gdaniel: A few comments:

With the current implementation, you can update B to version 1.0.1, this will update D to version 1.0.2, and update the references to D in C. This means that C will effectively depend on D@1.0.2. Updating a library will update its dependencies to the correct versions, it is just not possible to update D directly.

This last point is similar to maven dependencies: you control the versions of the dependencies you explicitly declare. If you want to force a version of a transitive dependency, you have to do it by adding a direct dependency to the artifact you want to force.

This being said, once you re-open the project you have an issue, because D will be loaded twice (D@1.0.0 from the dependencies of C, and D@1.0.2 from the dependencies of B). Since we cannot change the dependencies of libraries, this is the expected behavior.

What we are missing here is a dependency mediation mechanism to choose which version to use in the project. Maven does it transparently by choosing the "nearest definition" in the dependency tree. But as I said above, I don't know if this is something we want in a vacuum. Multiple versions of the same library could also live together in the same project (which is not true of maven dependencies), and in this case the current implementation of the editing context loader works. 

So I see one issue here: when we update B to version 1.0.1 we shouldn't update the references to D in C. We should load D@1.0.2 from the dependencies of B, and keep D@1.0.0 for C. 

We can later add a dependency mediation mechanism, which could (should) be configurable by the specifier, the same way you pointed out that the update mechanism probably needs an additional parameter and not be a default implementation. This has some implications, e.g. ranges of dependencies supported by a library.

gdaniel: Regarding the update of libraries via the contextual menu:
- I can get behind the fact that this is not the best user experience, I didn't think much about it to be honest, I followed the figma we had for this feature.
- I don't think this emphasizes a lifecycle issue though. There is nothing that prevents you from updating your maven dependencies one by one, and reloading them between each update. This is not the typical way to do it of course, but it can be done. If your project supports multiple versions of the same dependency OR if there is a dependency mediation mechanism, this shouldn't be an issue, and each intermediate reloading should produce a valid project.

gdaniel: Note that not having a dependency mediation mechanism will lead us to load libraries that contains objects with the same ID. This is an issue we already have with the import, which can import multiple versions of the same library.

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/UpdateLibraryTreeItemContextMenuContribution.tsx
gdaniel: Oops, we can use `library.name` in the title, this map thing comes from an older version where we didn't have the library at this point

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/UpdateLibraryHandler.java
gdaniel: Now that you point this out, I suspect it would be better to pass the original library to update in the input, in addition to the ID of the new library to use.

sbegaudeau (approved):

PR #4843, opened 2025-04-16
[4308] Handles now take bending points into account

### LINKED ISSUE #4308: Edge handles do not take bending points into account
image
image

Handles should use first and last bending point to find the best position and not the source/target node

frouene:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/4308

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/edge/EdgeLayout.types.ts
+++ b/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/edge/EdgeLayout.types.ts
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023, 2024 Obeo.
+ * Copyright (c) 2023, 2025 Obeo.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
@@ -47,7 +47,8 @@ export type GetEdgeParameters = (
   source: InternalNode<Node<NodeData>>,
   target: InternalNode<Node<NodeData>>,
   nodeLookup: NodeLookup<InternalNode<Node<NodeData>>>,
-  layoutDirection: string
+  layoutDirection: string,
+  bendingPoints: XYPosition[]
 ) => EdgeParameters;
 
 export interface EdgeParameters {
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -28,7 +28,7 @@
 
 === Improvements
 
-
+- https://github.com/eclipse-sirius/sirius-web/issues/4308[#4308] [diagram] Handles now take bending points into account
 
 
 == v2025.4.0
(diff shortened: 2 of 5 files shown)


packages/diagrams/frontend/sirius-components-diagrams/src/renderer/edge/EdgeLayout.types.ts
```diff
@@ -47,7 +47,8 @@ export type GetEdgeParameters = (
   source: InternalNode<Node<NodeData>>,
   target: InternalNode<Node<NodeData>>,
   nodeLookup: NodeLookup<InternalNode<Node<NodeData>>>,
-  layoutDirection: string
+  layoutDirection: string,
+  bendingPoints: XYPosition[] | null
```
sbegaudeau: Is there a point in accepting `null` instead of an empty array? I don't see any

sbegaudeau (commented):
From the description of the bug, Im not entirely sure what issue does this fixes, can you explain to me the original issue?

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/edge/EdgeLayout.types.ts
frouene: Yes, you're right, in both case we want to use the source and target node as references.

frouene:
The problem was that without this enhancement, it's always the position of the source and target nodes that are taken into account to determine where handles should be positioned.
This can lead to inconsistencies like this one.
without-handle-pos

When you have a custom edge path, the position of the last (or first) bendpoint is now used.

with-handle-pos

sbegaudeau:
It may be an improvement but it is still a bit odd that the edge can cross the node to go to the other side. Maybe there's another PR for this but I don't understand why if I move my node, it goes from this

Capture d’écran 2025-04-23 à 10 39 12

to this

Capture d’écran 2025-04-23 à 10 39 20

Why isn't the handle on the top in the second screenshot? If I move it a little more to the right, the handle can move so why can't it move to the top?

Capture d’écran 2025-04-23 à 10 40 33

sbegaudeau:
By the way, you should move your CHANGELOG entry to 2025.6.0 now :)

frouene:
The bendpoints are fixed by the user, I don't think this algo should moved them.
If you want what is in your last image, you just have to move the bendpoint.

frouene:
> By the way, you should move your CHANGELOG entry to 2025.6.0 now :)

Done

sbegaudeau:
> The bendpoints are fixed by the user, I don't think this algo should moved them. If you want what is in your last image, you just have to move the bendpoint.

I don't think that the users will care about those details, if they move the node in some cases, the edge is moved around the node, in other cases it goes through the node, that seems like a bug. I'm not against merging it along with the other improvements but the final result will still not be considered perfect. It is just improving a small subset of the issue we have.

sbegaudeau (approved):

PR #4845, opened 2025-04-17
[fix] Fix an issue where treem items could wrongly stay rendered as 'hovered'

pcdavid:
Before:

https://github.com/user-attachments/assets/a9eb4cae-c136-45ef-870f-39725a432d3c

After:

https://github.com/user-attachments/assets/d0490dc3-935e-4ac1-bb30-aedb05334d04


frouene (approved):

sbegaudeau (approved):
There are still some use cases which create issues (for example while dragging a tree item) but this seems like a different issue than this one.

PR #4906, opened 2025-05-05
[doc] Add ADR for configurable render variables to TreeDescription

pcdavid:

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -14,7 +14,7 @@
 - [ADR-197] Add support for backend-driven diagram customization
 - [ADR-198] Add the ability to share a URL containing the views of the workbench
 - [ADR-199] Add the ability to control the capabilities of the frontend from the backend
-
+- [ADR-200] Add configurable render variables to TreeDescription
 
 === Deprecation warning
 
--- a/doc/adrs/200_add_configurable_render_variables_to_tree.adoc
+++ b/doc/adrs/200_add_configurable_render_variables_to_tree.adoc
@@ -1,0 +1,90 @@
+= ADR-200 - Add configurable render variables to TreeDescription
+
+== Context
+
+When rendering a `Tree` on the backend (e.g. the Explorer), each property of each individual tree item is computed using a separate provider.
+For exemple, the icon URL for an item is computed by evaluating `iconURLsProvider`, whether an item is editable uses `editableProvider`, whether it has children by `hasChildrenProvider`, etc.
+
+Each of these is only provided with the default/generic variables provided by `org.eclipse.sirius.components.trees.renderer.TreeRenderer`.
+For properties which depend on more global state, or on costly information, each individual provider of each tree item will have to "start from scratch" with just the default variables.
+Any global information a provider needs to compute, or any computation that other providers (sometimes even providers on the same tree item) will need to recompute it.
+
+This can have severe performance impacts, see for example https://github.com/eclipse-sirius/sirius-web/issues/4102 where every leaf tree item has to make a separate query to the database *on every render*.
+
+== Decision
+
+We will provide the ability for a `org.eclipse.sirius.components.trees.description.TreeDescription` to customize (including extends) the variables used to render the tree.
+
+For this we will introduce a new interface:
+
+[source,java]
+----
+public interface IRepresentationRenderVariableCustomizer {
+    VariableManager customize(IRepresentationDescription representationDescription, VariableManager variableManager);
+}
+----
+
+Implementations of this interface can customize the variables available during the refresh/render of a representation.
+For example this can be used to *add* new custom variables that will be computed only once per render, and leveraged many times inside the various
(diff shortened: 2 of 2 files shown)


doc/adrs/192_add_configurable_render_variables_to_tree.adoc
```diff
@@ -0,0 +1,114 @@
...
+}
+----
+
+[NOTE]
+====
+This proposed signature, `Function<VariableManager, VariableManager>`, may not be the best choice.
+It's good enough to show the idea and implement a a proof of concept, but can probably be improved.
+====
+
+In `TreeRenderer`, this provider will be executed _once_, before rendering the root elements (and recursively, the tree items).
```
sbegaudeau: Since the `TreeRenderer`, like other renderers / top level components are being given the variable manager, it could also be done before the `TreeRenderer`, I'm not sure which approach would be better

doc/adrs/192_add_configurable_render_variables_to_tree.adoc
```diff
@@ -0,0 +1,114 @@
...
+
+As long as they do not conflict with system-provided variables, the new variables returned will be added to the context (VariableManager) used for the main part of the tree rendering:
+
+[source,java]
+----
+// Compute any custom render variables and merge them (if they do not conflict with the default ones)
```
sbegaudeau: On the other hand I would 100% let people do anything they want and just reuse the newly provided variable manager. I want to let people change the value of `self`(because it has been requested one way or another a couple of time).

doc/adrs/192_add_configurable_render_variables_to_tree.adoc
```diff
@@ -0,0 +1,114 @@
...
+
+=== Rabbit holes
+
+== Status
+
+Proposed
+
+== Consequences
+
+* If we implement this in `Tree` (notably to improve the _Explorer_) and it proves useful, the same mechanism could be generalized to all representations.
```
sbegaudeau: After some digging I remember that we did it this way for forms already and it was a bit limited

doc/adrs/192_add_configurable_render_variables_to_tree.adoc
```diff
@@ -0,0 +1,114 @@
...
+* It could probably also be exposed at the View DSL level, with the ability for studio makers to define their own custom render variables using AQL expressions, and then leverage these in the rest of their expressions.
```
sbegaudeau: :+1:, in form it already is

doc/adrs/192_add_configurable_render_variables_to_tree.adoc
sbegaudeau: Technically this improvement gives us the ability to change / extend the variables that will be used at the top. I've looked into this and we already tried this approach in forms, see `FormDescription#getVariableManagerInitializer` which has the same signature but it ended up a bit limited for some use cases since it did not allow for enough customization (it's tightly coupled, it allows only one provider).

This task could instead help build a more generic/powerful API to let people customize the initial variables used by the representation with something like this:

```
public interface IRepresentationRenderVariableCustomizer {
    VariableManager customize(IRepresentationDescription representationDescription, VariableManager variableManager);
}
```

And we could leverage it in all the `IXxxRepresentationCreationService` (which are being organized more properly on top of the work for the graphical undo/redo). This would give us the ability to remove the current state with both `TreeService` and `TreeCreationService` which have different variables for persisted and not persisted tree representations (think persisted tree representation vs explorer, reference widget, selection dialog) to have something like this:

```
public interface ITreeCreationService {
  Tree create(TreeCreationParameters parameters);
}
```

with

```
@Service
public class TreeCreationService implements ITreeCreationService {

  private final List<IRepresentationRenderVariableCustomizer> renderVariableCustomizers;

  @Override
  public Tree create(TreeCreationParameters parameters) {
    var variableManager = new VariableManager();
    // setup the variables by default thanks to the parameters

    for (var renderVariableCustomizer: this.renderVariableCustomizers) {
      variableManager = renderVariableCustomizer.customize(parameters.getDescription(), variableManager);
    }

    TreeRenderer treeRenderer = new TreeRenderer(variableManager, treeCreationParameters.getTreeDescription());
    return treeRenderer.render();
  }
}
```

It would thus give the ability to a Sirius Web extension to add, remove, or even change variables which can be especially useful given that downstream applications can change the content of the tree description of the explorer for example. We will not be the only one who want additional variables.

sbegaudeau (changes requested):

sbegaudeau (approved):

PR #4926, opened 2025-05-14
[doc] Add ADR on manage visibility actions API

mcharfadi:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -10,6 +10,7 @@
 
 - [ADR-192] Improve table representation lifecycle
 - [ADR-193] Leverage React router for nested routes
+- [ADR-194] Add the ability to contribute new actions to manage the visibility
 
 === Deprecation warning
 
--- a/doc/adrs/194_add_the_ability_to_contribute_new_actions_to_manage_the_visibility.adoc
+++ b/doc/adrs/194_add_the_ability_to_contribute_new_actions_to_manage_the_visibility.adoc
@@ -1,0 +1,86 @@
+= ADR-194 - Add the ability to contribute new actions to manage the visibility
+
+== Context
+
+A node action is used to open a modal in order to manage the visibility of the children of a node.
+
+
+== Decision
+
+We will add an API for retreiving and executing actions from the backend.
+
+
+=== Frontend
+
+We will add 2 hooks that will be used to execute the needed query and mutation.
+
+[code,typescript]
+----
+const { manageVisibilityActions } = useManageVisibilityActions(diagramElementId);
+const { invokeManageVisibilityAction } = useInvokeManageVisibilityAction(diagramElementId);
+----
+
+=== GraphQL shema
+
+We will provide a new type to the GraphQL schema to manage the visibility.
+
+----
+type ManageVisibilityAction {
+  id: ID!
+  label: String!
+}
+----
+
+We will be able to retrieve them thanks to the diagram description type
+
+----
+type DiagramDescription implements RepresentationDescription {
+  manageVisibilityActions(diagramElementId: ID!): [ManageVisibilityAction!]!
+}
+----
+
+We will add one mutation to invoke the action:
+
+----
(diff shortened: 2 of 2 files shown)


doc/adrs/193_add_menu_extension_on_manage_visibility_modal.adoc
```diff
@@ -0,0 +1,69 @@
+= ADR-193 - Add menu extension on manage visibility modal
+
+== Context
+
+A node action is used to open a modal in order to manage the visibility of the children of a node.
+
+== Decision
+
+We will add an extension point in order to add menu item to the menu of the modal.
```
sbegaudeau: While we want to add new item to the menu, I'm not sure that a frontend extension point is require or relevant. Here is a use case that we want to consider:

Let's consider a domain diagram as an example (we won't actually implement that but it gives a proper example to think about), I want to add a new entry in the management of the visibility that only hides attributes of an entity if they have the type boolean and a name starting with `is`. I thus need access to the semantic data to compute the list of elements to be hidden, I can't possibly know it from the structure of the diagram nor know it beforehand.

Why a frontend extension point? Why would I need to have a custom checkbox with a custom mutation and some custom backend behavior if I could just instead contribute some custom services on the backend and have the frontend perform a query to retrieve the custom actions?

mcharfadi: Okay so I'll need something like a `IManageVisibilityMenuActionsProvider` to populate the list of actions availables and a `IManageVisibilityMenuActionsHandler` to execute the action.

mcharfadi: I updated the ADR with a different approach

mcharfadi:
With this implementation we will have 
```
extend type Mutation {
 invokeAction(input: InvokeActionInput!): InvokeActionPayload!
 invokeManageVisibilityAction(input: InvokeManageVisibilityActionInput!): InvokeManageVisibilityActionPayload!
 }
```
and

```
type DiagramDescription implements RepresentationDescription {
  id: ID!
...
  palette(diagramElementId: ID!): Palette!
  actions(diagramElementId: ID!): [Action!]!
  manageVisibilityActions(diagramElementId: ID!): [Action!]!
}
```
Since `manageVisibilityActions `and `actions `are behaving in the same way, I wonder if we could just add an `actionType `like this 

```
  palette(diagramElementId: ID!): Palette!
  actions(actionType : String!, diagramElementId: ID!): [Action!]!
```
`actionType ` could be `manageVisibility` or  `nodeAction`, it could also be an enum.
Then the handler would behave the same way.
It would allow us to create new action easily and call them from any part of the UI.

doc/adrs/193_add_manage_visibility_actions_api.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+We will reuse the existing Action type.
+
+----
+type Action {
+  id: ID!
+  tooltip: String!
+  iconURLs: [String!]!
+}
+----
```
sbegaudeau: Don't reuse anything, you would create some coupling between two distinct features:

- A list item used specifically to update the visibility of some diagram elements
- A button visible on over on a node

Both feature may not evolve in the same direction, do not create this kind of coupling

sbegaudeau: You don't really need a tooltip or an iconURLs for example but you are required to have a label. Never create useless coupling.

sbegaudeau (changes requested):

sbegaudeau (approved):
I've updated the ADR to remove the useless coupling

Issue #4950, opened 2025-05-20
Add loading indicator for project/document download


flatombe:
* [X] **I have checked that this feature has not yet been suggested by someone else.**

Follow-up to #4935

When downloading a project/document (which can take a bit of time for larger projects/documents) there is no user feedback in our application (the browser tab does show loading but that's outside of the application). This can be confusing, especially for possibly large items.

### Expected behavior

There should be a loading indicator that helps the end user understand the operation is ongoing.

A shape will be used to refine how to implement this loading indicator.

PR #4953, opened 2025-05-21
[doc] Add shape for loading indicators for project/document downloads

flatombe:
Shape for #4950.

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [X] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

### DIFF
--- a/doc/iterations/2025.4/add_loading_indicator_for_downloads.adoc
+++ b/doc/iterations/2025.4/add_loading_indicator_for_downloads.adoc
@@ -1,0 +1,41 @@
+= (S-M) Add loading indicator for downloads
+
+== Problem
+
+Projects can be downloaded through the contextual menu of the `/projects` page and of the `/project/$projectId/edit` page.
+Documents can be downloaded through the contextual menu in the Explorer view.
+
+These operations can take a bit of time as the backend produces the corresponding file (zip file for projects, serialized file for the document, etc.), especially on larger projects and documents.
+
+During these operations, the user has no feedback that something is actually going on, resulting in the possibility of them getting confused (has an issue occurred?) and/or leaving the page or re-trying the operation.
+
+== Key Result
+
+There should be a loading indicator for these operations.
+
+=== Acceptance Criteria
+
+* The loading indicator starts when the operation gets triggered by the user.
+* The loading indicator ends when the bulk of the operation is finished and the user can interact with the application to decide what to do with the downloaded file.
+
+== Solution
+
+* The loading indicator used will be a Material UI CircularProgress.
+* It will replace the icon of the ongoing action upon selection of the action.
+* We will rely on the completion of the resource fetching (e.g. the call to `https://sirius-web.staging.factory.obeo.fr/api/projects/$projectId` has finished) to terminate the loading indicator.
+
+=== Breadboarding
+
+
+
+=== Cutting backs
+
+* If possible we want to keep the action as an href to the corresponding API call (e.g. `https://sirius-web.staging.factory.obeo.fr/api/projects/$projectId`).
+
+== Rabbit holes
+
+
+
+== No-gos
+
+
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -4,7 +4,7 @@
 
 === Shapes
 
-
+- Add loading indicator for downloads
 
 === Architectural decision records
 


doc/iterations/2025.4/add_loading_indicator_for_downloads.adoc
```diff
@@ -0,0 +1,41 @@
+= (S-M) Add loading indicator for downloads
+
+== Problem
+
+Projects can be downloaded through the contextual menu of the `/projects` page and of the `/project/$projectId/edit` page.
+Documents can be downloaded through the contextual menu in the Explorer view.
+
+These operations can take a bit of time as the backend produces the corresponding file (zip file for projects, serialized file for the document, etc.), especially on larger projects and documents.
+
+During these operations, the user has no feedback that something is actually going on, resulting in the possibility of them getting confused (has an issue occurred?) and/or leaving the page or re-trying the operation.
```
sbegaudeau, line 10: Technically since it's a regular HTTP download, they have the feedback from their web browsers

doc/iterations/2025.4/add_loading_indicator_for_downloads.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+== Key Result
+
+There should be a loading indicator for these operations.
+
+=== Acceptance Criteria
+
+* The loading indicator starts when the operation gets triggered by the user.
+* The loading indicator ends when the bulk of the operation is finished and the user can interact with the application to decide what to do with the downloaded file.
```
sbegaudeau, line 19: I'm not against moving to such an approach which is adopted by various applications but there are some things to consider.

As I said, the time to create the archive may not be that important compared to downloading it so the benefit may not be that important. Do you have some numbers on a big project on how much time is taken waiting for the zip to be created in the various use cases that you have described?

We would also need to switch the approach from a stateless one (you ask for a zip to download, we create and give you one) to a stateful one (where we keep this zip somewhere until the end user decide to download it). This can be very complex and/or expensive to setup. In a regular cloud based application, we would just push that to something like AWS S3 and be done with it but here it creates tons of new questions like:

- Where should this zip be stored while we wait for the user to download it?
- What happens if the end user does not download it?

So right now:

- There's already the feedback from the web browser
- I don't know how much time of the download it would improve with a "better user experience". If it takes 1s to create the zip and 10s to download it, it feels useless. If it's the over way around, sure we can start talking about this but in this case I'm confident that just improving the code used to create the zip would be a better alternative.
- It can be very complex technically

Technically this can be a mine field.

sbegaudeau (changes requested):

doc/iterations/2025.4/add_loading_indicator_for_downloads.adoc
flatombe: I agree that there is already the feedback from the web browser but we want to have a bit more feedback if it's doable without too much hassle.

I did a prototype to check out whether what I had in mind was actually possible, here's the result:

[SiriusWeb-LoadingIndicators-Proto-FLA.webm](https://github.com/user-attachments/assets/97a3f6d2-82e2-43f1-ab89-e107ed0883d0)

(note that the operation takes ~10s on my local instance while on staging it is ~3s)

This works by having the frontend, instead of offering an href:

1. Fetch the file while displaying the loading indicator
2. Create an on-the-fly href to the fetched zip file
3. Programmatically click the href

Required changes in the backend is minimal (expose header "Content-Disposition" so we can retrieve the file name).

What do you think?

sbegaudeau: That seems interesting but what happens if the end user tries to do anything during the download? For example, click on the project, etc.?

Since you have a prototype of the code, could you push it (even if it's some quick and dirty code) somewhere so I could have a look.

flatombe: For now, I think closing the menu does not halt the ongoing fetch.

The prototype is here: https://github.com/flatombe/sirius-web-dev/commit/b3995ce0146e411aceaf60004fd4f7d798d40bc9

sbegaudeau:
Unless I'm missing something, your code shows what I was worried about. You are downloading the entire zip in memory and then once it is downloaded, it opens a dialog to ask the end user how to save it in the file system.

- It's slower than the current solution
- If you moved to another part of the application or if you close the tab while it is downloading, you have lost everything

I've talked about it with Melanie and she is opposed to a change like this, I will thus close this pull request.

PR #4955, opened 2025-05-22
[doc] Add ADR for backend-driven diagram customization support

ntinsalhi-obeo:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [X] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
```diff
@@ -0,0 +1,226 @@
+= ADR-XXX - Add support for backend-driven diagram layout customization
+
+== Context
+
+In Sirius Web, diagram layouting (sizing and positioning of nodes and edges) is handled through a combination of backend and frontend processes.
+The backend’s `DiagramEventProcessor` manages diagram rendering and persistence, distinguishing two primary event causes:
+
+- `Refresh`: Triggered by actions such as creating new representations, creating/deleting semantic elements, edge creation, or synchronized representation creation.
```
sbegaudeau: I don't think that creating a new representation should refresh a diagram. For example I have a diagram opened and someone creates a form it should not trigger anything. The refresh cause is mainly triggered by:

- semantic change
- diagram element visibility change
- diagram element collapsing state change

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
```diff
@@ -0,0 +1,226 @@
...
+
+More precisely, the layout rules are:
+
+- The root Outcome (e.g., New Outcome1) must be placed at the top center of the diagram.
+- Its immediate child Outcomes (e.g., New Outcome2, New Outcome3, New Outcome4) should be laid one level below, maintaining a vertical orientation.
+- Any Opportunities linked to those Outcomes (e.g., New Opportunity1, New Opportunity2, etc.) must be placed at the same vertical level as the Outcomes they are derived from.
+- Each Opportunity has an integer priority property. The lower (1) the value, the more to the left it should appear in the diagram. This helps visually identify top-priority items from left to right.
+
+_Such a layout expresses both semantic relations and business rules.
+Making it explicit and declarative, or at least configurable, is essential for similar diagram types in other domains._
```
sbegaudeau: Pretty much 100% of everything you wrote until this point is the description of the state of the art and why your needs should make the application change. This functional analysis is what we describe in our shapes :)

Please take all this content and just extract it in a shape in the relevant folder `/doc/iterations/2025.6/`

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
```diff
@@ -0,0 +1,226 @@
...
+As detailed in the *Needs* section, algorithm customization requires access to diagram definitions and semantic data.
+
+=== [front-end] Extension point to customize ELK or incremental front-end algorithms
+- The diagram description would need to be loaded with the diagram.
+- Specific queries would be needed to load semantic information
+
+This solution is both complex to implement and would negatively impact the performance.
+Moreover, there is no guarantee of successfully customizing the ELK algorithm if we pursue this path.
+
+=== [back-end] Extension point to customize back-end rendering
```
sbegaudeau: Small nitpick but on the backend side, we are not really talking about extension points. Instead we have services or APIs.

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
```diff
@@ -0,0 +1,226 @@
...
+
+Both diagram description and semantic data are easily and quickly accessible.
+
+== Decision
+
+We will introduce a backend extension point in the `DiagramEventProcessor` to inject custom layout algorithms computed server-side, specific to diagram types or applied globally.
```
sbegaudeau: The `DiagramEventProcessor` is already in charge of tons of things and we have some awaiting contributions to split some of its responsibilities to smaller services, does it have to be there? Couldn't it be in `DiagramCreationService` for example?

We may need to split one part of `DiagramEventProcessor` first, for you to add your new API

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
```diff
@@ -0,0 +1,226 @@
...
+As explained further, this approach requires no frontend modifications, as the existing `useLayout` mechanism will apply on the backend-provided `DiagramLayoutData`.
+
+The custom layouts are invoked *automatically* for the following scenarios:
+
+* *Refresh Cause*: Creation of new representation, create/delete semantic elements, edge creation, synchronized representation creation.
+* *Layout Cause*: Click on "Arrange All", move or resize nodes.
+
+=== Extension Point for Layout Providers
+
+A new backend extension point is introduced using Spring beans to support domain-specific or diagram-type-specific layout strategies.
```
sbegaudeau: Technically it should work without Spring so I don't think it's necessary to mention Spring here

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * Indicates whether this provider can handle layout for the given diagram.
+     *
+     * @param diagram The diagram to process
+     * @param diagramDescription Optional diagram description
+     * @param diagramContext The current diagram context
+     * @return true if the provider can handle the diagram layout, false otherwise
+     */
+    boolean canHandle(Diagram diagram, Optional<DiagramDescription> diagramDescription, IDiagramContext diagramContext);
```
sbegaudeau: - Why would the diagram description be optional?
- Why would you need the diagram context?

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
sbegaudeau: You should also step back a little more from your very specific need. You need to change the layout of the diagram, sure but why should such an API be specific to a change in layout? It's just changing a diagram.

Why couldn't I use such an API to add new nodes to existing diagrams? Let's consider an extension to Sirius Web or SysON which would add, change, remove diagram elements using some custom logic. I think that in the "ADR" part of this PR, you should consider the bigger picture and thus not be too coupled with the layout. 

This issue of layout is your need, and thus part of the shape, the capabilities to be added to fulfill your needs (as described in an ADR) may not be coupled to the layout.

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
```diff
@@ -0,0 +1,226 @@
...
+     * Attempts to apply a custom layout using a registered provider if available.
+     *
+     * @param diagram The diagram to layout
+     * @param diagramDescription Optional diagram description
+     * @param diagramContext The diagram context
+     * @return The updated diagram (potentially with custom layout)
+     */
+    private Diagram applyCustomLayoutIfAvailable(Diagram diagram, Optional<DiagramDescription> diagramDescription, IDiagramContext diagramContext);
+----
+
```
sbegaudeau, line 103: You could make a simpler and more extensible API:

```
public interface IDiagramPostProcessor {
    Diagram postProcess(Diagram diagram, VariableManager variableManager);
}
```

and then use it like that:

```
var diagram = ...;
for (var diagramPostProcessor: this.diagramPostProcessors) {
    diagram = diagramPostProcessor.postProcess(diagram, variableManager);
}
return diagram;
```

It provides a simpler and more powerful API, it allows multiple customization to be applied for example. Would it give you access to anything you need?

sbegaudeau (changes requested):

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
sbegaudeau: I am not mentioning something like this without any reason, we had a similar need in the past for forms and thus built this API with `org.eclipse.sirius.components.collaborative.forms.api.IFormPostProcessor`. So I'm wondering if it's really necessary to deviate from the pattern adopted for forms and why.

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
ntinsalhi-obeo: When a new representation is created, a mutation for the `layoutDiagram` operation with the cause `refresh` is sent to the frontend, including a list of nodes and edges that correspond to the semantic elements associated with the new representation.

One could argue this qualifies as a semantic change, since it introduces the potential presence of nodes in the diagram — whether the representation is synchronized.

image

This makes it a relevant case in the context of applying custom layouts, as the layout engine may need to handle the appearance of newly added semantic elements, even if they result from another user's action.

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
ntinsalhi-obeo: The layout-related scenario I described was intended as a concrete example of how this API, that we wrongly called an extension point, could be useful. It highlights a current limitation in Sirius Web: the inability to enforce custom layouts, and mainly custom layouts based on the semantic nature of nodes.

This example felt appropriate because the need for semantic-aware layout is what initially motivated this contribution. Currently, the only layout customization available is through the `ArrangeLayoutDirection` enum (e.g., RIGHT, LEFT, UP, DOWN) — which applies globally and does not involve any semantic awareness. It’s also limited to "Arrange All" and doesn’t offer a mechanism for injecting custom node placements.

That being said, I do agree with your point: the extension API itself should not be limited to layout use cases. A custom layout logic is one specific need among potentially many, but the present need revolves around this custom layout injection.

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
ntinsalhi-obeo: Dont acte, thanks !

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
ntinsalhi-obeo: We considered placing the hook in the `DiagramEventProcessor` to be essential, based on our understanding of how layouting currently works in Sirius Web. This component is where the different scenarios that trigger layouting are distinguished, making it the most context-aware location to intercept and apply a custom layout strategy.

Placing the logic there ensures we can intercept layout-related events at the right time — particularly after the frontend has returned its initial coordinates. This allows us to override them with coordinates defined by a custom layout, tailored to semantic structures.

While we’re aware that this introduces a "blink" effect due to frontend rendering happening before our custom layout is applied, it is still preferable to a "ping-pong" effect — where layout updates bounce endlessly between frontend and backend.

ntinsalhi-obeo: That said, I understand your concern regarding `DiagramEventProcessor` becoming too overloaded. If part of it is planned to be extracted or restructured soon (e.g., into a dedicated layout or creation service), I’d be happy to rebase the contribution accordingly.
Just let me know if there's a specific service or scope already planned for this kind of logic, and I’ll align with it.

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
ntinsalhi-obeo: Noted — Spring was mentioned as it appeared to be the most straightforward approach to define the extension point.
However, if the same goal can be achieved with a simpler or more framework-agnostic solution, I'm fine with removing the reference to Spring.

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
ntinsalhi-obeo: These parameters are included to give the specifier greater autonomy and flexibility in their implementation by providing access to relevant contextual information:

- `diagramContext` grants access to the underlying Diagram object, which includes the target object ID. This ID is key to retrieving the semantic element associated with the representation. This is essential when determining whether a diagram is concerned by a specific custom layout.

- Additionally, `diagramContext` exposes node and edge descriptions, which are often useful in layout decisions.

As for `diagramDescription`, it is marked as Optional because its contents can be derived from `DiagramContext#getDiagram()`. It acts more as a convenience than a necessity. If present, it simplifies access to semantic and structural information — but it's not strictly required, hence the optionality.

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
ntinsalhi-obeo: Noted — this is indeed a valid alternative, especially if we take into account the previous suggestions mentioned above.

doc/adrs/XXX_add_support_for_backend-driven_diagram_layout_customization.adoc
ntinsalhi-obeo: - `diagramDescription` is optional because it is not directly available in the handle method of `DiagramEventProcessor`. Instead of resolving it manually—which could introduce performance loss or reduce flexibility—we allow the specifier to provide it explicitly when needed.

- `diagramContext` contains essential information, including the list of `DiagramEvent` impls, which helps identify what triggered the current diagram update (e.g., element visibility changes, edge modifications).

PR #4957, opened 2025-05-22
[4094] Improve explorer rendering performances

### LINKED ISSUE #4094: Hovering/leaving a tree item causes a re-render or the item and all its descendants
In `TreeItem.tsx` we modify the state of the TreeItem when the mouse pointer enters or leaves the item. This causes a re-render of the item and any descendent. When an explorer displays a significant number of elements, moving the mouse over the top-level ones which contain many childrens can feel very slow for something that should be immediate.

We should probably try to handle this purely in CSS if possible. It's not an immediate change as currently we use this flag in `TreeItemActionProps`, see for example `WidgetReferenceTreeItemAction` which uses it.

pcdavid:
This uses a different approach than the "full CSS" one envisioned initialy.

The "isHovered" state is, well, inherent part of the item's actual state as we need to pass it along to `treeItemActionRender` and `TreeItemAction` for example.

The new approach is simply to make `TreeItem` only responsible for rendering *itself* and not its children. This way it's not a performance issue anymore, and the time needed to re-render a given item is independent on the number of descendants it has in the tree.

The recursive part of the rendering is pulled up into `Tree.tsx` in a dedicated recursive component `TreeItemWithChildren`.

Before the change, with a large (~900 items) sub-tree expanded under "Eclipse EMF", simply hovering the mouse on the "Eclipse EMF" top item show a lag of about 1s, with all the 900+ items underneath re-rendered along the the "Eclipse EMF" item itself.

https://github.com/user-attachments/assets/cdd1a6a1-ac5e-4b45-aa4b-908b25be87bd

After the change, with an equivalent initial state, only the item hovered (in/out) is re-rendered, not its descendants. The UI is much more reactive.

https://github.com/user-attachments/assets/35146c69-9ee0-4f91-aa05-ce503c03d204

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -184,7 +184,8 @@ This changes the behavior if there are delegates which *override* the default re
 - https://github.com/eclipse-sirius/sirius-web/issues/4874[#4874] [diagram] Move View DSL node layout strategy from node description to node style.
 This modification will enable a conditional style with a different layout strategy.
 - https://github.com/eclipse-sirius/sirius-web/issues/667[#667] [sirius-web] Restore layout and unsynchronized nodes when importing a project with a diagram
-
+- https://github.com/eclipse-sirius/sirius-web/issues/4094[#4094] [sirius-web] Improve the reacivity of the explorer rendering when hovering on a tree with many items are expanded/visible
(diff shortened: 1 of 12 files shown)


sbegaudeau (approved):
This looks good, I'll wait for the build to merge it

Issue #4886, opened 2025-04-29
Navigation to other view/panel


jessymallet:
# Need

As a developer, I need a way to open a specific side panel with certain parameters from other parts of the workbench, especially with the search for a discussion using the omnibox.

# Example 

For example, if I add a specific view called discussion and use the omnibox search command to search a specific discussion, I need to open the discussion panel to display the specific discussion selected from the search results.

# Resolution track:

In the same way that we can now have a URL that indicates the selection, we could have an optional parameter in the URL that indicates which views/panels should be opened (default if nothing: Explorer on the left & Details on the right, but it will depend anyway of each application).

For passing parameter, another solution could be to pass state/values via a custom context. The command component on the frontend would set some state in the context and the discussions view component reacts to the hook to display the selected discussion

jessymallet:
The URL should also contain additional argument on opening view. 
In addition to the id of the view, URL should contain information specific to this view such as which filter has been activated on this view or which tab is opened.

frouene:
Another PR is still in progress on this issue.
https://github.com/eclipse-sirius/sirius-web/pull/5214

PR #4961, opened 2025-05-22
[doc] Add shape for sharing the workbench state

flatombe:
This is a first tentative shape for bug: https://github.com/eclipse-sirius/sirius-web/issues/4886

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [X] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management


doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+If at least 1 representation editor is opened, then the URL starts with: `/projects/:projectId/edit/:representationId` where `representationId` is the ID of the representation opened in the active editor.
+
+And since https://github.com/eclipse-sirius/sirius-web/issues/4376[#4376], the URL also contains the current selection of the workbench, meaning the URL can be of the form `/projects/:projectId/edit/?selection=id1%2Cid2` or `/projects/:projectId/edit/:representationId?selection=id1%2Cid2` (where `%2C` encodes `,`).
+
+Reversely, when resolving one of the above-mentioned URLs, the corresponding project, representation and selections are set.
+
+However, there are parts of the workbench state which are not encoded in the URL.
+
+These parts are the views opened in the left and right panels (e.g. in Sirius Web application, between `Explorer` and `Validation` on the left, and between `Details`, `Query`, `Representations` and `Related Elements` on the right).
```
sbegaudeau: Just a detail but the views available can be changed in downstream applications which may remove default views or even add new ones with different configurations. A downstream application can even change the position of a view (explorer on the right for example)

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,109 @@
...
+** The focused part of the workbench (one of the active views or the active editor)
+** The selection
+* When resolving a URL that specifies a workbench state, the matching workbench state is set up.
+
+== Solution
+
+* We will rework the URL to change based on the state of the workbench.
+* The base URL for an open project will be `/projects/:projectId/edit`.
+* The state of the workbench will be encoded in the URL through the following parameters:
+** `views`: will encode which views are active, e.g. `explorer%2Cdetails`.
```
sbegaudeau: You will need to encode the view that is active per panel, you can't assume that the explorer is on the left

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,109 @@
...
+** `editors`: will encode which editors are opened, through the ID of their input, e.g. `representationId1%2CrepresentationId2%2CrepresentationId3`. (Note: what's the real difference between views and editors in this context? Maybe we should merge views and editors as parts).
```
sbegaudeau: You could indeed encode those with just a set of id associated with the center panel

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,109 @@
...
+** `configurations`: will encode the configurations of the active views and editors, e.g. `explorer_state_closed%2Cexplorer_tree_activeFilterId_filterId1`.
```
sbegaudeau: It's a bit odd to store the configuration in another property, you could technically encode a configuration for a view which is not opened

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,109 @@
...
+** The active view on the right (and its configuration)
+** The focused part of the workbench (one of the active views or the active editor)
+** The selection
+* When resolving a URL that specifies a workbench state, the matching workbench state is set up.
+
+== Solution
+
+* We will rework the URL to change based on the state of the workbench.
+* The base URL for an open project will be `/projects/:projectId/edit`.
+* The state of the workbench will be encoded in the URL through the following parameters:
```
sbegaudeau: query parameters

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,109 @@
...
+** `editors`: will encode which editors are opened, through the ID of their input, e.g. `representationId1%2CrepresentationId2%2CrepresentationId3`. (Note: what's the real difference between views and editors in this context? Maybe we should merge views and editors as parts).
+** `configurations`: will encode the configurations of the active views and editors, e.g. `explorer_state_closed%2Cexplorer_tree_activeFilterId_filterId1`.
+** `selection`: will encode the workbench selection, just like it already does, e.g. `id1%2Cid2`.
+** `active`: will encode the active part, e.g. `representationId2`.  
+
+=== Examples
+
+image::images/add_workbench_views_to_url_example1.png[Example 1]
+
+URL: `/projects/projectId/edit/views=explorer%2Cdetails`
```
sbegaudeau: This is not using a query parameter, you should use: `/projects/projectId/edit?views=explorer%2Cdetails`. Something like this may be more universal and robust (let's not care about the character encoding for a minute):

```
/projects/projectId/edit?workbench={"panels":{"left":["explorer"],"center":[],"right":["explorer"]}}
```

I'll rather use a JSON object also to have a very simple way to encode/decode the state of the workbench. This way, we just have to talk about the type of this JSON object, something like this:

```
interface WorkbenchConfiguration {
  panels: WorkbenchPanelConfiguration;
}

interface WorkbenchPanelConfiguration {
  left: string[];
  center: string[];
  right: string[];
}
```

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,109 @@
...
+
+URL: `/projects/projectId/edit?views=explorer%2Cdetails&configurations=explorer_state_closed%2Cdetails_state_closed`
+
+* In parameter `configurations`:
+** `explorer_state_closed` encodes that the `Explorer` view is closed.
+** `details_state_closed` encodes that the `Details` view is closed.
+
+image::images/add_workbench_views_to_url_example3.png[Example 3]
+
+URL: `/projects/projectId/edit/views=explorer%2CrelatedElements&editors=topographyId%2Ctopography2Id&configurations=topographyId_zoomLevel_75%2CtopographyId_position_123_456&relatedElements_incoming%2CrelatedElements_current%2CrelatedElements_outgoing&selection=id1%2Cid2&active=topographyId`
```
sbegaudeau: See this part of the shape is almost unreadable, by encoding everything using one JSON object as a parameter, we can talk about it more efficiently. Since the system has to be extensible, it has to be understandable

sbegaudeau (changes requested):

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
flatombe: Noted, thanks.
I've reworked the text a bit to make it more obvious that it is the running application that defines which views are available, and which are used as default.

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
flatombe: Fixed

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
flatombe: Fixed, merged editors and views as 'parts'

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
flatombe: Fixed, merged configuration/state alongside the parts

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
flatombe: fixed

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
flatombe: You are right, I did not think of encoding a whole JSON object into the URL.
Reworked the shape to do that.

flatombe:
Stéphane, I have reworked the shape to rely on a JSON representation of the workbench state, which should the issues you've mentioned.

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,378 @@
...
+
+Being able to share the complete workbench state through the application URL would facilitate collaboration between users.
+
+== Key Result
+
+The URL should be synchronized with the full state of the workbench parts.
+
+This means that:
+* Our frontend components must make explicit their state/configuration that may be exposed, the associated default values, and offer APIs to conform to a configuration provided by the user through the URL.
+* The Sirius Web application must be able to retrieve the workbench state from its components, encode it into the URL, and reversely, be able to decode such a state from the URL in order to apply it to its constituting components.
```
sbegaudeau: Just to be clear, this does not have to be the case 100% of the time, we can have a mechanism to share a workbench which would then and only then create this complex URL. Given some of our recent work on performances, I strongly suspect that synchronizing everything in the URL at all time will have a sizable impact performance wise. On top of that it may be more complex than just doing it once. Some change deep down in a view could update the URL and thus trigger a re-render of everything in the workbench.

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,378 @@
...
+
+=== Acceptance Criteria
+
+* The URL is synchronized with the full state of the workbench, which includes:
```
sbegaudeau: We can share an URL that can be resolved to the full state of the workbench, which includes:

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,378 @@
...
+----
+
+where:
+** `WorkbenchState.focus: string`: the ID of the `WorkbenchPart` that has the user focus.
+** `WorkbenchPart.id: string`: the ID of the `WorkbenchPart` (e.g. `left`, `right`, `center`, `explorer`, `details`, etc.).
+** `WorkbenchPart.configuration: object`: the part-specific configuration encoded as an arbitrary JSON object.
+* The workbench state is encoded in and decoded from the URL through query parameter `workbench`, e.g. `/projects/projectId/edit?workbench=<encoded JSON object>`.
+* The URL will still encode the selection through query parameter `selection`, just like it already does, e.g. `/projects/projectId/edit?workbench=<encoded JSON object>&?selection=id1%2Cid2`.
+
+=== Examples
```
sbegaudeau: It's not an issue but all these examples are not necessary in a shape, that's the concern of the ADR, this shape could end up with 80 lines without issues. And in this examples part, you are making TONS of assumptions at once on what we can do and how much it can cost. This part could very well end up being separated in multiple ADRs and commits (or even new shapes in some cases).

I'll merge both the shape and this part extracted in a work in progress ADR.

doc/iterations/2025.4/encode_workbench_state_in_url.adoc
```diff
@@ -0,0 +1,378 @@
...
+
+
+=== Cutting backs
+
+N/A
+
+
+== Rabbit holes
+
+* The maximum size for a URL is usually around 2000 characters. Considering the size of UUIDs, the possibility through some representations to select many elements, and the possibility for workbench parts and their configurations to be arbitrarily deep and complex, we may need to consider having a mechanism to make sure we do not go overboard.
```
sbegaudeau: Given the size of your examples, sharing all the time the state of the workbench in the URL may not even possible in any real use case.

doc/adrs/198_add_the_ability_to_share_a_url_containing_the_views_of_the_workbench.adoc
```diff
@@ -0,0 +1,327 @@
+= ADR-198 - Add the ability to share a URL containing the views of the workbench
```
sbegaudeau, line 1: I've extracted all the technical part in this ADR. I'll leave you to rewrite it to start with an ADR to share the workbench parts active in the workbench. Be careful, this ADR will be one way or another in conflict with the ADR 195. I don't think I'll start working on implementing the ADR 195 soon so your contribution should come first.

After that a second ADR should highlight how we could expand your first work by adding the ability for a workbench part to contribute some configuration to the URL shared. I suspect that some additional mechanisms will be necessary for those details.

Just stick to the list of parts in each panel for this first ADR / contribution. I suspect that for this second part, you may need something like `useImperativeHandle` to let the parent component ask it child component some information regarding its state. It would give you the ability to do things like:

```
const explorerRefView = useRef(); // It will be interesting to see the type of this ref

const onShareButtonClick = () => {
  const configuration = explorerRefView.current.getConfiguration();
  // serialize it in the URL
};

<ExplorerView ref={explorerViewRef} />
```

(that's just one approach)

sbegaudeau (approved):

PR #4964, opened 2025-05-23
[4183] Use react resizable panels to layout the workbench

### LINKED ISSUE #4183: Resizing a panel re-renders the opposite one
### Steps to reproduce

1. Open a project with a significant amount of element (e.g. "Papaya Example").
2. Expand parts of the Explorer so that at least a few hundreds (or more) items are visible
3. Resize the "Details" view on the right

Even though changing the width of the "Details" view has no impact on the size or content of the "Explorer" view, the Explorer is re-rendered anyway, and in such a case where many elements are visible this can currently be very slow (here for absolutely no good reason).

Actually I think the whole workbench is re-rendered in such a case: both panels and the main representation area. I have not tested but I suppose even with nothing expanded in the Explorer but a large diagram visible, we re-render the diagram and get some slowness here too (this is at least partially understandable as the diagram is adjacent to the resized panel so the resize changes the diagram's viewport size, but in the case of a diagram (as opposed to a form) we can maybe avoid an actual re-render in this case?).

jerome-obeo:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/4183

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies
- [x] Are the new / upgraded dependencies mentioned in the relevant section of the `CHANGELOG.adoc`?

## Frontend

### General purpose
- [x] Is the code properly tested? (Plain old JavaScript tests for business code and tests based on React Testing Library for the components)

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


CHANGELOG.adoc
```diff
@@ -31,7 +31,7 @@ Of particular note: projects which create View-based, in-memory, representation
 [source,java]
 ----
 Map<String, Object> customOptions = Map.of(JsonResource.OPTION_FORCE_DEFAULT_REFERENCE_SERIALIZATION, Boolean.TRUE);
-JsonResource resource = jsonResourceFactory.createResource(uri, customOptions);
+JsonResource resource = jsonResourceFactory.createResource(uri, this.customOptions);
```
pcdavid: Why this change? In the example code, `customOptions` is a local variable defined on the previous line, not a field.

jerome-obeo: automatic change made by IntelliJ 🫤

packages/core/frontend/sirius-components-core/src/workbench/Site.types.ts
```diff
@@ -16,7 +16,5 @@ export interface SiteProps {
   editingContextId: string;
   readOnly: boolean;
   side: WorkbenchViewSide;
-  expanded: boolean;
-  toggleExpansion: () => void;
-  contributions: WorkbenchViewContribution[];
+  contribution: WorkbenchViewContribution | undefined;
```
sbegaudeau: - Prefer `null` to `undefined`
- Why would a site be limited to 0..1 contribution instead of 0..*? it's odd, it may be a good idea but I'm not sure that it would still be a site technically. Does the concept even make sense anymore with your changes? I'll have to look a bit more into this because it seems that the role of some concepts are changed while still keeping the concepts.

packages/core/frontend/sirius-components-core/src/workbench/Site.tsx
```diff
@@ -157,11 +68,5 @@ export const Site = ({ editingContextId, readOnly, side, expanded, toggleExpansi
...
   }
 
-  let classSite = side === 'left' ? classes.leftSite : classes.rightSite;
-  return (
-    <div className={classSite} data-testid={`site-${side}`}>
-      {viewSelector}
-      {selectedView}
-    </div>
-  );
+  return null;
```
sbegaudeau: The fact that a site can return null is odd, wouldn't that be a part instead of a site?

sbegaudeau (changes requested):
I'll have to play with it a bit to understand some parts of the changes and check the behavior

packages/core/frontend/sirius-components-core/src/workbench/Site.types.ts
jerome-obeo: The contribution is the current visible contribution among all contributions given to Sidebar which is now in charge of controlling the content of the panel.
Previously, Site contained both view selector and view, now Site is only containing the view.
Maybe this is misnamed now?

jerome-obeo: I have changed `undefined `into `null`

sbegaudeau (approved):

sbegaudeau:
I've tested how it behaves with regards to the re-rendering of the workbench and the difference is massive. It does not even re-render a panel when it is resized, so it does not re-render other panels too

https://www.youtube.com/watch?v=e3ED6hWQXRo

PR #5040, opened 2025-06-11
[5039] Decouple the impact analysis dialog from diagrams

### LINKED ISSUE #5039: Decouple the impact analysis dialog from the diagram

The impact analysis dialog is coupled to the diagrams, but the display of an impact analysis report could be made generic and not coupled to any representation.

This requires to split `showImpactAnalysisDialog` (from `useImpactAnalysisDialog`) in two:
- Get the impact analysis report (representation specific)
- Display the result (generic)

As a result, the files responsible for _displaying_ the impact analysis report could be moved to Sirius Component Core, allowing to reuse them in other representations.

This issue is linked to #4971

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/5039

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/impact-analysis/useImpactAnalysis.ts
```diff
@@ -47,43 +44,9 @@ const getImpactAnalysisReportQuery = gql`
...
-        variables === null,
-      variables: {
-        editingContextId,
-        representationId,
-        toolId,
-        diagramElementId,
-        variables,
-      },
-    }
+export const useInvokeImpactAnalysis = (): UseInvokeImpactAnalysisValue => {
```
gdaniel, line 49: We probably don't need a hook for that

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/impact-analysis/ImpactAnalysisDialogContext.tsx
```diff
@@ -30,43 +30,37 @@ export const ImpactAnalysisDialogContextProvider = ({ children }) => {
...
+      loading,
+      impactAnalysisReport: data?.viewer.editingContext.representation.description.diagramImpactAnalysisReport ?? null,
       onConfirm,
-      editingContextId,
-      representationId,
-      toolId,
-      toolLabel,
-      diagramElementId,
-      variables,
     });
```
gdaniel: This is the part I am skeptical about. 
The only way I have found to update the state when the query completes is to use `then`, which seems odd to me. I can't use `onCompleted`: it doesn't trigger, most likely because the palette is closed at that point (I tried to force the palette open and the `onCompleted` triggers).

I have the feeling there is a lifecycle issue here, and the way the query is created/passed to the context is not right, but I can't find the actual issue/how to solve it. Let me know if you have any lead.

Note that the actual implementation "works": it does what it is supposed to do, but I am quite sure it doesn't do it the right way.

gdaniel:
The PR is a draft: I still have to move things I can move in Sirius Components Core, do some cleanup, changelog, etc but I'd like some advice on the issue I have with the refactoring before proceeding with the rest of the work.

The underlying question is also whether this refactoring is useful: if the only way to make it work is to do some weird and complex stuff, are we better duplicating what we have in diagrams in trees?

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/impact-analysis/ImpactAnalysisDialogContext.tsx
```diff
@@ -30,43 +30,37 @@ export const ImpactAnalysisDialogContextProvider = ({ children }) => {
...
-    toolId: string,
+  const showImpactAnalysisDialog = <TVariables extends GQLInvokeImpactAnalysisVariables>(
+    query: LazyQueryResultTuple<GQLGetImpactAnalysisReportData, TVariables>,
+    queryParameters: TVariables,
     toolLabel: string,
-    diagramElementId: string,
-    variables: GQLToolVariable[],
     onConfirm: () => void
   ) => {
+    const [getImpactAnalysisReport, { loading, data }] = query;
```
sbegaudeau: In my opinion this would not have anything to do here. This context, use to "show an impact analysis dialog" should not be in charge of performing the query anymore. You are still coupled here with the query of the diagram, you have a lifecycle issue and the signature of the function showImpactAnalysisDialog starts to get quite complex.

This function should just show the dialog and nothing else, it should expect an impact analysis report and that's it. That's what I meant when I said that we should split the computation of the report and its display. This context should be contained by the workbench probably and the query to compute the report should be done by someone called by `useSingleClickTool`

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/impact-analysis/useImpactAnalysis.ts
sbegaudeau: You should add error management and everything as usual here, you would then need this hook as usual :)

sbegaudeau (changes requested):

sbegaudeau:
By the way, you could start using the newly available `cooldown` branch as a target for this PR since the master branch is currently frozen until the release

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/impact-analysis/ImpactAnalysisDialogContext.tsx
gdaniel: The split you mention was also my first thought, but I ran into the following issues:
- Since the palette closes as soon as the tool has been clicked, I can't send the query to get the impact analysis report _and get the result_. At least I didn't manage to do it: the query is sent, but when the response is available the palette has been closed and the received value is not propagated anywhere.
- The dialog displays a loading indicator while waiting for the query response. With the approach you describe this wouldn't be the case anymore, and the loading would be done between the click on the tool and the display of the dialog, which looks weird to me. Now that I think more about it I can maybe work around that by providing the report and whether the data is loading.

The solution I came up with was to still send the query in the context (this was we aren't bothered by the palette closing), but pass the query function to it. I agree with you, this solution is not good. I'll try some more things tomorrow and update the PR when I get an implementation more aligned with the original idea.

gdaniel: In any case I think that to work around the issue of the palette closing we will have to do something like the `invokeSingleClickOnDiagramElement` [mutation](https://github.com/eclipse-sirius/sirius-web/blob/381c77d69ca2d95d4b1f67aac410bb8cd40b1202/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useSingleClickTool.tsx#L70) and rely on `onCompleted`/`onError`. In this case we can probably get rid of the hook you mentioned above

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/tools/useSingleClickTool.tsx
```diff
@@ -120,6 +123,41 @@ export const useSingleClickTool = (): UseSingleClickToolValue => {
     });
   };
 
+  const { getImpactAnalysisReport, loading: impactAnalysisLoading, impactAnalysisReport } = useInvokeImpactAnalysis();
+
+  const [state, setState] = useState<UseSingleClickToolState>({
```
gdaniel, line 128: I had to use a state here to pass the information that is only known in `invokeSingleClickTool` to the `useEffect`.
Let me know if it needs to be modified, I have the feeling we shouldn't need a state but I don't know how to fix that.

PR #5094, opened 2025-06-18
[5093] Remove state machine from TextfieldPropertySection

### LINKED ISSUE #5093: Remove state machine from TextfieldPropertySection

mcharfadi:


frouene:
I've found a regression with the new version, when editing a number textfield with a non-number entry, the value is not removed after lost focus.
(see on the left, comparing the existant on the right)
textfield-regression

frouene (changes requested):
Regression with number textfield

frouene:
It seems ok now.
Can you add cypress tests, in particular to ensure non-regression on autocompletion, submit on focus lost, and on the number textfield validation.

frouene:
On completion without result, we still display an empty box 
image

integration-tests/cypress/e2e/project/details/textfield-auto-completion.cy.ts
```diff
@@ -0,0 +1,57 @@
...
+          details.getTextField('Domain Type').type('{ctrl}', { release: false });
+          details.getTextField('Domain Type').trigger('keydown', { keyCode: 32, which: 32 });
+          cy.getByTestId('completion-proposals').should('exist');
+          cy.getByTestId('completion-proposals')
+            .contains(actualValue as string)
+            .click();
+          details.getTextField('Domain Type').contains(actualValue as string);
+        });
+
+      // const firstElementText : string = cy.getByTestId('completion-proposals').children[0];
```
frouene: dead code

frouene (changes requested):

frouene:
A new strange behaviour with the auto completion feature, juste after use it to add some text if you press `space` (without `ctrl`) proposals appear (it's the previous behaviour, don't know if that's what's intended, though.)
But with the new version when try to add a proposal nothing happen.

textfield-proposal

mcharfadi:
> A new strange behaviour with the auto completion feature, juste after use it to add some text if you press `space` (without `ctrl`) proposals appear (it's the previous behaviour, don't know if that's what's intended, though.) But with the new version when try to add a proposal nothing happen.
> 
I don't think that this was supposed to happened. I think it was an issue with how we handled `ctrl+space` with a state to know if ctrl was pressed and it seems that the state was still true. 
I changed the listener to use` event.ctrlKey`

Also note that pressing the left or right arrow was not working on master so I removed this part.
We can still select a value with the keyboard by pressing tab and then use the arrows.

mcharfadi:
I added the `autofocus `prop to the `ProposalsList ` `MenuList `so we don't have to press tab before using the arrows.

frouene:
> I added the `autofocus `prop to the `ProposalsList ` `MenuList `so we don't have to press tab before using the arrows.

Yes, auto-completion is more pleasant to use with these changes

frouene (approved):

frouene:
Could be merged once the test updated

PR #5149, opened 2025-06-26
[5147] Add impact analysis before updating a library

### LINKED ISSUE #5147: Add impact analysis before updating a library

### Expected behavior
The _Update Library_ contextual menu entry should trigger an impact analysis before performing the update. 
The impact analysis report should (at least) provide the following information:
- References that will be broken by the update (new proxies that cannot be resolved)
- References that will be fixed by the updates (old proxies that can now be resolved)

In addition, the impact analysis report could display the following information:
- List of references that have been updated (from target@v1 to target@v2)
- Number of new elements in the editing context
- Number of deleted elements in the editing context

gdaniel:
Fixes 
- https://github.com/eclipse-sirius/sirius-web/issues/5147

Based on:
- https://github.com/eclipse-sirius/sirius-web/pull/5124
- https://github.com/eclipse-sirius/sirius-web/pull/5253

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -72,8 +72,9 @@ Stop the propagation on the dialog instead of the input.
 - https://github.com/eclipse-sirius/sirius-web/issues/5343[#5343] [core] `IRepresentationApplicationService.findRepresentationMetadataById()` can now resolve metadata for transient representations.
 - https://github.com/eclipse-sirius/sirius-web/issues/5359[#5359] [sirius-web] Add context to the images project settings tab capability.
 - https://github.com/eclipse-sirius/sirius-web/issues/5354[#5354] [diagram][tree] Add unicode support for direct edit input
-
-
+- https://github.com/eclipse-sirius/sirius-web/issues/5147[#5147] [sirius-web] Add impact analysis before performing a library update.
+The impact analysis report displays the following information:
+* References that will be broken by the update (new proxies that won't be resolved and will be deleted after the update)
 
 
 == v2025.8.0
--- a/packages/core/frontend/sirius-components-core/src/modals/impact-analysis/ImpactAnalysisDialog.tsx
+++ b/packages/core/frontend/sirius-components-core/src/modals/impact-analysis/ImpactAnalysisDialog.tsx
@@ -161,7 +161,7 @@ const ReportViewer = ({
                 <ListItemIcon sx={listItemIconStyle}>
                   <FiberManualRecordIcon sx={iconStyle} />
                 </ListItemIcon>
-                <ListItemText primary={additionalReport} />
+                <ListItemText primary={additionalReport} data-testid={`additionalReportData_${index}`} />
               </ListItem>
             );
           })}
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/UpdateLibraryExecutor.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/UpdateLibraryExecutor.java
@@ -1,0 +1,228 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.web.application.studio.services.library;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.EStructuralFeature.Setting;
+import org.eclipse.emf.ecore.resource.Resource;
+import org.eclipse.emf.ecore.resource.ResourceSet;
+import org.eclipse.emf.ecore.util.EcoreUtil.UnresolvedProxyCrossReferencer;
+import org.eclipse.sirius.components.collaborative.api.ChangeKind;
+import org.eclipse.sirius.components.core.api.IEditingContext;
+import org.eclipse.sirius.components.events.ICause;
+import org.eclipse.sirius.components.representations.Failure;
+import org.eclipse.sirius.components.representations.IStatus;
+import org.eclipse.sirius.components.representations.Message;
+import org.eclipse.sirius.components.representations.MessageLevel;
+import org.eclipse.sirius.components.representations.Success;
+import org.eclipse.sirius.web.application.UUIDParser;
+import org.eclipse.sirius.web.application.editingcontext.EditingContext;
+import org.eclipse.sirius.web.application.editingcontext.services.api.IEditingContextDependencyLoader;
+import org.eclipse.sirius.web.application.library.dto.UpdateLibraryInput;
+import org.eclipse.sirius.web.application.library.services.LibraryMetadataAdapter;
+import org.eclipse.sirius.web.application.studio.services.library.api.IUpdateLibraryExecutor;
+import org.eclipse.sirius.web.domain.boundedcontexts.library.Library;
+import org.eclipse.sirius.web.domain.boundedcontexts.library.services.api.ILibrarySearchService;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.SemanticData;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.SemanticDataDependency;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.services.api.ISemanticDataSearchService;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.services.api.ISemanticDataUpdateService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.data.jdbc.core.mapping.AggregateReference;
+import org.springframework.stereotype.Service;
+
+/**
+ * Used to execute a library update.
+ *
+ * @author gdaniel
+ */
+@Service
+public class UpdateLibraryExecutor implements IUpdateLibraryExecutor {
+
+    public static final String REMOVED_PROXIES_PARAMETER_KEY = "removed_proxies_parameter_key";
+
+    private final Logger logger = LoggerFactory.getLogger(UpdateLibraryExecutor.class);
+
+    private final ILibrarySearchService librarySearchService;
+
+    private final ISemanticDataSearchService semanticDataSearchService;
+
+    private final ISemanticDataUpdateService semanticDataUpdateService;
+
+    private final IEditingContextDependencyLoader editingContextDependencyLoader;
+
+    public UpdateLibraryExecutor(ILibrarySearchService librarySearchService, ISemanticDataSearchService semanticDataSearchService, ISemanticDataUpdateService semanticDataUpdateService,
+            IEditingContextDependencyLoader editingContextDependencyLoader) {
+        this.librarySearchService = Objects.requireNonNull(librarySearchService);
+        this.semanticDataSearchService = Objects.requireNonNull(semanticDataSearchService);
+        this.semanticDataUpdateService = Objects.requireNonNull(semanticDataUpdateService);
+        this.editingContextDependencyLoader = Objects.requireNonNull(editingContextDependencyLoader);
+    }
+
+    @Override
+    public IStatus updateLibrary(ICause cause, IEditingContext editingContext, UUID libraryId) {
+        IStatus result = new Failure(List.of());
+        if (editingContext instanceof EditingContext siriusWebEditingContext) {
+            Optional<Library> optionalNewLibrary = this.librarySearchService.findById(libraryId);
+            if (optionalNewLibrary.isPresent()) {
+                Library newLibrary = optionalNewLibrary.get();
+
+                // Find the version of the library already in the dependencies.
+                Optional<Library> optionalOldLibrary = this.getLibraryDependencyWithDifferentVersion(siriusWebEditingContext, newLibrary);
+                if (optionalOldLibrary.isPresent()) {
+                    Library oldLibrary = optionalOldLibrary.get();
+                    List<Resource> oldLibraryResources = this.getAllResourcesFromLibraryRecursiveDependencies(siriusWebEditingContext, oldLibrary);
+
+                    for (Resource oldLibraryResource : oldLibraryResources) {
+                        oldLibraryResource.unload();
+                        siriusWebEditingContext.getDomain().getResourceSet().getResources().remove(oldLibraryResource);
+                    }
+
+                    if (cause instanceof UpdateLibraryInput) {
+                        // Do not update the semantic data bounded context if we are not performing an actual library
+                        // update (this is for example the case when performing an impact analysis).
+                        this.addLibraryDependency(cause, siriusWebEditingContext, newLibrary);
+                        this.removeLibraryDependency(cause, siriusWebEditingContext, oldLibrary);
+                    }
+
+                    List<SemanticData> newDependencies = new ArrayList<>();
+                    this.semanticDataSearchService.findById(newLibrary.getSemanticData().getId())
+                            .ifPresent(newLibrarySemanticData -> {
+                                newDependencies.add(newLibrarySemanticData);
+                                newDependencies.addAll(this.semanticDataSearchService.findAllDependenciesRecursivelyById(newLibrarySemanticData.getId()));
+                            });
+
+                    this.editingContextDependencyLoader.loadDependencies(siriusWebEditingContext, newDependencies);
+
+                    long start = System.nanoTime();
+                    Map<EObject, Collection<Setting>> removedProxies = this.removeUnresolvedProxies(siriusWebEditingContext.getDomain().getResourceSet());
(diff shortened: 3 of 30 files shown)


packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextSnapshotService.java
```diff
@@ -65,14 +79,32 @@ public Optional<IEditingContextSnapshot> createSnapshot(IEditingContext editingC
...
+            result = Optional.of(new EditingContextSnapshot(snapshotDocuments, libraryAdapterResourcesMap, initialDependencies));
         }
         return result;
     }
 
     @Override
-    public void restoreSnapshot(IEditingContext editingContext, IEditingContextSnapshot snapshot) {
+    public void restoreSnapshot(IEditingContext editingContext, IEditingContextSnapshot snapshot, ICause cause) {
         if (editingContext instanceof EditingContext siriusWebEditingContext && snapshot instanceof EditingContextSnapshot siriusWebSnapshot) {
+            List<AggregateReference<SemanticData, UUID>> dependencies = new UUIDParser().parse(editingContext.getId())
```
gdaniel: Performing an update to compute the impact analysis report will change the dependencies of the editing context's semantic data, so we need to reset them to their original values once we restore the snapshot. If we don't, the editing context will look ok until we reload it.

I am not sure this is the way to go, but I didn't find a satisfying alternative

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/TreeItemContextMenu.tsx
```diff
@@ -73,6 +73,7 @@ export const TreeItemContextMenu = ({
         <TreeItemMenuContextComponent
           editingContextId={editingContextId}
           item={item}
+          entry={null}
```
gdaniel, line 76: I don't know if I should pass the `entry` to the component (and set it as `null` for frontend-only entries), or pass the `withImpactAnalysis` property. I went for the `entry` assuming this could be helpful in the future, but this `null` value is dubious at best.

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/libraries/LibraryControllerIntegrationTests.java
```diff
@@ -86,7 +86,7 @@ public class LibraryControllerIntegrationTests extends AbstractIntegrationTests
     private ISemanticDataSearchService semanticDataSearchService;
 
     @Autowired
-    private IObjectService objectService;
+    private IObjectSearchService objectSearchService;
```
gdaniel: Since I was looking at this package I fixed this deprecated dependency

packages/sirius-web/backend/sirius-web-e2e-tests/src/main/java/org/eclipse/sirius/web/e2e/tests/templates/TablePapayaTemplatesProvider.java
```diff
@@ -35,8 +35,8 @@ public class TablePapayaTemplatesProvider implements IProjectTemplateProvider {
 
     @Override
     public List<ProjectTemplate> getProjectTemplates() {
-        var flowTemplate = new ProjectTemplate(TABLE_PAPAYA_TEMPLATE_ID, "PapayaTable", "/project-templates/Papaya.png", List.of(new ProjectTemplateNature(PAPAYA_NATURE)));
-        return List.of(flowTemplate);
+        var papayaTemplate = new ProjectTemplate(TABLE_PAPAYA_TEMPLATE_ID, "PapayaTable", "/project-templates/Papaya.png", List.of(new ProjectTemplateNature(PAPAYA_NATURE)));
```
gdaniel, line 38: Fixed a typo in a class I used as inspiration

CHANGELOG.adoc
```diff
@@ -137,6 +137,8 @@ public class CanEditProjectCapabilityVoter implements ICapabilityVoter {
 ----
 - https://github.com/eclipse-sirius/sirius-web/issues/5181[#5181] [sirius-web] Remove the contribution that contribute the link to libraries to the `navigationBarMenuEntryExtensionPoint` extension point.
 The link to libraries has become a standard menu item which can be hidden providing a `ICapabilityVoter` with the type `Library` and the capability `view`.
+- https://github.com/eclipse-sirius/sirius-web/issues/5147[#5147] [sirius-web] The `EditingContextSnapshotService` now handles semantic data's dependencies.
+The method `void restore(IEditingContext, IEditingContextSnapshot)` has been updated to `void restore(IEditingContext, IEditingContextSnapshot, ICause cause)` to reflect this change.
```
sbegaudeau: If there is a cause, it should be the first parameter, always. Otherwise it should be the editing context, if it is present

CHANGELOG.adoc
```diff
@@ -205,6 +207,10 @@ The dialog can now display an impact analysis report, but isn't in charge of ret
 - https://github.com/eclipse-sirius/sirius-web/issues/5118[#5118] [tree] Add support for impact analysis in tree item contextual menu actions.
 - https://github.com/eclipse-sirius/sirius-web/issues/5123[#5123] [tree] Add support for tree item context menu action impact analysis in the view DSL.
 - https://github.com/eclipse-sirius/sirius-web/issues/5185[#5185] [sirius-web] Hide the download button instead of disabling it when the download capability is disabled.
+- https://github.com/eclipse-sirius/sirius-web/issues/5147[#5147] [sirius-web] Add impact analysis before performing a library update.
+The impact analysis report displays the following information:
+* Number of references that will be broken by the update (new proxies that won't be resolved after the update)
```
sbegaudeau: What do you mean by broken? Those proxies will be removed from the model right? You won't save models with broken proxies? We have always worked under the assumption that models are properly resolved in the database.

packages/core/frontend/sirius-components-core/src/modals/impact-analysis/ImpactAnalysisDialog.tsx
```diff
@@ -161,7 +161,7 @@ const ReportViewer = ({
                 <ListItemIcon sx={listItemIconStyle}>
                   <FiberManualRecordIcon sx={iconStyle} />
                 </ListItemIcon>
```
sbegaudeau, line 163: This doesn't have to use an expression

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextSnapshotService.java
```diff
@@ -68,14 +82,32 @@ public Optional<IEditingContextSnapshot> createSnapshot(IEditingContext editingC
...
+                    .toList();
+
+            result = Optional.of(new EditingContextSnapshot(snapshotDocuments, libraryAdapterResourcesMap, initialDependencies));
         }
         return result;
     }
 
     @Override
-    public void restoreSnapshot(IEditingContext editingContext, IEditingContextSnapshot snapshot) {
+    public void restoreSnapshot(IEditingContext editingContext, IEditingContextSnapshot snapshot, ICause cause) {
```
sbegaudeau: The cause should be the first parameter

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextSnapshotService.java
sbegaudeau: The entire premise of the mechanism to compute the impact analysis is based on the fact that there won't be any change in the system (i.e. the database). This cannot be accepted, there has to be another way to do this. I'm not sure why would that be necessary.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/api/IEditingContextSnapshotService.java
```diff
@@ -25,5 +26,5 @@ public interface IEditingContextSnapshotService {
 
     Optional<IEditingContextSnapshot> createSnapshot(IEditingContext editingContext);
 
-    void restoreSnapshot(IEditingContext editingContext, IEditingContextSnapshot snapshot);
+    void restoreSnapshot(IEditingContext editingContext, IEditingContextSnapshot snapshot, ICause cause);
```
sbegaudeau: Now that I've seen why this parameter has been added, I think that it should be removed

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/api/IUpdateLibraryExecutor.java
```diff
@@ -0,0 +1,30 @@
...
+import org.eclipse.sirius.components.representations.IStatus;
+
+/**
+ * Used to execute a library update.
+ *
+ * @author gdaniel
+ */
+public interface IUpdateLibraryExecutor {
+
+    IStatus updateLibrary(IEditingContext editingContext, UUID libraryId, ICause cause);
```
sbegaudeau: The cause should be first

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/TreeItemContextMenu.tsx
sbegaudeau: It indeed seems odd, we may need two signature for frontend and backend entries

sbegaudeau (changes requested):
I haven't tested it and there may be some small minor issues remaining but the core problem is the fact that the computation of the impact analysis would have a side effect.

CHANGELOG.adoc
gdaniel: I didn't know about this assumption, and I just realized you can't upload a model that has unresolved proxies.
I don't think I understand why Sirius Web makes this assumption.

Looking at this feature, allowing proxies ensures that switching from v1 to v2 to v1 back always restores the initial state. If we remove the proxies we lose information, which doesn't seem right from my point of view.

I assume I am missing a key point here.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextSnapshotService.java
gdaniel: I can probably work at the editing context level and replace the resources without changing the dependencies.
I didn't want to go that way because it felt odd to compute the impact analysis on "something else" than the actual operation we want to perform.

If this compromise is fine for you I can work in this direction.

CHANGELOG.adoc
sbegaudeau: > I didn't know about this assumption, and I just realized you can't upload a model that has unresolved proxies.
I don't think I understand why Sirius Web makes this assumption

We guarantee that the data are not broken to be able to perform any kind of operation free of mind. Since we load / save everything at once, we don't want to try to resolve a proxy while we are working on the data.

> Looking at this feature, allowing proxies ensures that switching from v1 to v2 to v1 back always restores the initial state. If we remove the proxies we lose information, which doesn't seem right from my point of view.

Wouldn't that be the responsibility of some versioning system. If I migrate my code to another version of Spring Boot with tons of API breaks and save my file, close my editor and do something else for a while. The information is lost too, Git saves this information for me and it allows me to go back but versioning is not in the scope of Sirius Web. As soon as we are out of the stack of undo/redo of the editor, the information is lost too in VS Code for example.

Having broken models in memory to keep some information to go back to a previous version is thus not our concern.

gdaniel: Right, this is the responsibility of the versioning system.
Based on this discussion, I'll create another PR to fix the update, because at the moment it does save proxies in the database.

gdaniel: I have created the PR with the feature we discussed https://github.com/eclipse-sirius/sirius-web/pull/5253

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextSnapshotService.java
sbegaudeau: I'm not sure why would that be necessary, we may need to talk directly to see how that can be fixed.

gdaniel: We discussed it, the solution we chose is to _not_ update anything in the database if we aren't performing the actual update.

CHANGELOG.adoc
gdaniel: The cause has been removed, as the entry in the changelog

CHANGELOG.adoc
gdaniel: I updated and clarified the changelog: the message will display references that will be broken (_and deleted_) after the update

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/EditingContextSnapshotService.java
gdaniel: The cause has been removed

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/editingcontext/services/api/IEditingContextSnapshotService.java
gdaniel: done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/api/IUpdateLibraryExecutor.java
gdaniel: fixed

packages/core/frontend/sirius-components-core/src/modals/impact-analysis/ImpactAnalysisDialog.tsx
gdaniel: Not sure I understand what you mean, should I update this as follows?
```
<ListItemIcon sx={{minWidth: 'auto'}}>
    <FiberManualRecordIcon sx={iconStyle} />
</ListItemIcon>
```

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/TreeItemContextMenu.tsx
gdaniel: I can do that but it would have a few impacts on existing menu contributions and potentially downstream applications.
The idea would be to change
```
export const treeItemContextMenuEntryExtensionPoint: ComponentExtensionPoint<TreeItemContextMenuComponentProps>
```
to
```
export const treeItemContextMenuEntryExtensionPoint: ComponentExtensionPoint<TreeItemContextMenuFrontendComponentProps>
```

We should agree on the naming, but to stay consistent, we may want to 
- Change the name of the extension point to reflect that it only contains _frontend_ contributions
- Change `TreeItemContextMenuComponentProps` to reflect that values of this type are bound to a backend menu entry

If I understood the direction tree menus are taking, I assume pure frontend menu entries will disappear in the near future (or at least be discouraged). If this is the case we will most likely not need this refactoring at all since pure frontend contributions won't be supported anymore.

I may be off though, let me know if you want me to do the refactoring I proposed above.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/library/UpdateLibraryExecutor.java
```diff
@@ -0,0 +1,228 @@
...
+                            });
+
+                    this.editingContextDependencyLoader.loadDependencies(siriusWebEditingContext, newDependencies);
+
+                    long start = System.nanoTime();
+                    Map<EObject, Collection<Setting>> removedProxies = this.removeUnresolvedProxies(siriusWebEditingContext.getDomain().getResourceSet());
+                    Duration timeToRemoveProxies = Duration.ofNanos(System.nanoTime() - start);
+                    this.logger.trace("Removed proxies in {}ms", timeToRemoveProxies.toMillis());
+
+                    result = new Success(ChangeKind.SEMANTIC_CHANGE, Map.of(REMOVED_PROXIES_PARAMETER_KEY, removedProxies), List.of(new Message("Library " + oldLibrary.getName() + " updated to version " + newLibrary.getVersion(), MessageLevel.SUCCESS)));
```
gdaniel, line 124: I am not sure of this implementation.
I added a parameter in the `Success` with the map of unresolved proxies found after the update.
The proxies have been deleted (this is the main goal of #5252 ), and this parameter is a trace of what has been deleted. I use it in the impact analysis to compute the message to display.

This implementation allows to find proxies only once (when we want to delete them), and not re-do the same work as part of the impact analysis computation.

The thing I am not sure about though, is whether these `Success#parameters` can be used for such purpose.

sbegaudeau (approved):
I still have a details or two that I'll need to look a bit more carefully but it can be merged as is for now. Those are very minor internal details. I'll rebase it and merge it after the build to ensure that it will be part of tomorrow's release.

PR #5187, opened 2025-07-08
[4886] Add action to create a shareable URL for a project

flatombe:
First implementation for #4886.

# Pull request template

## General purpose
What is the main goal of this pull request?
- [ ] Bug fixes
- [X] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management
- [ ] Has the pull request been added to the relevant project and milestone? (Only if you know that your work is part of a specific iteration such as the current one)
- [ ] Have the `priority:` and `pr:` labels been added to the pull request? (In case of doubt, start with the labels `priority: low` and `pr: to review later`)
- [ ] Have the relevant issues been added to the pull request?
- [ ] Have the relevant labels been added to the issues? (`area:`, `difficulty:`, `type:`)
- [ ] Have the relevant issues been added to the same project and milestone as the pull request?
- [ ] Has the `CHANGELOG.adoc` been updated to reference the relevant issues?
- [ ] Have the relevant API breaks been described in the `CHANGELOG.adoc`? (Including changes in the GraphQL API)
- [ ] In case of a change with a visual impact, are there any screenshots in the `CHANGELOG.adoc`? For example in `doc/screenshots/2022.5.0-my-new-feature.png`

## Frontend

This section is not relevant if your contribution does not come with changes to the frontend.

### General purpose
- [ ] Is the code properly tested? (Plain old JavaScript tests for business code and tests based on React Testing Library for the components)

### Typing
We need to improve the typing of our code, as such, we require every contribution to come with proper TypeScript typing for both changes contributing new files and those modifying existing files.
Please ensure that the following statements are true for each file created or modified (this may require you to improve code outside of your contribution).

- [X] Variables have a proper type
- [X] Functions’ arguments have a proper type
- [X] Functions’ return type are specified
- Hooks are properly typed:
	- [ ] `useMutation<DATA_TYPE, VARIABLE_TYPE>(…)`
	- [ ] `useQuery<DATA_TYPE, VARIABLE_TYPE>(…)`
	- [ ] `useSubscription<DATA_TYPE, VARIABLE_TYPE>(…)`
	- [ ] `useMachine<CONTEXT_TYPE, EVENTS_TYPE>(…)`
	- [X] `useState<STATE_TYPE>(…)`
- [X] All components have a proper typing for their props
- [X] No useless optional chaining with `?.` (if the GraphQL API specifies that a field cannot be `null`, do not treat it has potentially `null` for example)
[...]

PR #5214, opened 2025-07-15
[4886] Add some view data in the shareable URL

### LINKED ISSUE #4886: Navigation to other view/panel

flatombe:
Edit: Can now be reviewed, remember to remove last commit before merge.
Leaving past questions below as they may be of interest of reviewers

---
Follow-up to #5187 for #4886.

The goal is to get some early feedback to validate the implementation approach.
(note there is an extraneous commit on top just to set up a second page in the "Details" view for testing purposes)

Some questions/notes:
* In `packages/sirius-web/frontend/sirius-web-application/src/extension/DefaultExtensionRegistry.tsx` we now have to provide the RefObject creation code instead of creating it directly because we need to be inside of a BrowserRouter context for the instantiation to work. I wonder if maybe there's a better (or more React-y) way to do that?
* In the different view components (QueryView, DetailsView, etc.), the ref is typed as `WorkbenchViewHandle` when we actually provide a more specific type (e.g. `DetailsViewHandle` or `QueryViewHandle`). I think we might be able to fit it in the type system with generics, but I wonder if we want/need to do that?
* In the same spirit we need to cast the `WorkbenchViewConfiguration` down to the view-specific type (e.g. `QueryViewConfiguration`). Not sure if we want to tweak our types to not need that cast (I'm not well-versed enough in Typescript generics to see if it would be a good idea)
* On a more functional note, the configuration of a view can only be retrieved if the view is being shown, because otherwise the underlying data does not exist (e.g. when setting up an expression in the Query view and then switching back to the Details view, when creating a shareable URL then the contents of the Query view will not be shared).
---

# Pull request template

## General purpose
- [ ] Bug fixes
- [X] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Frontend

### General purpose

### Typing

- [X] Variables have a proper type
- [X] Functions’ arguments have a proper type
- [X] Functions’ return type are specified
- Hooks are properly typed:
	- [X] `useState<STATE_TYPE>(…)`
- [X] All components have a proper typing for their props
- [X] No useless optional chaining with `?.` (if the GraphQL API specifies that a field cannot be `null`, do not treat it has potentially `null` for example)
- [X] Nullable values have a proper type (for example `let diagram: Diagram | null = null;`)

### How to test this PR?

* Query view:
  * Open the view, add an expression into it, create a shareable URL => check that the URL contains the expression from the view.
  * Load the above shared URL (possibly in another browser because the view relies on local storage) => the expression is set up in the Query view.
* Explorer view:
  * Create or open a studio project. In the Explorer view, use the second tree description ("Domain explorer by DSL") and disable filter "Hide in-memory studio color palettes". Create a shareable URL => check that the URL contains the ID of the selected tree description and of the disabled filter.
  * Load the above shared URL => check that in the Explorer view, the second tree description is selected and the filter is disabled.
* Details view:
  * In a studio project, create and select an `Entity` element, open the Details view and open the second tab of the form. Create a shareable URL => check that the URL contains the ID of the second tab.
  * Load the above shared URL => check that the second tab is opened in the Details view.

### DIFF
--- a/integration-tests-cypress/cypress/e2e/project/workbench/workbench-configuration.cy.ts
+++ b/integration-tests-cypress/cypress/e2e/project/workbench/workbench-configuration.cy.ts
@@ -19,6 +19,7 @@ import { Workbench } from '../../../workbench/Workbench';
 import {
   workbenchConfigurationWithClosedPanels,
   workbenchConfigurationWithExpandedPanels,
+  workbenchConfigurationWithQueryView,
 } from './workbench-configuration.data';
 
 const projectName = 'Cypress - Workbench Configuration Resolution';
@@ -186,11 +187,32 @@ describe('Workbench Configuration Resolution', () => {
             },
           ]);
           const diagram: Diagram = new Diagram();
-          diagram.getDiagram('Topography2').should('not.exist');
+          diagram.getDiagram('Topography1').should('not.exist');
           diagram.getDiagram('Topography2').should('exist');
         });
       }
     );
+
+    context('When opening the project with a workbench configuration with a configuration for the "Query" view', () => {
+      let workbench: Workbench;
+      beforeEach(() => {
+        new Project().visit(projectId, undefined, {
+          qs: {
+            workbenchConfiguration: JSON.stringify(workbenchConfigurationWithQueryView),
+          },
+        });
+        workbench = new Workbench();
+      });
+      it('Then, in the "Query" view, the text is set as specified in the configuration', () => {
+        workbench.checkPanelContent('right', ['Query']);
+        cy.getByTestId('query-textfield')
+          .should('exist')
+          .get('div')
+          .get('textarea')
+          .invoke('val')
+          .should('equal', 'aql:self');
+      });
+    });
   });
 });
 
--- a/integration-tests-cypress/cypress/e2e/project/workbench/workbench-configuration.data.ts
+++ b/integration-tests-cypress/cypress/e2e/project/workbench/workbench-configuration.data.ts
@@ -60,3 +60,16 @@ export const workbenchConfigurationWithExpandedPanels: WorkbenchConfiguration = 
     },
   ],
 };
+
+const queryViewConfiguration = { id: 'query', isActive: true, queryText: 'aql:self' };
+
+export const workbenchConfigurationWithQueryView: WorkbenchConfiguration = {
+  mainPanel: null,
+  workbenchPanels: [
+    {
+      id: 'right',
+      isOpen: true,
+      views: [queryViewConfiguration],
+    },
+  ],
+};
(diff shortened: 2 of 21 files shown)


jessymallet:
Is it possible to add in this MR the capacity to know if the right or left panel are closed/opened? (expand or not)

flatombe:
> Is it possible to add in this MR the capacity to know if the right or left panel are closed/opened? (expand or not)

This is done in https://github.com/eclipse-sirius/sirius-web/pull/5287
Since it did not require all the changes from this PR, I've placed the corresponding commit at the bottom of the pile, so it could be merged faster.

flatombe:
@sbegaudeau I've left this PR as draft as it includes a dummy commit simply to facilitate the validation of the last commit.
If you have any hints regarding the questions I've left in the PR description, I'd like to take them into account if possible.
I think the most important one is about cleaning up the URL after loading it, because I feel it may be disconcerting for users to remain with a URL that is de-synch'd from the real workbench state.

packages/core/frontend/sirius-components-core/src/workbench/Workbench.types.ts
```diff
@@ -37,12 +37,17 @@ export interface WorkbenchViewContribution {
   side: WorkbenchViewSide;
   title: string;
   icon: React.ReactElement;
-  component: (props: WorkbenchViewComponentProps) => JSX.Element | null;
+  component: ForwardRefExoticComponent<
+    PropsWithoutRef<WorkbenchViewComponentProps> & RefAttributes<WorkbenchViewHandle>
+  >;
+  ref: MutableRefObject<WorkbenchViewHandle | null> | undefined;
+  createRef: () => MutableRefObject<WorkbenchViewHandle | null>;
```
sbegaudeau: I haven't reviewed everything but I feel that there is an easier way to type this, it's a bit odd to have something this complex. I'll look into it

sbegaudeau (commented):
I think that I'll split the first commit into two commits (one for the API change and one for its usage in the query view) because I think I'll need it for another concern soon.

sbegaudeau (approved):

PR #5328, opened 2025-08-12
[releng] Bump to @xyflow/react 12.10.2

mcharfadi:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [x] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


mcharfadi:
This version of XYFlow brings a stronger typage of nodes so we can no longer use  `Node<NodeData, DiagramNodeType>`

CHANGELOG.adoc
```diff
@@ -22,6 +22,7 @@ The interface `IDiagramContext` is now deprecated and it will be removed soon.
 All usage of `IDiagramContext` within Sirius Web have been replaced by `DiagramContext` for example in `IDiagramEventHandler#handle`.
 The record `DiagramContext` is still implementing `IDiagramContext` for the moment and it still provides methods like `getDiagram()` or `getDiagramEvents()` but consumers of Sirius Web should switch to the new methods like `diagram()` or `diagramEvents()` as soon as possible.
 - https://github.com/eclipse-sirius/sirius-web/issues/5277[#5277] [sirius-web] Update the `ProjectSearchRepositoryDelegate` to include natures along with the returned projects for `#findAll(Before|After)`.
+- [releng] Since the switch to @xyflow/react 12.8.2 everything typed as `Node<NodeData, DiagramNodeType>` is now typed `Node<NodeData>`
```
sbegaudeau: Why? I don't see any trace of such change on the changelog of ReactFlow. NodeProps seems to be still typed with both NodeData and NodeType.

sbegaudeau (changes requested):

CHANGELOG.adoc
mcharfadi: This is from [12.7.0](https://github.com/xyflow/xyflow/releases/tag/%40xyflow%2Freact%4012.7.0) `Improve typing for Nodes` 
We now have this error in some files :
```
Argument of type 'Node<NodeData>[]' is not assignable to parameter of type 'Node<NodeData, string>[]'.
  Type 'Node<NodeData>' is not assignable to type 'Node<NodeData, string>'.
    Type 'Node<NodeData>' is not assignable to type '{ type: string; }'.
      Types of property 'type' are incompatible.
        Type 'string | undefined' is not assignable to type 'string'.
          Type 'undefined' is not assignable to type 'string'.
```
This occurs when pushing a node typed as `Node<NodeData>`  into an array typed as `Node<NodeData, string>[]` for example
I think that the easiest way to fix is to get rid of `Node<NodeData, string>` for typing nodes

mcharfadi: As a note not related to the bump.
The xyflow doc recommend to type the nodes like this :
```
export type FreeFormNodeType = Node<
  {
    imageURL: string | null;
    positionDependentRotation: boolean;
  } & NodeData,
  'freeFormNode'
>;
```
And to use a union type representing all types available 
```
export type DiagramNodeType =
  | FreeFormNodeType
  | IconLabelNodeType
  | ListNodeType
  | EdgeAnchorNodeCreationHandlesType
  | EdgeAnchorNodeType;
```
Then use this union type when possible for example like this 
`const store = useStoreApi<DiagramNodeType, DiagramEdgeType>();`
But this is a considerable change and I was not able to make it work

sbegaudeau: > This occurs when pushing a node typed as `Node<NodeData>` into an array typed as `Node<NodeData, string>[]` for example
> I think that the easiest way to fix is to get rid of `Node<NodeData, string>` for typing nodes

Why remove some typing information instead of fixing the code to ensure that we always have `Node<NodeData, NodeType>`?

mcharfadi: If we set the type of the node, this information should not be lost.

The issue is not there, the issue is that `NodeType `from `xyflow ` is actually `string | undefinied`
See [here](https://github.com/xyflow/xyflow/blob/main/packages/react/src/types/nodes.ts)

So with `Node<NodeData> `we are correctly typing like `xyflow `and with `Node<NodeData, string> `we are not

So an alternative fix would be to do :
```
export type DiagramNodeType = string | undefined;

export type DiagramNodeTypes = {
  [key in DiagramNodeType as string]: ComponentType<NodeProps>;
};
```
And to replace every `Node<NodeData, string> `with `Node<NodeData, string | undefined>`

Using `Node<NodeData> ` seems easier

sbegaudeau: I understand that reactflow allows us to remove the type by making it optional with `string | undefined`, what I don't understand would be why we would like to do so since we have so much code with this type already.

In a similar fashion, React is fine with [not having some type for `useState`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v18/index.d.ts#L1891) and Apollo client is also fine with [the lack of typing for useQuery](https://github.com/apollographql/apollo-client/blob/de367f4af03a5ca2add764264df5431cd5629674/src/react/hooks/useQuery.ts#L196) and other hooks.

There are tons of example of TypeScript frameworks that are onboarding their consumers nicely without requiring everything to be strictly typed. In Sirius Web, we are constantly moving in a direction with additional safety checks so I don't see why, we should remove some existing typing.

mcharfadi: Well when we type a node like this 
`Node<NodeData, string>` we use the `Node type` from `xyflow`, see [here](https://github.com/xyflow/xyflow/blob/a58568f11bc0e1a1bdca1b3549e959e2e1ca0cdd/packages/react/src/types/nodes.ts#L13).

I understand that we could do our own Node type with string not undefined but that would requires a lot of work, we probably want to see how to improve this issue later.

sbegaudeau (commented):
This looks good on paper, I'll perform some tests and merge it if I don't encounter anything odd.

sbegaudeau (approved):
No issue detected with my tests, I'll wait for the continuous integration to merge it.

PR #5365, opened 2025-08-19
[releng] Use a distinct name for the diagram image server Docker image

pcdavid:
On "normal" commits we publish the result of building `packages/sirius-web/backend/sirius-web/Dockerfile` as `eclipsesirius/sirius-web`. Fine (altough maybe we don't need to do it so often).

On tags, we do this, and *then* we build and publish *under the same name* the result of building `diagram-image-server/Dockerfile`, which overwrites the actual Sirius Web docker image with the diagram-image-server one.


sbegaudeau (approved):

pcdavid:
See https://hub.docker.com/repository/docker/eclipsesirius/sirius-web-diagram-image-server for the new, distinct repo for the diagram image server image.

Issue #5222, opened 2025-07-17
Make getRepresentationPath more flexible in EditProjectView and DisplayLibraryView


flatombe:
* [X] **I have checked that this feature has not yet been suggested by someone else.**

### Expected behavior

Currently, both `EditProjectView` and `DisplayLibraryView` use a fixed `getRepresentationPath` function that is used as value for `RepresentationPathContext.Provider`.

But downstream applications may be using custom routes (especially to support the several `SemanticData `of an `EditingContext`) and so would need to customize how such paths are computed (e.g. to replace `/projects/${projectId}/edit/${representationId}` with `/projects/${projectId}@${branchId}/edit/${representationId}`).

`getRepresentationPath` should probably passed as a prop to `EditProjectView` / `DisplayLibraryView` instead.

### Cuttings backs

None

### Potential side effects

This adds a prop to `EditProjectView` and `DisplayLibraryView`, but we could use the currently-used values as the default value.

### Describe alternatives you've considered

None

sbegaudeau:
Why would that be necessary? What is the issue right now?

flatombe:
Sorry I did not include the original problem that led to the creation of this issue.

The original problem is that the "Share diagram" button from the diagram editor toolbar, and the "Share" action from the project contextual menu currently do not include the current contextual `SemanticData` of the project, so the URLs they produce always resolve to the default one (`main`).

More generally, it seems the `getRepresentationPath` used in these views should be maintained in sync with the routes of the application (as per the comment that is currently alongside the function definition). Since routes can be contributed via an extension point, I think these functions should be included also as part of the extension point.
For instance through a prop of these views (or another extension point, or any other extension mechanism).

PR #5478, opened 2025-09-12
[doc] Add shape to make path computation more flexible

flatombe:
Shape for #5222 

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [X] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

### DIFF
--- a/doc/iterations/2025.10/make_path_computation_more_flexible.adoc
+++ b/doc/iterations/2025.10/make_path_computation_more_flexible.adoc
@@ -1,0 +1,41 @@
+= (M) Make path computation more flexible
+
+== Problem
+
+In the frontend, the application sometimes needs to produce a URL for a particular state of the application.
+For instance in `ShareRepresentationModal` we want to produce a URL for a representation.
+Or in `ShareProjectModal` we want to produce a URL for a project.
+
+Currently, `EditProjectView` and `DisplayLibraryView` provide the `getRepresentationPath` function used to produce these paths.
+It relies on a hard-coded route.
+
+There are 2 issues with this hard-coded route:
+* It does not support URLs with named SemanticData, e.g. `/projects/123@main/edit`, which can be used by downstream applications to implement branches inside projects.
+* It makes assumptions on the routes of the application, i.e. that the routes contributed in the extension registry (for the default Sirius Web Application, that is in `packages\sirius-web\frontend\sirius-web-application\src\extension\DefaultExtensionRegistry.tsx`) are compatible with the hard-coded route.
+
+== Key Result
+
+* The hard-coded route should be made modular, contributed alongside the route itself in the extension registry, to ensure downstream applications can manage their own routes.
+* The default implementation should support URLs with named SemanticData.
+
+=== Acceptance Criteria
+
+* The 'Share representation' and 'Share project' actions in the default Sirius Web Application should support named SemanticDatas (i.e. work for `/projects/123@main/edit/456`).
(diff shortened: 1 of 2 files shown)


flatombe:
I pieced together a prototype implementation that seems to work as expected ('Share representation' and 'Share project' produce URLs containing the '@main' segment when it is in my URL): https://github.com/flatombe/sirius-web-dev/tree/dev/FLA/5222/impl

doc/iterations/2025.10/make_path_computation_more_flexible.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+In the frontend, the application sometimes needs to produce a URL for a particular state of the application.
+For instance in `ShareRepresentationModal` we want to produce a URL for a representation.
+Or in `ShareProjectModal` we want to produce a URL for a project.
+
+Currently, `EditProjectView` and `DisplayLibraryView` provide the `getRepresentationPath` function used to produce these paths.
+It relies on a hard-coded route.
+
+There are 2 issues with this hard-coded route:
+* It does not support URLs with named SemanticData, e.g. `/projects/123@main/edit`, which can be used by downstream applications to implement branches inside projects.
```
sbegaudeau, line 13: - There's no concept of branches in Sirius Web, that's not our concern.
- There is no UI, no real change in behavior, the change desired is very minor. As such, I don't think that it requires a shape nor even an ADR. How is this anything more than just that:

Image

So unless I'm mistaken, up until now, we are just talking about a typo in the Sirius Web code.

doc/iterations/2025.10/make_path_computation_more_flexible.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+=== Acceptance Criteria
+
+* The 'Share representation' and 'Share project' actions in the default Sirius Web Application should support named SemanticDatas (i.e. work for `/projects/123@main/edit/456`).
+* The 'Share representation' and 'Share project' actions should produce URLs matching the routes contributed to the extension registry.
+
+== Solution
+
+* The components specified in the routes should include a prop that is a function that can be used to produce paths for URLs.
+* `EditProjectView` and `DisplayLibraryView` should retrieve this prop and pass it down so that `ShareRepresentationModal` and `ShareProjectModal` can use it to produce a valid URL.
```
sbegaudeau, line 29: Since we are now talking about ADR-level concerns, this computation has to be registered in the extension registry to be extensible, those view don't need a props they should be able to retrieve it  directly from the extension registry with something like this:

In EditProjectView
```
  const { data: computeRepresentationPath } = useData(editProjectViewRepresentationPathProviderExtensionPoint);

  const getRepresentationPath = (representationId: string) => {
    return computeRepresentationPath(rawProjectId, representationId);
  };
```

With something a bit similar for the `DisplayLibraryView`...

sbegaudeau (changes requested):

sbegaudeau:
I've looked at the code of your branch and for me there are two potential concerns:

- Keeping the entire `projectId` which can very easily be fixed in less than 10 lines of code unless I'm mistaken and that does not deserve neither a shape nor an ADR, it's just a typo to fix here and there
- Making the path used for `EditProjectView` and `DisplayLibraryView` configurable and thus make those view independent from their paths. Which is a nice thing but is that really an issue for you today?

If you only have the first issue, just send a PR with the fix for the 2/3 places where the issue is present.

flatombe:
Ok, I'm closing this PR and opening one just for the fix of the first concern.

PR #5589, opened 2025-09-29
[5588] Set stable ids for DiagramFilterDescriptionProvider widgets

### LINKED ISSUE #5588: The filter element modal loses its expanded elements when refreshed
https://github.com/user-attachments/assets/98cb8ff9-db67-459f-aca1-b2fe0a32c562

The widgets are removed from the dom when refreshing the form, this is causing the expended elements to be lost.
We're rendering the widget here in `Group.tsx

```
        {group.widgets
          .filter((widget) => visibleWidgetIds.includes(widget.id))
          .map((widget, index) => (
            <PropertySection
              formId={formId}
              widget={widget}
              readOnly={readOnly}
              key={`${widget.id}#${index}`}
            />
          ))}
```

The issue is that some of the widgets don't have the same `widget.id `after a refresh so React can't tell it's the same component that need to be rendered since the key depends on `widget.id`

We should ensure that the widgets in `DiagramFilterDescriptionProvider `have a stable id between refresh.

mcharfadi:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/5588

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -113,6 +113,7 @@ This occured for the tree filter menu in the Explorer when a table was opened, a
 - https://github.com/eclipse-sirius/sirius-web/issues/5222[#5222] [sirius-web] Fix the 'Share Project' and 'Share Representation' actions to always provide the right URL
 - https://github.com/eclipse-sirius/sirius-web/issues/5568[#5568] [diagram] Prevent empty labels to be highlighted
 - https://github.com/eclipse-sirius/sirius-web/issues/5570[#5580] [explorer] Make sure newly created elements are automatically selected in the _Explorer_
+- https://github.com/eclipse-sirius/sirius-web/issues/5588[#5588] [diagram] Fix an issue where the collapsed state of the diagram filter modal was lost when refreshing the modal
 
 
 === New Features
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/diagram/services/filter/DiagramFilterDescriptionProvider.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/diagram/services/filter/DiagramFilterDescriptionProvider.java
@@ -198,7 +198,7 @@ public class DiagramFilterDescriptionProvider implements IDiagramFilterDescripti
     private CheckboxDescription createTreeCheckboxDescription() {
         return CheckboxDescription.newCheckboxDescription("diagram-filter/tree-checkbox")
                 .targetObjectIdProvider(variableManager -> variableManager.get(VariableManager.SELF, Object.class).map(this.objectService::getId).orElse(null))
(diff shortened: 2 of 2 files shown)


pcdavid (approved):

sbegaudeau (dismissed):
To be clear on the source of the issue:

- The id computed by the `WidgetIdProvider` uses four parameters `parentElementId`, `controlDescriptionId`, `targetObjectId` and `label`
- If one of those parameters changes, then the widget will have a new identity and thus lose its frontend state

If we look at the video of the issue, we can see that the issue comes from the fact that the label of the tree has been modified to be updated dynamically. Each time an element is selected, the label changes and thus its default identity changes too.

I suspect that this issue could be fixed by ONLY changing the id computed for the tree description and keep the `WidgetIdProvider` for all the other widgets. The solution proposed here is fine for this specific use case. It's designed this way.

sbegaudeau:
And don't hesitate to remove my request for change if I'm right (it should be fairly easy to test and thus merge).

pcdavid:
Maybe I misunderstood your remark, but if I revert this commit and do this instead:

```
     private TreeDescription createTreeDescription() {
         return TreeDescription.newTreeDescription("diagram-filter/tree")
-                .idProvider(new WidgetIdProvider())
+                .idProvider(variableManager -> "diagram-filter/tree")
                 .targetObjectIdProvider(variableManager -> variableManager.get(VariableManager.SELF, Object.class).map(this.objectService::getId).orElse(null))
                 .diagnosticsProvider(variableManager -> List.of())
                 .kindProvider(variableManager -> "")
```

I get the following behavior:

https://github.com/user-attachments/assets/1e101570-f640-4231-b79e-740f6a05d223

mcharfadi:
> To be clear on the source of the issue:
> 
> * The id computed by the `WidgetIdProvider` uses four parameters `parentElementId`, `controlDescriptionId`, `targetObjectId` and `label`
> * If one of those parameters changes, then the widget will have a new identity and thus lose its frontend state
> 
> If we look at the video of the issue, we can see that the issue comes from the fact that the label of the tree has been modified to be updated dynamically. Each time an element is selected, the label changes and thus its default identity changes too.
> 
> I suspect that this issue could be fixed by ONLY changing the id computed for the tree description and keep the `WidgetIdProvider` for all the other widgets. The solution proposed here is fine for this specific use case. It's designed this way.

Yes but actually, it's the the checkbox and the splitbutton that have a computed label that make the id instable
I modified the PR to only change these then

pcdavid:
Yes, changing the `idProvider()` in `createTreeCheckboxDescription` seems to be enough.

PR #5665, opened 2025-10-14
[doc] Add a shape and ADR to add the omnibox on the projects page

gdaniel:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


doc/iterations/2025.12/add_command_palette_on_projects_page.adoc
```diff
@@ -0,0 +1,44 @@
...
+== Key Result
+
+- A button (similar to the one in the project workbench) should be available on the navigation bar of the `projects` page
+- Clicking on the button (or pressing `CTRL + K`) should open the command palette
+- The command palette should display projects-related commands
+- The command palette should allow user to search commands
+
+=== Acceptance Criteria
+
+- A palette containing commands to create a project from the available templates should be provided.
```
gdaniel, line 17: I didn't mention the cross-project search here, it will be based on the result of this shape, but I don't think it's key for the integration of the command palette in the projects page

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+}
+----
+
+These new parameters will be added in other classes/functions to ensure they are correctly forwarded to the Omnibox.
+Specifically, the `OmniboxProvider#openOmnibox` function will require these parameters to let client code open an command palette with the appropriate `getCommands` and `executeCommand` functions.
+
+Sirius Web will provide these functions for the command palette that is displayed in the project workbench.
+These functions will rely on existing hooks to get the editing context (e.g. `useCurrentProject()`) and the selection (`useSelection()`), and will perform the same operation as they do now.
+
+Currently, the command palette has 2 modes: _commands_ and _search_.
```
gdaniel, line 35: I am not sure it is acceptable to keep these two modes hardcoded.
I did look at the code to remove the dependency between the omnibox and the editing context, but dealing with the modes seemed like an extra layer of complexity we could handle later.
If it is not the case, let me know, and I'll look how we can deal with this

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+== Decision
+
+Remove the `editingContextId` from the `Omnibox` and `OmniboxProvider` and add function parameters to retrieve and execute commands.
+Remove the `useSelection` hook from the `Omnibox`.
+
+The properties of the `Omnibox` will be changed as follows:
+
+[source, typescript]
+----
+export interface OmniboxProps {
```
gdaniel, line 17: I don't really like this API but I don't see a way around it, let me know if you have a better solution

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+  open: boolean;
+  getCommands: (query: string) => void;
+  commandLoading: boolean;
```
sbegaudeau: `loading`

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+  commandData: GQLGetOmniboxCommandsQueryData | null;
```
sbegaudeau: commands

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+Remove the `editingContextId` from the `Omnibox` and `OmniboxProvider` and add function parameters to retrieve and execute commands.
+Remove the `useSelection` hook from the `Omnibox`.
+
+The properties of the `Omnibox` will be changed as follows:
+
+[source, typescript]
+----
+export interface OmniboxProps {
+  open: boolean;
+  getCommands: (query: string) => void;
```
sbegaudeau: `onQuery`

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+  commandLoading: boolean;
+  commandData: GQLGetOmniboxCommandsQueryData | null;
+  executeCommand: (commandId: string) => void;
```
sbegaudeau: `onCommandClick`

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
sbegaudeau: I think this API could be better by thinking about the omnibox by itself as a reusable components, instead of thinking about its usage. I'll change some names

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+  executeCommandData: GQLExecuteOmniboxCommandData | null;
```
sbegaudeau: Is this necessary? Why should the omnibox be responsible for that? Right now, it is used to update the selection but this is highly specific to the use of the omnibox within the workbench. It would not have any sense in the project page which would instead execute some other omnibox command which would redirect the user on another page. I don't think the updated omnibox should know anything about the impact of the commands it will run.

We will probably have WorkbenchOmniboxCommands with the newSelection and ProjectOmniboxCommands with, among others an omnibox command redirecting the user on another page (executing this command would not even perform a GraphQL query). The command would literally be "move the page /new/project. Keep in mind that ALL projects will now be created using this /new/project page including projects created using templates. So to create a project thanks to the omnibox, you will not have to perform any request to the backend (apart from retrieve the commands).

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
sbegaudeau: I think those modes should stay hardcoded for now, we also want to search in the omnibox in the project page

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+Sirius Web will provide these functions for the command palette that is displayed in the project workbench.
+These functions will rely on existing hooks to get the editing context (e.g. `useCurrentProject()`) and the selection (`useSelection()`), and will perform the same operation as they do now.
+
+Currently, the command palette has 2 modes: _commands_ and _search_. 
+While the modes should be provided as parameters of the command palette, a first implementation will keep them hardcoded, leaving this refactoring as future work.
+This may imply to provide the following additional parameters to the Omnibox:
+
+[source, typescript]
+----
+getSearchResults: (query: string) => void;
```
sbegaudeau: onSearch

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+searchResultsLoading: boolean;
```
sbegaudeau: still loading, no need for multiple loading props

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
```diff
@@ -0,0 +1,52 @@
...
+searchResultsData: GQLGetOmniboxSearchResultsQueryData | null;
```
sbegaudeau: shouldn't those be commands too? Because in the project page, clicking on a search result should not use `useSelection()` to select it in the workbench

sbegaudeau (changes requested):
I'll update the shape / adr and merge it so that it can move forward. Feel free to contact me if any of my changes create an issue (we would amend the shape / ADR accordingly).

doc/adrs/209_decouple_command_palette_from_editing_context.adoc
sbegaudeau: It could even be a simple change to `onQuery` to include the mode...

sbegaudeau (approved):
I've simplified things a bit more by including the mode and by removing the fact that the Omnibox should be in charge of handling the result of the commands since those will be different from one case to another.

sbegaudeau (approved):

PR #5676, opened 2025-10-16
[doc] Add a shape to support tree presentation of impact analysis

gdaniel:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


doc/iterations/2025.12/add_support_for_tree_presentation_of_impact_analysis_reports.adoc
```diff
@@ -0,0 +1,54 @@
...
+- The tag "[type=Feature]" has been removed from the "tags" containment reference  
+- The name of the object has been changed from "Syson" to "SysON"
+- A new tag "[type=Bug]" has been added as a child of SysON
+
+=== Cutting backs
+
+
+== Rabbit holes
+
+The lifecycle of this feature could be an issue: we have to find a way to pass the list of semantic changes computed by the impact analysis to the tree.
```
gdaniel, line 50: I don't really know what could be a good solution for that.
We can maybe store the EMF `ChangeDescription` computed from the last impact analysis in the editing context to make it easily accessible from the tree (like the `inputId2change` map we have for undo/redo), but I am not sure it is a good approach.
Another option we can maybe consider is to have the backend compute a data structure that can be rendered as a static tree by the frontend. But this seems quite complex to do since trees aren't based on an off the shelf library.
We should clarify this point before working on the ADRs.

doc/iterations/2025.12/add_support_for_tree_presentation_of_impact_analysis_reports.adoc
```diff
@@ -0,0 +1,54 @@
...
+- Addition/deletion of a containment reference
+- Addition/deletion of a non-containment reference (i.e. not visible in the default explorer)
+- Update of an attribute's value
+
+== Solution
+
+Sirius Web will provide an impact analysis tree presenting the semantic data enriched with the changes computed from the impact analysis.
+Specifically, it will provide icon overlays to mark elements as new/deleted/modified, and virtual elements (with their dedicated icons) to represent changes that cannot be displayed in the usual containment structure of a tree (like attribute values or non-containment references).
+
+An element marked as modified has changes in its own feature (we do not mark the entire containment structure as modified, just the elements that are actually modified).
```
gdaniel, line 30: Since all the elements visible in the tree are either modified or contained modified objects, we don't see a reason to propagate the "modified" mark to parents.

doc/iterations/2025.12/images/tree_based_impact_analysis_dialog.png
gdaniel, line 1: This is a first version we should iterate on. 
The main issue we wanted to tackle here is to ensure we can display modifications in the containment structure, but also in non-containment references and attributes.

For example, the line
`(-) tags: [type=Feature]`
Could be replaced by an actual tag, with a red dot overlay to indicate it has been removed.

doc/iterations/2025.12/add_support_for_tree_presentation_of_impact_analysis_reports.adoc
gdaniel: More context on this: the way impact analysis works implies that the resources used to compute the `ChangeDescription` are no longer in the editing context once we leave the impact analysis computation.
All the resources are removed from the editing context and loaded again (see `IEditingContextSnapshotService#restoreSnapshot`) to ensure all the changes have been reset.
Working with the `EObjects` referenced in the `ChangeDescription` (assuming we have a way to access it) will be tedious since we can't compare them directly: code like this won't work out of the box:
```java
if(changeDescription.getObjectChanges().keySet().contains(self)) {
    computeChangedLabel(self);
}
```
This is not a major issue: we can work with IDs, but it makes me feel going this direction is maybe not the best option.

sbegaudeau: I think the most straightforward way to do it would be for the impact analysis to compute a tree shaped data structure. Someone will have to do it at some point, we should do it while we have all the information. This tree like structure (may not be a real tree in memory or in the GraphQL API) should then be rendered as a basic MUI Tree I think. Not as a tree representation with a subscription which would be overkill. This computation should be customizable to allow downstream projects to change the structure, label, etc of this tree.

We may reuse this static tree structure in other context where we want to display for a short time a static tree of objects. I have other use cases in mind but those are completely out of scope of this topic. We will just have to make sure not to hardcode to much things in this tree data structure.

We should have an API (with a default implementation and some delegates) to compute a tree from a `org.eclipse.emf.ecore.change.ChangeDescription`. Something like this:

```
public interface IDiagramImpactAnalysisReportProvider {
  ImpactAnalysisReport getReport(IEditingContext editingContext, List<IDiagramEvent> diagramEvents, ChangeDescription changeDescription);
}
```

With

```
public record ImpactAnalysisReport(String title, XxxTree tree, List<String> comments) {}
```

And use it in `InvokeDiagramImpactAnalysisToolEventHandler`. Same thing for the tree impact analysis. We just have to define what this simple tree like structure should look like and how much overlap it would have with the tree representation.

sbegaudeau (changes requested):

doc/iterations/2025.12/images/tree_based_impact_analysis_dialog.png
sbegaudeau: ~~I suspect that it would be better with a bigger dialog and with the three trees side by side. Because viewing a complex set of changes may not be easily readable otherwise~~

Nevermind I did not look properly at the screenshot

doc/iterations/2025.12/add_support_for_tree_presentation_of_impact_analysis_reports.adoc
sbegaudeau: ~~This report may need three trees if we want to display things like in the screenshot. Or one with a tree item for deleted, one for created and one for modified... I don't mind.~~

I did not look properly at the screenshot

doc/iterations/2025.12/images/tree_based_impact_analysis_dialog.png
gdaniel: I am not sure I see what are the 3 trees you mention.
Do you want a tree for additions, a tree for deletions, and a tree for modifications?

doc/iterations/2025.12/add_support_for_tree_presentation_of_impact_analysis_reports.adoc
sbegaudeau: This tree data structure will also be used for the search results of the new search view which will need to display a subset of the models as a tree.

sbegaudeau (approved):

Issue #2884, opened 2024-01-08
Authorize graphical DnD of multiple selected nodes 


jessymallet:
* [x] **I have checked that this feature has not yet been suggested by someone else.**

### Expected behavior

When user D&D two or more nodes in a container, all selected nodes should be moved in the target container and not only the first selected.

### Actual behavior : 
![multipleElementD D](https://github.com/eclipse-sirius/sirius-web/assets/24290600/99d483ff-d3f0-49b9-9c21-b13a8cb04378)

### Describe alternatives you've considered

DnD element one by one but it is not efficient.

pcdavid:
The basic idea will be to update the `dropNode` mutation, and in particular `DropOnDiagramInput` to replace the single `String droppedElementId` with a `List<String> droppedElementIds`.

The mutation should probably be renamed into `dropNodes`.

[Here](https://github.com/eclipse-sirius/sirius-web/blob/master/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/dropNode/useDropNode.ts#L167) on the frontend is where we get the full set of dragged nodes but currently only support the case where a single node is dragged.

PR #5716, opened 2025-10-24
[doc] Add ADR for DnD of multiple nodes

pcdavid:
Corresponding issue: https://github.com/eclipse-sirius/sirius-web/issues/2884

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,7 +12,6 @@
 - Improve the creation of new project
 - Add support for representations edit capability
 
-
 === Architectural decision records
 
 - [ADR-209] Decouple the command palette from the editing context
@@ -23,6 +22,7 @@
 - [ADR-214] Add the support for i18n
 - [ADR-215] Add support for displaying data trees
 - [ADR-216] Lower the coupling between Palette and DiagramDescription
+- [ADR-217] Add support for drag'n'drop of multiple elements at once on diagrams
 
 
 === Deprecation warning
--- a/diagram-image-server/package-lock.json
+++ b/diagram-image-server/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "diagram-image-server",
-  "version": "2025.10.4",
+  "version": "2025.10.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "diagram-image-server",
-      "version": "2025.10.4",
+      "version": "2025.10.5",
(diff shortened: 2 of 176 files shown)


doc/adrs/211_drop_multiple_elements.adoc
```diff
@@ -0,0 +1,88 @@
...
+  y: Float!
+}
+
+# Current name: DropNodePayload
+union DropNodesPayload = ErrorPayload | SuccessPayload
+----
+
+i.e. beside the name, the only actual change will be the *addition* of a new `droppedElementIds: [ID!]!` field to `DropNodesInput`.
+The previous `droppedElementId: ID!` field will be kept for compatibility to avoid breaking existing modelers.
+It will contain the first element of the new `droppedElementIds` array.
```
sbegaudeau: Why would we need that? Especially at the GraphQL level. I don't see the point of maintaining something like that.

sbegaudeau (changes requested):

doc/adrs/211_drop_multiple_elements.adoc
```diff
@@ -0,0 +1,88 @@
...
+
+==== Backend
+
+`MutationDropNodeDataFetcher` and `DropNodeEventHandler` (in particular `invokeDropNodeTool()`) will need to be updated to handle multiple nodes.
+
+New variables `DROPPED_NODES = "droppedNodes"` and `DROPPED_ELEMENTS = "droppedElements"` will be added, with the `List` of nodes and semantic elements dropped.
+
+==== View DSL
+
+No change is required in the View DSL definition itself, but when they are invoked, View-based `DropNodeTools` will see the new `droppedNodes` and `droppedElements` variables.
```
sbegaudeau: I would honestly replace the existing variables instead because there will be other use cases in which tools will have to be ready to accept one or more elements with tools on multiple selection for example. I would rather explain clearly once and for all to everybody that it's now a collection and you should get used to it. I tend to prefer a clean meaningful API break

doc/adrs/211_drop_multiple_elements.adoc
pcdavid: OK, I'll remove it from the GraphQL layer.

doc/adrs/211_drop_multiple_elements.adoc
pcdavid: OK

sbegaudeau (approved):
I'll merge it after the build of the bump

PR #5724, opened 2025-10-24
[5722] Decouple the command palette from the workbench

### LINKED ISSUE #5722: Decouple the command palette from the workbench

The behavior of the command palette is tightly coupled to the workbench: it requires an `editingContextId` and works with the `selection`.
In the future we'll want to put the command palette in other places, where these parameters are not available (see [this shape](https://github.com/eclipse-sirius/sirius-web/blob/master/doc/iterations/2025.12/add_command_palette_on_projects_page.adoc) for more information).

### Expected behavior

The command palette's behavior should be configurable via its props.
The existing GraphQL APIs (`Viewer#omniboxCommand`, `Viewer#omniboxSearch`, `Mutation#executeOmniboxCommand`) should be renamed to clearly indicate they are related to the omnibox in the workbench.

### Cuttings backs

Having a command palette in another location is out of the scope of this issue.
We'll just refactor the command palette to allow it in the future.

### Potential side effects

API breaks

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/5722

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


gdaniel:
This PR is quite large, but most of the changes are classes/components I have renamed to make it explicit that the current APIs are aimed at the integration of the command palette in the workbench.

The actual key changes are:
- Rework of the omnibox props to make it configurable
- Move the execution logic of the omnibox in `WorkbenchOmnibox`
- Removed `OmniboxObjectList` component, which wasn't necessary anymore

This PR does not contain new tests, since it doesn't add any feature to Sirius Web.
The command palette should work as before, future PR will use this work to integrate it in other places.

gdaniel:
Note that I didn't rename `IOmniboxCommandOrderer`, since it does not depend on the editing context.
I feel having context-dependent command ordering is a bit overkill, but I can add it if needed.

packages/core/frontend/sirius-components-omnibox/src/OmniboxCommandList.tsx
```diff
@@ -53,15 +41,11 @@ export const OmniboxCommandList = forwardRef(
       }
     };
 
-    const { selection } = useSelection();
-    const selectedObjectIds: string[] = selection.entries.map((entry) => entry.id);
-
     const handleOnActionClick = (command: GQLOmniboxCommand) => {
       if (command.id === 'search') {
         onModeChanged('Search');
```
gdaniel, line 46: As we discussed, this PR keeps the mode hardcoded in the omnibox.

packages/core/frontend/sirius-components-omnibox/src/OmniboxProvider.tsx
```diff
@@ -11,44 +11,58 @@
...
         event.preventDefault();
-        setState((prevState) => ({ ...prevState, open: false }));
+        onClose();
       }
     };
 
     document.addEventListener('keydown', keyDownEventListener);
     return () => document.removeEventListener('keydown', keyDownEventListener);
-  }, []);
+  }, [open]);
```
gdaniel, line 47: We need to have a dependency to `open` otherwise the value read from the props is always false. This change will attach/detach the listener everytime an user opens the omnibox.

packages/core/frontend/sirius-components-omnibox/src/index.ts
```diff
@@ -11,8 +11,18 @@
...
 export * from './OmniboxExtensionPoints';
 export * from './OmniboxExtensionPoints.types';
 export * from './OmniboxProvider';
-export { type GQLOmniboxCommand } from './useOmniboxCommands.types';
+export * from './useExecuteWorkbenchOmniboxCommand';
+export * from './useExecuteWorkbenchOmniboxCommand.types';
+export * from './useWorkbenchOmniboxCommands';
+export type {
+  GQLGetWorkbenchOmniboxCommandsQueryVariables,
+  GQLOmniboxCommand,
```
gdaniel, line 25: This object is defined in `useWorkbenchOmniboxCommands.types` and `useWorkbenchOmniboxSearchCommands.types`, I don't think it should be like that, but I don't know if we want to move it in `Omnibox.types` or somewhere else.

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/WorkbenchOmnibox.tsx
```diff
@@ -0,0 +1,121 @@
...
+    data: executeCommandData,
+  } = useExecuteWorkbenchOmniboxCommand();
+
+  useEffect(() => {
+    if (!executeCommandLoading && executeCommandData) {
+      onClose();
+    }
+  }, [executeCommandLoading, executeCommandData]);
+
+  const handleCommandClick = (command: GQLOmniboxCommand, mode: OmniboxMode) => {
```
gdaniel, line 91: This function needs the current `mode` of the omnibox to know how it should handle commands.

The initial solution @sbegaudeau proposed was to create helpers like `isWorkbenchCommand(command)` and `isWorkbenchSearchCommand(command)`.
This helper can only work if the `id` of the command follows a pattern such as `workspaceCommand://createModel`, or if the backend returns different subtypes of `OmniboxCommand`.
The first solution seemed sketchy to me: ids are not formatted this way for now, and downstream application might be silently impacted by this change (commands that do not follow the workspace command template wouldn't be executed anymore).
I have an alternative version of this PR with the second solution, but I discarded it because it created fluff and typing complexity in the backend. The backend would instantiate record types like `WorkbenchCommand` and `WorkbenchSearchCommand`, but these commands would all hold the same data. The multiplication of concrete types of commands also has an impact on the services and GraphQL API, which should be restricted to the actual type of command they can manipulate.

In the end, the current implementation looks better to me: it does not impact downstream application nor add new requirements on command ids, and it doesn't require various types of commands that would grow as the command palette is used in other contexts.

Anyways, if this solution doesn't work for you I can push the version I have with the second solution, let me know.

sbegaudeau (approved):
I'll merge it in a couple of minutes

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/WorkbenchOmnibox.tsx
sbegaudeau: I don't mind since I'm pretty sure that the mode will be entirely extracted from the omnibox and into the one using the omnibox.

PR #5747, opened 2025-10-29
[5726] Virtualize the display of search results to improve performance on large result sets

### LINKED ISSUE #5726: Virtualize frontend rendering of large result sets
When a search query returns lots of matches (hundreds or thousands), the current frontend component causes the browser to hang while rendering the full list, even though only a few of the elements are actually visible at a given time.

We should use the same virtualization technique (possibly sharing some code) as [already used in the Query view to handle this issue](https://github.com/eclipse-sirius/sirius-web/blob/10024bdd6923f06e101f4df6ad5062cde2095784/packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/query/QueryView.tsx#L317).

pcdavid:
- **[5725] Virtualize search results list to improve performance**
- **[fix] Fix warning from react-resizable-panels with opening/closing a side view**

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng


sbegaudeau (changes requested):
There seems to be a bit too much of overflow in the query view since it consider also the title of the result and the button to export results as a CSV file

https://www.youtube.com/watch?v=rv6tQkQoCmc

sbegaudeau (approved):

PR #5793, opened 2025-11-07
[doc] Add an ADR to ensure index results contain enough information

gdaniel:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
+++ b/doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
@@ -1,0 +1,99 @@
+= ADR-218 - Ensure Elasticsearch query results contain enough information
+
+== Context
+
+Sirius Web can populate Elasticsearch indices with information from project's semantic data.
+
+The main goal of these indices is to provide an API to retrieve information from semantic data without loading the corresponding editing context, for example, one can query Elasticsearch to retrieve all the elements matching a given string, regardless of where they are actually stored.
+
+The minimal information to display an object in Sirius Web is usually composed of a label and an icon.
+This information is computed by `ILabelService` on a semantic object, and can be extended by downstream applications.
+Since we don't want to load editing contexts, and we can't make any assumption about how labels and icons are computed, we need to ensure the Elasticsearch query API returns enough information to display query results.
+
+We should also ensure that client code can load and manipulate the corresponding semantic object if needed (e.g. to display additional information not stored in the index).
+
+== Decision
+
+At the moment, indexed documents contain only the raw data of their source objects.
+As explained above, this information is not sufficient to compute relevant label/icons without loading the source object.
+We will add `label` and `iconURLs` fields to indexed documents to ensure they can be presented to the end user in contexts where the actual semantic object is not accessible.
+
+We will also add an `editingContextId` field to indexed documents to ensure the semantic object can be loaded if needed.
+
+This information will be set during the indexing process (where the semantic object is accessible) using `ILabelService` (for `label` and `iconURL`) and `IEditingContext` (for `editingContextId`).
+
+We will define the following interface, which defines the base information Sirius Web needs to index.
+This interface will be used to populate the indices and reify results when performing queries on them.
+[source, java]
+----
+public interface IIndexEntry {
+  
+  String editingContextId();
+
+  String id();
+
+  String type();
+  
+  String label();
+  
+  List<String> iconURLs();
+
+}
+----
+
+Note that the interface does not contain the `projectId`.
+It can be found from the `editingContextId` via `ProjectSemanticDataSearchService` without loading the editing context.
+
+We will add a service to let downstream application define the content of the index.
+This interface will be implemented with the typical delegate mechanism we have in Sirius Web.
+[source, java]
+----
+public interface IIndexEntryProvider {
+  Optional<IndexEntry> getIndexEntry(IEditingContext editingContext, Object object);
+}
+----
+
+For example, we could define an `IIndexEntry` implementation for Papaya types as follows:
+[source, java]
+----
+public record PapayaTypeIndexEntry (
+  String editingContextId,
+  String id,
+  String label,
+  String type,
+  List<String> iconURLs,
+  String qualifiedName,
+  List<PapayaTag> tags
+) implements IIndexEntry {}
+
+public record PapayaTag (String key, String value) {}
+----
+
+We will add an `IIndexQueryService` interface (and its default implementation) to query indices and get `IIndexEntry` instances.
+[source, java]
+----
+public interface IIndexQueryService {
+
+    List<IIndexEntry> search(String query);
+
+}
+----
+
+== Consequences
+
+As for other fields in indexed document, we recognize this information can become outdated if the index is not properly synchronized with the semantic data.
+The synchronization between semantic data and indexed documents is out of the scope of this ADR.
+
+[NOTE]
+====
+In any case, the actual model objects can always be loaded using the `editingContextId` and the `id` fields of the indexed document.
+This would have a significant impact on performances, since it may require to load several editing contexts, but it would ensure that the information is accurate and up-to-date.
+This solution should be avoided as much as possible by indexing enough information to work with index entries directly, or by defining better indexing policies to keep indices up-to-date.
+====
+
+Adding new fields to the index will increase its size, and may require to handle field exclusions as part of query definition.
+For example, we may not want to consider the `iconURLs` and `editingContextId` fields when searching for a given term in the index.
+
+== Status
+
+Proposed.
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -9,6 +9,7 @@
 
 === Architectural decision records
 
+- [ADR-218] Ensure Elasticsearch query results contain enough information
 
 
 === Deprecation warning


doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
```diff
@@ -0,0 +1,61 @@
...
+  
+  String label();
+  
+  List<String> iconURLs();
+
+}
+----
+
+Note that the interface does not contain the `projectId`.
+It can be found from the `editingContextId` via `ProjectSemanticDataSearchService` without loading the editing context.
```
gdaniel, line 45: This will be the case for cross project search: we will need to load the `ProjectSemanticData` of each result to get
- The `projectId`
- The `ProjectSemanticData#name`
This will allow to pass information to the frontend to build an URL as follows: `projects/<projectId>@<semanticDataName>/edit?selection=<objectId>`

I think it is fine to do that instead of adding other fields in the index: loading `ProjectSemanticData` doesn't require to load the entire editing context.

doc/adrs/217_ensure_elasticsearch_query_results_contain_enough_information.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+We will also add an `editingContextId` field to indexed documents to ensure the semantic object can be loaded if needed.
+
+This information will be set during the indexing process (where the semantic object is accessible) using `ILabelService` (for `label` and `iconURL`) and `IEditingContext` (for `editingContextId`).
+
+We will define the following interface, which will contain enough information to display index query results to the end user, as well as load the actual semantic object if needed.
+This interface may evolve in the future to match new requirements like complex ES|QL query results.
+[source, java]
+----
+public interface IndexHit {
```
sbegaudeau: I'm not a fan of `IndexHit`, it seems more like an action related to caching with hit and miss. I would rather use something like `IndexEntry`. Wether we "hit" it or not is another thing.

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
sbegaudeau: I don't mind, another advantage of NOT using `projectId` is the fact that it can be used to index things that are in editing contexts but not in a project such as the content of a library. The smaller the dependency between Elasticsearch and the concept of "project", the better.

It could even allow us one day to extract some of this code in something like `sirius-components-elasticsearch` which could be reused by anyone using an editing context without any dependency to `sirius-web-application`. Some downstream projects could be interested in something like that one day. So in the end, it would be nice if the link with the concept of "project" remained in `sirius-web-application` and I think that it can be done.

doc/adrs/217_ensure_elasticsearch_query_results_contain_enough_information.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+Note that the interface does not contain the `projectId`.
+It can be found from the `editingContextId` via `ProjectSemanticDataSearchService` without loading the editing context.
+
+We will add an `IIndexQueryService` interface (and its default implementation) to query indices and get `IIndexHit` instances.
+[source, java]
+----
+public interface IIndexQueryService {
+
+    List<IIndexHit> searchInProjects(String query);
```
sbegaudeau: I would thus rename this `search` not `searchInProjects` :)

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
```diff
@@ -0,0 +1,71 @@
...
+public interface IIndexQueryService {
+
+    List<IIndexHit> searchInProjects(String query);
+
+}
+----
+
+== Consequences
+
+As for other fields in indexed document, we recognize this information can become outdated if the index is not properly synchronized with the semantic data.
```
sbegaudeau, line 84: That's the core assumption of the entire feature, our code exists to try to fix that by reindexing. Our entire architecture is built on the hypothesis of eventual consistency. No-one should expect that from Sirius Web (even on other cross-bounding context features)

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
```diff
@@ -0,0 +1,71 @@
...
+The synchronization between semantic data and indexed documents is out of the scope of this ADR.
+
+[NOTE]
+====
+In any case, the actual model objects can always be loaded using the `editingContextId` and the `id` fields of the indexed document.
+This would have a significant impact on performances, since it may require to load several editing contexts, but it would ensure that the information is accurate and up-to-date.
```
sbegaudeau, line 90: I really don't think that we should do that, we have asynchronous processes and we will have to live with that.

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
```diff
@@ -0,0 +1,71 @@
...
+We will define the following interface, which will contain enough information to display index query results to the end user, as well as load the actual semantic object if needed.
+This interface may evolve in the future to match new requirements like complex ES|QL query results.
+[source, java]
+----
+public interface IndexHit {
+  
+  String editingContextId();
+
+  String id();
+
```
sbegaudeau, line 36: The type could be useful too

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
```diff
@@ -0,0 +1,71 @@
...
+In any case, the actual model objects can always be loaded using the `editingContextId` and the `id` fields of the indexed document.
+This would have a significant impact on performances, since it may require to load several editing contexts, but it would ensure that the information is accurate and up-to-date.
+====
+
+Adding new fields to the index will increase its size, and may require to handle field exclusions as part of query definition.
+For example, we may not want to consider the `iconURLs` and `editingContextId` fields when searching for a given term in the index.
+
+== Status
+
+Proposed.
```
sbegaudeau, line 99: I agree with this ADR but I have a couple of points to add:

I don't care if what is being added in the index is related to semantic data, it does not have to be. By redirecting to `projects/<projectId>@<semanticDataName>/edit?selection=<objectId>`, the `objectId` does not have to be the id of some semantic data. I don't mind that this ADR talks about semantic data but the code does not have to have this constrain. I don't think that it will require any change in what you have done, I just need to be sure that it's understood.

To ensure that we have an example, I would like two things:

- Papaya types (class, interface, record, enum) should be indexed with their qualifiedName and with their tags (key and value)
- Papaya tags should not be indexed at all by themselves


So for example, a Papaya type like a Class, should be indexed with something like this:

```
public record PapayaTypeIndexEntry (
  String editingContextId,
  String id(),
  String label(),
  String type(),
  List<String> iconURLs(),
  String qualifiedName,
  List<PapayaTag> tags
) implements IIndexEntry {}

public record PapayaTag (String key, String value) {}
```

This way, one could ensure that:

- Some useless stuff is not indexed by our code (think about OwningMembership or whatever those are called in SysMLv2)
- Some additional stuff can be indexed like some specific features, computed or not. Otherwise we only index the label and type. I don't want us to choose for others what can be indexed, apart from the id, label, type and iconURLs

sbegaudeau (changes requested):

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
sbegaudeau: For those requirements, we should probably have a service like this:

```
public interface IIndexEntryProvider {
  Optional<IndexEntry> getIndexEntry(Object object);
}
```

Which we would use to ask someone to provide an index entry for a given object. Our default implementation would create the default `IndexEntry` but one could add a delegate (think about the same pattern as `IObjectSearchService`) to return an empty optional for some objects (like Papaya tags) or another implementation for other objects (like Papaya Types).

doc/adrs/217_ensure_elasticsearch_query_results_contain_enough_information.adoc
gdaniel: Renamed

doc/adrs/217_ensure_elasticsearch_query_results_contain_enough_information.adoc
gdaniel: Done

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
gdaniel: I agree, I put this sentence to hammer down the fact that eventual consistency also applies in this context.

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
gdaniel: I agree, I clarified this solution should be avoided. I just wanted to point out that you can retrieve the actual objects if needed.

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
gdaniel: added

doc/adrs/218_ensure_elasticsearch_query_results_contain_enough_information.adoc
gdaniel: I reworked a bit the ADR to clarify that `IIndexEntry` is used to define the content of the index as well as reify the query results.
I integrated the `IIndexEntryProvider` you propose in the ADR and added the `PapayaTypeIndexEntry` as an example.

Looking at your comment on the term "semantic data", I agree it is too restrictive, but I don't know what to use instead: data, element ?

sbegaudeau (approved):
We are in agreement, let's merge this one now

PR #5844, opened 2025-11-20
[5843] Add range selection to tree representation

### LINKED ISSUE #5843: Range selection currently not supported inside tree representations
Currently, we are supporting mult-selection using the CTRL key, but we cannot select a range of items using Shift.

theogiraudet:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/5843

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/browser/frontend/sirius-components-browser/src/modelbrowser/ModelBrowserTreeView.types.ts
+++ b/packages/browser/frontend/sirius-components-browser/src/modelbrowser/ModelBrowserTreeView.types.ts
@@ -10,7 +10,7 @@
  * Contributors:
  *     Obeo - initial API and implementation
  *******************************************************************************/
-import { GQLTreeItem } from '@eclipse-sirius/sirius-components-trees';
+import { GQLTree, GQLTreeItem } from '@eclipse-sirius/sirius-components-trees';
 
 export interface ModelBrowserTreeViewProps {
   editingContextId: string;
@@ -22,7 +22,7 @@ export interface ModelBrowserTreeViewProps {
   title: string;
   leafType: 'reference' | 'container';
   ownerKind: string;
-  onTreeItemClick: (event: React.MouseEvent<HTMLDivElement, MouseEvent>, item: GQLTreeItem) => void;
+  onTreeItemClick: (event: React.MouseEvent<HTMLDivElement, MouseEvent>, tree: GQLTree, item: GQLTreeItem) => void;
   selectedTreeItemIds: string[];
 }
 
--- a/packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerView.tsx
+++ b/packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerView.tsx
@@ -21,6 +21,7 @@ import {
 import {
   FilterBar,
   GQLGetTreePathVariables,
+  GQLTree,
   GQLTreeItem,
   TreeFilter,
   TreeToolBar,
@@ -29,6 +30,7 @@ import {
   TreeView,
   useTreeFilters,
   useTreePath,
+  useTreeSelection,
 } from '@eclipse-sirius/sirius-components-trees';
 import { Theme } from '@mui/material/styles';
 import { ForwardedRef, forwardRef, useCallback, useContext, useEffect, useRef, useState } from 'react';
@@ -80,6 +82,7 @@ export const ExplorerView = forwardRef<WorkbenchViewHandle, WorkbenchViewCompone
       tree: null,
       selectedTreeItemIds: [],
       singleTreeItemSelected: null,
+      selectionPivotTreeItemId: null,
     });
 
     // If we are requested to reveal the global selection, we need to compute the tree path to expand
@@ -195,6 +198,7 @@ export const ExplorerView = forwardRef<WorkbenchViewHandle, WorkbenchViewCompone
     }, [treeElement]);
 
     const { selection, setSelection } = useSelection();
+    const { treeItemClick } = useTreeSelection();
 
     const selectionKey: string = selection?.entries
       .map((entry) => entry.id)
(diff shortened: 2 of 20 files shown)


theogiraudet:
I was considering moving `useTreeSelection` to `@eclipse-sirius/sirius-components-selection` instead of `@eclipse-sirius/sirius-components-trees`, since `useTreeSelection` is not a hook specific to the tree representation.
However, `@eclipse-sirius/sirius-components-trees` does not currently depend on `@eclipse-sirius/sirius-components-selection`.

Should I add this dependency, or is it preferable to keep the hook in `@eclipse-sirius/sirius-components-trees`?

For context, `useTreeSelection` is currently used in:
- `@eclipse-sirius/sirius-components-trees`
- `@eclipse-sirius/sirius-components-widget-reference`
- `@eclipse-sirius/sirius-components-selection`

packages/browser/frontend/sirius-components-browser/src/modelbrowser/ModelBrowserTreeView.types.ts
```diff
@@ -22,7 +22,7 @@ export interface ModelBrowserTreeViewProps {
   title: string;
   leafType: 'reference' | 'container';
   ownerKind: string;
-  onTreeItemClick: (event: React.MouseEvent<HTMLDivElement, MouseEvent>, item: GQLTreeItem) => void;
+  onTreeItemClick: (event: React.MouseEvent<HTMLDivElement, MouseEvent>, item: GQLTreeItem, tree: GQLTree) => void;
```
sbegaudeau: The tree should be before the tree item in terms of importance of the parameter (since the tree item is derived from the tree)

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerView.tsx
```diff
@@ -289,36 +293,68 @@ export const ExplorerView = forwardRef<WorkbenchViewHandle, WorkbenchViewCompone
...
-          const newEntry: SelectionEntry = { id };
-          const newSelection: Selection = { entries: [...selection.entries, newEntry] };
-          setSelection(newSelection);
+    const computeSelectedElementsInRange = (
+      subtree: GQLTree | GQLTreeItem,
+      firstSelectionItemId: string,
+      secondSelectionItemId: string,
+      isSelecting: boolean = false,
+      accumulator: string[] = []
+    ): string[] => {
```
sbegaudeau: Is the accumulator returned in the end? So its both an entry argument which is being modified by the function and a return value? That's confusing. Don"t try to modify entry arguments, consider them immutable

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { TreeItemClickResult, UseTreeSelectionValue } from './useTreeSelection.types';
+
+/**
+ * Hook to handle the different scenarios of tree item selection: simple click, multi-selection (CTRL), range selection (SHIFT).
+ * @returns An object containing the function to handle tree item click events.
+ */
+export const useTreeSelection = (): UseTreeSelectionValue => {
+  const [pivotId, setPivotId] = useState<string | null>(null);
+
+  const computeSelectedElementsInRange = useCallback(
```
sbegaudeau: - Why useCallback?
- Why is that function defined twice, once here and once in the explorer view?

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.tsx
```diff
@@ -0,0 +1,127 @@
...
+          }
+        }
+      }
+
+      return isSelecting;
+    },
+    []
+  );
+
+  const treeItemClick = useCallback(
```
sbegaudeau: What issue did you encounter that require the use of `useCallback`?

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.tsx
```diff
@@ -0,0 +1,127 @@
...
+    (
+      event: React.MouseEvent<HTMLDivElement, MouseEvent>,
+      item: GQLTreeItem,
+      tree: GQLTree,
+      selectedElements: string[],
+      allowMultiSelection: boolean = true
```
sbegaudeau: Don't specify default value, I prefer people to specify what they want explicitely

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.types.ts
```diff
@@ -0,0 +1,29 @@
...
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+import { GQLTree, GQLTreeItem } from '../views/TreeView.types';
+
+export interface UseTreeSelectionValue {
+  treeItemClick: (
+    event: React.MouseEvent<HTMLDivElement, MouseEvent>,
+    item: GQLTreeItem,
+    tree: GQLTree,
```
sbegaudeau, line 19: tree before treeId

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.types.ts
```diff
@@ -0,0 +1,29 @@
...
+    selectedElements: string[],
+    allowMultiSelection?: boolean
```
sbegaudeau: Do not use optional parameters

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerView.tsx
```diff
@@ -289,36 +293,68 @@ export const ExplorerView = forwardRef<WorkbenchViewHandle, WorkbenchViewCompone
...
-        if (isItemInSelection) {
-          const newSelection: Selection = { entries: selection.entries.filter((entry) => entry.id !== item.id) };
-          setSelection(newSelection);
-        } else {
-          const { id } = item;
-          const newEntry: SelectionEntry = { id };
-          const newSelection: Selection = { entries: [...selection.entries, newEntry] };
-          setSelection(newSelection);
+    const computeSelectedElementsInRange = (
+      subtree: GQLTree | GQLTreeItem,
```
sbegaudeau: Prefer one specific type, here I suspect that you only need the children not the parent

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.tsx
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Hook to handle the different scenarios of tree item selection: simple click, multi-selection (CTRL), range selection (SHIFT).
+ * @returns An object containing the function to handle tree item click events.
+ */
+export const useTreeSelection = (): UseTreeSelectionValue => {
+  const [pivotId, setPivotId] = useState<string | null>(null);
+
+  const computeSelectedElementsInRange = useCallback(
+    (
+      subtree: GQLTree | GQLTreeItem,
```
sbegaudeau: Same comment

sbegaudeau (changes requested):

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.tsx
theogiraudet: The initial call to this function, in `useTreeSelection` directly pass the tree. It avoids me to call the function for all tree's children. I you prefer, I can replace the `subtree: GQLTree | GQLTreeItem` by `children: GQLTreeItem[]` which is another solution, but I think this is less clear.

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.tsx
```diff
@@ -0,0 +1,123 @@
...
+        const includeSecond = selectedElements.includes(secondSelectionItemId);
+        const hasEndpoint = includeFirst || includeSecond;
+        const hasBothEndpoints = includeFirst && includeSecond;
+        // Early return if range was completed in children
+        if (hasBothEndpoints || (isSelecting && hasEndpoint)) {
+          return selectedElements;
+        }
+        // If we didn't complete, and we are not already selecting but we found an endpoint in children, then we start selecting
+        isSelecting = isSelecting || hasEndpoint;
+      }
```
theogiraudet, line 69: I think this part may be a bit tricky.

The main idea is to update the `isSelecting` state depending on whether a range endpoint is found during the recursive calls. One way to simplify this, and avoid performing an `.includes` check at every recursive step, would be to return something like `{ selectedElement: string[], foundEndpoints: number }` instead of just `string[]`.

However, this makes the function signature less clear in my opinion, since `foundEndpoints` is more of an internal detail of the recursion rather than a meaningful result for `useTreeSelection`.

Another option would be to wrap the whole function inside a parent function that manages the state (`isSelecting`, `selectedElements`) externally:

```ts
const computeSelectedElementsInRange = (
  tree: GQLTree,
  firstSelectionItemId: string,
  secondSelectionItemId: string,
): string[] => {
  let isSelecting = false;
  const selectedElements: string[] = [];

  const computeSelectedElementsInRangeInternal = (
    subtree: GQLTree | GQLTreeItem,
  ) => {
    <actual code>
  };

  computeSelectedElementsInRangeInternal(tree);
  return selectedElements;
};
```

But this solution introduces three nested functions, which we may prefer to avoid.

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const computeSelectedElementsInRange = (
+    subtree: GQLTree | GQLTreeItem,
+    firstSelectionItemId: string,
+    secondSelectionItemId: string,
+    isSelecting: boolean
+  ): string[] => {
+    const selectedElements: string[] = [];
+    for (const element of subtree.children) {
+      const isEndpoint = element.id === firstSelectionItemId || element.id === secondSelectionItemId;
+      debugger;
```
sbegaudeau: To be removed

packages/trees/frontend/sirius-components-trees/src/trees/useTreeSelection.tsx
sbegaudeau: > One way to simplify this, and avoid performing an .includes check at every recursive step, would be to return something like { selectedElement: string[], foundEndpoints: number } instead of just string[].

We are returning custom data structure from functions pretty much everywhere in the code base without issues. I prefer smaller functions that do less things or that are clearer instead of being smarter and more complex.

sbegaudeau (approved):

PR #5877, opened 2025-11-27
[5873] Add button for mini map visibility in diagram panel

### LINKED ISSUE #5873: Add the ability to hide the mini map in diagram

frouene:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/5873

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanel.types.ts
```diff
@@ -16,6 +16,8 @@ export interface DiagramPanelProps {
   onSnapToGrid: (snapToGrid: boolean) => void;
   helperLines: boolean;
   onHelperLines: (helperLines: boolean) => void;
+  showMiniMap: boolean;
+  onShowMiniMap: (showMiniMap: boolean) => void;
```
sbegaudeau: The goal of the DiagramPanel is not to expose the API of all its child components, we removed some of those props from here in the past, it's not to add more now. Custom ones would not have this luxury and they deal with it. The goal is not to create additional coupling between the DiagramPanel and any of its content.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/mini-map/useMiniMap.types.ts
```diff
@@ -0,0 +1,17 @@
...
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+export interface UseMiniMapValue {
+  showMiniMap: boolean;
```
sbegaudeau: showMiniMap is a verb which indicates an action, prefer something like that:

```
export interface UseMiniMapValue {
  isMiniMapVisible: boolean;
  setMiniMapVisibility: (visible: boolean) => void;
}
```

sbegaudeau (changes requested):

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanel.types.ts
frouene: What is the objective?
Use `children` props to add content to DiagramPanel, but controlled by DiagramRender and therefore potentially by a custom implementation.

sbegaudeau: One can add some stuff directly in `DiagramRenderer` with `diagramRendererReactFlowPropsCustomizerExtensionPoint` and one can also add some stuff in `DiagramPanel` with `diagramPanelActionExtensionPoint`. The contribution to the `DiagramPanel` may need to interact with the diagram using others components. For example, in the panel I could have a button which would open a dialog which would need to trigger some change in the diagram.

I don't want any of those concerns to have any impact on the API and state of `DiagramPanel` which is not the case at all right now with stuff like this:

``
`
    const handleFitToScreen = () => {
      if (getSelectedNodes().length) {
        fitView({ duration: 200, nodes: getSelectedNodes() });
      } else {
        fitView({ duration: 200, nodes: getNodes() });
      }
    };
    const handleZoomIn = () => zoomIn({ duration: 200 });
    const handleZoomOut = () => zoomOut({ duration: 200 });
    const handleShare = () => setState((prevState) => ({ ...prevState, dialogOpen: 'Share' }));
    const handleCloseDialog = () => setState((prevState) => ({ ...prevState, dialogOpen: null }));

    const { fadeDiagramElements } = useFadeDiagramElements();
    const { hideDiagramElements } = useHideDiagramElements();
    const { pinDiagramElements } = usePinDiagramElements();

    const onUnfadeAll = () => fadeDiagramElements([...getAllElementsIds()], false);
    const onUnhideAll = () => hideDiagramElements([...getAllElementsIds()], false);
    const onUnpinAll = () => pinDiagramElements([...getAllElementsIds()], false);
```

The `DiagramPanel` is not supposed to have the state and API of all its contents. It doesn't have any reason to be tightly coupled with all those concerns. Each will end up extracted one day in different individual buttons with their own state and APIs. We are not there yet but I don't want to go in the wrong direction by adding more stuff.

The children props would not help in any way because it would only make `DiagramRender` more coupled with those concerns either. Here, it makes sense for "something" to be in `DiagramRenderer` to update the `MiniMap` component when something happens in the button but it doesn't mean that intermediary components need to know anything about it.

Comment (frouene) [reply packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanel.types.ts]: done

File: packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanelContext.tsx [hunk 3]
```diff
@@ -0,0 +1,73 @@
...
+import React, { useState } from 'react';
+import { DiagramPanelContextValue, DiagramPanelContextProviderProps } from './DiagramPanelContext.types';
+
+const defaultContextValue: DiagramPanelContextValue = {
+  isMiniMapVisible: false,
+  setMiniMapVisibility: () => {},
+  isSnapToGridEnabled: false,
+  setSnapToGridEnabled: () => {},
+  isHelperLineEnabled: false,
+  setHelperLineEnabled: () => {},
```
sbegaudeau: It's great to have removed the impact of all three features from the diagram panel, I was only expecting the minimap one in thisPR but:

- None of those features should take the name of a DiagramPanelContext. I don't want any of those features to feel hardcoded in the diagram panel in any way
- The three features should not be coupled together. I need three contexts for three features just like any other contribution would have to do it.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanelContext.tsx
```diff
@@ -0,0 +1,73 @@
...
+    if (!isLocalStorageAvailable()) {
+      return true;
+    }
+    const storedPreference = localStorage.getItem(localStorageKey);
+    return storedPreference ? JSON.parse(storedPreference) : true;
+  };
+
+  const [miniMap, setMiniMap] = useState<boolean>(getInitialMiniMapVisibility);
+  const [snapToGrid, setSnapToGrid] = useState<boolean>(false);
+  const [helperLine, setHelperLine] = useState<boolean>(true);
```
sbegaudeau: Those three states have nothing to do together, the support from the helper line or the snap to grid has nothing to do with the local storage support.

sbegaudeau (changes requested):
I don't want coupling between unrelated features, I don't want to trade some coupling between the minimap and the diagram panel for a new coupling between the helper line, the snape to grid and the minimap. Those are three unrelated features that should act like three unrelated features.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/panel/DiagramPanelContext.tsx
frouene: Done, I have extracted three contexts

sbegaudeau (approved):
I'll merge this now, soon we will be able to extract the various `DiagramPanel` buttons into dedicated components which will limit even further their impact on `DiagramPanel`

PR #5968, opened 2025-12-16
[1626] Add the ability to trigger diagram tools using key bindings

### LINKED ISSUE #1626: Add the ability to trigger diagram tools using key bindings
It should be possible for a specific to define a KeyBinding (`key` along with `ctrl`, `alt`, `meta`) for a tool. We should consider the fact that:

- We want key bindings to be displayed in the palette
- We may want multiple keybindings for one tool (or the ability to make some parts optional) to support a tool which would be trigged on both Ctrl+H and Cmd+H (windows vs macOS)

It should be able to trigger the impact analysis if is has been defined for the tool, along with a selection dialog too for example.

The hardcoded delete tools should then be deleted (GraphQL mutation, `deletable` property on the frontend and GraphQL schema, delete tool in the backend event handlers etc)

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/1626
Fixes #1626 

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

Create a studio.
Open the diagram, click on the diagram background and press `CTRL + e`
An entity is created and selected
Press `CTRL + s`
A text attribute is created in the entity and selected.


packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
```diff
@@ -386,6 +386,7 @@ interface Tool {
   id: ID!
   label: String!
   iconURL: [String!]!
+  keyBindings: [KeyBinding!]!
```
gdaniel: I added this for both `SingleClickOnDiagramElementTool` and `SingleClickOnTwoDiagramElementTool`, although for the moment it is only supported for `SingleClickOnDiagramElementTool` in the frontend.

To be precise, all the code is there in this PR to define key bindings on `SingleClickOnTwoDiagramElementTool` instances, the frontend will retrieve them as part of the tools in the palette, but the execution of the tool expects a `SingleClickOnDiagramElementTool` and will not execute other tools (see `useDiagramKeyBinding` and `useSingleClickTool`).

I decided to leave this as is and work on the support of edge tool key bindings in a future PR, but let me know if you think we should get rid of everything related to `SingleClickOnTwoDiagramElementTool` and I'll remove it.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
```diff
@@ -0,0 +1,161 @@
...
+    if (!loading && palette && state.currentEvent) {
+      const allPaletteTools = getAllPaletteTools(palette);
+      const toolToExecute = allPaletteTools.filter((tool) => toolHasKeyBinding(tool, state.currentEvent!));
+      state.currentEvent.preventDefault();
+      const { domNode } = store.getState();
+      const domElement = domNode?.getBoundingClientRect();
+      const { x, y } = computeCursorPosition(mousePosition, viewport, domElement);
+      const selectedDiagramElementIds = getSelectedDiagramElementIds();
+      const targetObjectId = getTargetObjectId(selectedDiagramElementIds);
+      if (toolToExecute.length === 1 && toolToExecute[0]) {
```
gdaniel, line 107: `toolToExecute[0]` was needed to make the TS compiler happy

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
```diff
@@ -0,0 +1,161 @@
...
+        executeTool(
+          x,
+          y,
+          selectedDiagramElementIds[0] ? selectedDiagramElementIds : [diagramId],
+          targetObjectId,
+          () => {},
```
gdaniel: TODO @gdaniel check if we can use `useSingleClickTool` instead

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
```diff
@@ -0,0 +1,161 @@
...
+  return tool.keyBindings.some(
+    (keyBinding) =>
+      keyBinding.key === event.key &&
+      keyBinding.isCtrl === event.ctrlKey &&
+      keyBinding.isAlt === event.altKey &&
+      keyBinding.isMeta === event.metaKey
+  );
+};
+
+export const useDiagramKeyBinding = (diagramTargetObjectId: string): UseDiagramKeyBindingValue => {
```
gdaniel, line 62: The `diagramTargetObjectId` is passed as a parameter because we can't retrieve it from the `DiagramContext`.
I don't know if this is a good solution or if we should make this available in the context.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.tsx
```diff
@@ -76,21 +80,13 @@ const isDiagramDescription = (
...
-    loading,
-    error: paletteError,
-  } = useQuery<GQLGetToolSectionsData, GQLGetToolSectionsVariables>(getPaletteQuery, {
-    variables: {
-      editingContextId,
-      diagramId,
-      diagramElementIds,
-    },
-  });
+  const [getPaletteContents, { data: paletteData, loading, error: paletteError }] = useLazyQuery<
```
gdaniel: The query has to be lazy now so we can call it on demand when a key binding is detected.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/DiagramPalette.tsx
```diff
@@ -43,7 +43,17 @@ export const DiagramPalette = memo(({ diagramId, diagramTargetObjectId }: Diagra
 
   const elementId = diagramElementIds[0] ? diagramElementIds[0] : diagramId;
   const elementsIds = diagramElementIds[0] ? diagramElementIds : [diagramId];
-  let { palette }: UsePaletteContentValue = usePaletteContents(elementsIds);
+  let { getPaletteContents, palette }: UsePaletteContentValue = usePaletteContents();
```
gdaniel: See my comment on the palette query being now lazy.

packages/palette/frontend/sirius-components-palette/src/Palette.ts
```diff
@@ -28,3 +28,24 @@ export const isPaletteDivider = (entry: GQLPaletteDivider): entry is GQLToolSect
   entry.__typename === 'PaletteDivider';
 
 export const isTool = (entry: GQLPaletteEntry): entry is GQLTool => !isPaletteDivider(entry) && !isToolSection(entry);
+
+export const getToolTooltip = (tool: GQLTool) => {
```
gdaniel: I put this helper here because it is used by both `Tool.tsx` (in `sirius-components-diagrams`) and `ToolListItem.tsx` (in `sirius-components-palette`).

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/representations/EntityNodeDescriptionProvider.java
```diff
@@ -232,9 +233,9 @@ private EdgeTool entitySupertypeEdgeTool(NodeDescription edgeTargetNodeDescripti
     }
 
     private NodeToolSection entityNodeCreationTools() {
-        var stringCreationTool = this.entityAttributeCreationTool("Text", "aql:'/icons/full/obj16/StringAttribute.svg'", "newString", "aql:domain::DataType::STRING");
-        var booleanCreationTool = this.entityAttributeCreationTool("Boolean", "aql:'/icons/full/obj16/BooleanAttribute.svg'", "newBoolean", "aql:domain::DataType::BOOLEAN");
-        var numberCreationTool = this.entityAttributeCreationTool("Number", "aql:'/icons/full/obj16/NumberAttribute.svg'", "newNumber", "aql:domain::DataType::NUMBER");
+        var stringCreationTool = this.entityAttributeCreationTool("Text", "aql:'/icons/full/obj16/StringAttribute.svg'", "newString", "aql:domain::DataType::STRING", "s");
```
gdaniel, line 236: Note that `CTRL + t` would be more aligned with the other key bindings (`CTRL + <first letter of the tool label>`).
But the default behavior for the browser on `CTRL + t` (open a new tab) cannot be overridden.

scripts/check-coverage.jsh
```diff
@@ -82,7 +82,7 @@ var moduleCoverageData = List.of(
   new ModuleCoverage("sirius-components-domain-edit", 70.0),
   new ModuleCoverage("sirius-components-domain-emf", 98.0),
   new ModuleCoverage("sirius-components-view", 64.0),
-  new ModuleCoverage("sirius-components-view-edit", 51.0),
+  new ModuleCoverage("sirius-components-view-edit", 48.0),
```
gdaniel: I added tests in this PR, but some of the generated code isn't tested.

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/java/org/eclipse/sirius/components/collaborative/diagrams/dto/KeyBinding.java
```diff
@@ -0,0 +1,77 @@
...
+package org.eclipse.sirius.components.collaborative.diagrams.dto;
+
+import java.util.Objects;
+
+/**
+ * A key binding.
+ *
+ * @author gdaniel
+ */
+public record KeyBinding(boolean isCtrl, boolean isMeta, boolean isAlt, String key) {
```
gdaniel: TODO @gdaniel move this in `sirius-components-collaborative` DTO package (this needs to be reusable for both trees and diagrams)

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
```diff
@@ -0,0 +1,161 @@
...
+          y
+        );
+      } else if (toolToExecute.length > 1) {
+        const keysPressed =
+          (state.currentEvent.ctrlKey ? 'CTRL + ' : '') +
+          (state.currentEvent.metaKey ? 'META + ' : '') +
+          (state.currentEvent.altKey ? 'ALT + ' : '') +
+          state.currentEvent.key;
+        const toolLabels = toolToExecute.map((tool) => tool.label).join(', ');
+        addErrorMessage(`Found multiple tools for associated to ${keysPressed}: ${toolLabels}`);
```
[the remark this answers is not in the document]
gdaniel: done

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
```diff
@@ -0,0 +1,161 @@
...
+    );
+  };
+
+  const { getPaletteContents, palette, loading } = usePaletteContents();
+
+  useEffect(() => {
+    if (!loading && palette && state.currentEvent) {
+      const allPaletteTools = getAllPaletteTools(palette);
+      const toolToExecute = allPaletteTools.filter((tool) => toolHasKeyBinding(tool, state.currentEvent!));
+      state.currentEvent.preventDefault();
```
gdaniel: TODO @gdaniel this cannot work: this is not happening in the bubbling chain of the event.
`preventDefault` needs to be called in `onKeyBinding` to work

gdaniel: Done

integration-tests-playwright/playwright/resources/studioKeyBindings.xml
gdaniel, line 1: This file is named **studio**KeyBinding because it will be used for both diagram and explorer key binding tests

packages/core/frontend/sirius-components-core/src/mouse/useMousePosition.ts
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+
+import { useEffect, useState } from 'react';
+import { UseMousePositionState, UseMousePositionValue } from './useMousePosition.types';
+
+export const useMousePosition = (): UseMousePositionValue => {
```
sbegaudeau, line 17: Do we really need something that perform a state update and thus a re-render anytime a mouse is moving? I'll look into it but that seems odd

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/java/org/eclipse/sirius/components/collaborative/diagrams/dto/SingleClickOnDiagramElementTool.java
```diff
@@ -65,6 +68,8 @@ public static final class Builder {
 
         private boolean withImpactAnalysis;
 
+        private List<KeyBinding> keyBindings = List.of();
```
sbegaudeau: Don't initialize things in builder, it's the responsibility of the caller to provide the value

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/java/org/eclipse/sirius/components/collaborative/diagrams/dto/SingleClickOnDiagramElementTool.java
```diff
@@ -99,8 +104,14 @@ public Builder withImpactAnalysis(boolean withImpactAnalysis) {
             return this;
         }
 
+        public Builder keyBindings(List<KeyBinding> keyBindings) {
+            this.keyBindings = keyBindings;
```
sbegaudeau: Objects.requireNonNull

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/java/org/eclipse/sirius/components/collaborative/diagrams/dto/SingleClickOnTwoDiagramElementsTool.java
```diff
@@ -57,6 +61,8 @@ public static final class Builder {
 
         private String dialogDescriptionId;
 
+        private List<KeyBinding> keyBindings = List.of();
```
sbegaudeau: Same comment

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/java/org/eclipse/sirius/components/collaborative/diagrams/dto/SingleClickOnTwoDiagramElementsTool.java
```diff
@@ -81,8 +87,13 @@ public Builder dialogDescriptionId(String dialogDescriptionId) {
             return this;
         }
 
+        public Builder keyBindings(List<KeyBinding> keyBindings) {
+            this.keyBindings = keyBindings;
```
sbegaudeau: Same comment

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
sbegaudeau: I don't know yet, I'm not a fan of dead code, I'll look into this to see how much code is involved and how visible it is

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/DiagramRenderer.tsx
```diff
@@ -97,6 +98,7 @@ export const DiagramRenderer = memo(({ diagramRefreshedEventPayload }: DiagramRe
 
   const { onDirectEdit } = useDiagramDirectEdit();
   const { onDelete } = useDiagramDelete();
+  const { onKeyBinding } = useDiagramKeyBinding(diagramRefreshedEventPayload.diagram.targetObjectId);
```
sbegaudeau, line 101: Am I wrong in believing that since this uses `useDiagramKeyBinding` which uses `useMousePosition` then it trigger a re-render of the entire `DiagramRenderer` anytime a mouse is moving somewhere?

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.tsx
sbegaudeau: Why not expose `skip` and skip calling it until you are ready to do so? It would have a lower impact on the code changed

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/DiagramPalette.tsx
sbegaudeau: I'm not sure that it's necessary

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  const { getPaletteContents, palette, loading } = usePaletteContents();
+
+  useEffect(() => {
+    if (!loading && palette && state.currentEvent) {
+      const allPaletteTools = getAllPaletteTools(palette);
+      const toolToExecute = allPaletteTools.filter((tool) => toolHasKeyBinding(tool, state.currentEvent!));
+      const { domNode } = store.getState();
+      const domElement = domNode?.getBoundingClientRect();
+      const { x, y } = computeCursorPosition(mousePosition, viewport, domElement);
```
sbegaudeau, line 104: Given the performance implication, I don't think that it's worth the cost of doing that compared to provide `[0,0]` by default which would be more than enough for the vast majority of use cases

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/representations/EntityNodeDescriptionProvider.java
sbegaudeau: Because Ctrl+s is better? O__ô

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/DropNodeToolBuilder.java
```diff
@@ -72,6 +72,18 @@ public DropNodeToolBuilder body(org.eclipse.sirius.components.view.Operation ...
         return this;
     }
 
+    /**
+     * Setter for KeyBindings.
+     *
+     * @generated
+     */
+    public DropNodeToolBuilder keyBindings(org.eclipse.sirius.components.view.KeyBinding ... values) {
```
sbegaudeau: How could that possibly work?

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/DropToolBuilder.java
```diff
@@ -72,6 +72,18 @@ public DropToolBuilder body(org.eclipse.sirius.components.view.Operation ... val
         return this;
     }
 
+    /**
+     * Setter for KeyBindings.
+     *
+     * @generated
+     */
+    public DropToolBuilder keyBindings(org.eclipse.sirius.components.view.KeyBinding ... values) {
```
sbegaudeau: Event better, this is related to the drag and drop from other representations, adding keybinding here should not be possible

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/EdgeReconnectionToolBuilder.java
```diff
@@ -56,6 +56,18 @@ public EdgeReconnectionToolBuilder body(org.eclipse.sirius.components.view.Opera
         return this;
     }
 
+    /**
+     * Setter for KeyBindings.
+     *
+     * @generated
+     */
+    public EdgeReconnectionToolBuilder keyBindings(org.eclipse.sirius.components.view.KeyBinding ... values) {
```
sbegaudeau: Same comments

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/EdgeToolBuilder.java
```diff
@@ -72,6 +72,18 @@ public EdgeToolBuilder body(org.eclipse.sirius.components.view.Operation ... val
         return this;
     }
 
+    /**
+     * Setter for KeyBindings.
+     *
+     * @generated
+     */
+    public EdgeToolBuilder keyBindings(org.eclipse.sirius.components.view.KeyBinding ... values) {
```
sbegaudeau: I don't mind a bit of dead code in our code but here in the view DSL that's not possible. People would expect that to work

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/SourceEdgeEndReconnectionToolBuilder.java
```diff
@@ -72,6 +72,18 @@ public SourceEdgeEndReconnectionToolBuilder body(org.eclipse.sirius.components.v
         return this;
     }
 
+    /**
+     * Setter for KeyBindings.
+     *
+     * @generated
+     */
+    public SourceEdgeEndReconnectionToolBuilder keyBindings(org.eclipse.sirius.components.view.KeyBinding ... values) {
```
sbegaudeau: Same comments

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/LabelEditToolBuilder.java
```diff
@@ -72,6 +72,18 @@ public LabelEditToolBuilder body(org.eclipse.sirius.components.view.Operation ..
         return this;
     }
 
+    /**
+     * Setter for KeyBindings.
+     *
+     * @generated
+     */
+    public LabelEditToolBuilder keyBindings(org.eclipse.sirius.components.view.KeyBinding ... values) {
```
sbegaudeau: Can this work with your PR?

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/TargetEdgeEndReconnectionToolBuilder.java
```diff
@@ -72,6 +72,18 @@ public TargetEdgeEndReconnectionToolBuilder body(org.eclipse.sirius.components.v
         return this;
     }
 
+    /**
+     * Setter for KeyBindings.
+     *
+     * @generated
+     */
+    public TargetEdgeEndReconnectionToolBuilder keyBindings(org.eclipse.sirius.components.view.KeyBinding ... values) {
```
sbegaudeau: Same comments

packages/view/backend/sirius-components-view-diagram/src/main/java/org/eclipse/sirius/components/view/diagram/impl/NodeToolImpl.java
```diff
@@ -286,6 +288,7 @@ public Object eGet(int featureID, boolean resolve, boolean coreType) {
      *
      * @generated
      */
+    @SuppressWarnings("unchecked")
```
sbegaudeau, line 323: ?

packages/view/backend/sirius-components-view-diagram/src/main/resources/model/diagram.ecore
```diff
@@ -355,6 +355,9 @@
         eType="ecore:EDataType ../../../../../sirius-components-view/src/main/resources/model/view.ecore#//InterpretedExpression"/>
     <eStructuralFeatures xsi:type="ecore:EReference" name="body" upperBound="-1" eType="ecore:EClass ../../../../../sirius-components-view/src/main/resources/model/view.ecore#//Operation"
         containment="true"/>
+    <eStructuralFeatures xsi:type="ecore:EReference" name="keyBindings" upperBound="-1"
```
sbegaudeau: I don't think that this can be available on any kind of tool

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/tools/api/IKeyBindingFactory.java
```diff
@@ -0,0 +1,25 @@
...
+package org.eclipse.sirius.components.view.emf.diagram.tools.api;
+
+import org.eclipse.sirius.components.collaborative.dto.KeyBinding;
+
+/**
+ * Used to create key bindings.
+ *
+ * @author gdaniel
+ */
+public interface IKeyBindingFactory {
```
sbegaudeau: Shouldn't that be a KeyBindingConverter instead?

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/tools/ViewConnectorToolsProvider.java
```diff
@@ -182,6 +187,16 @@ private SingleClickOnTwoDiagramElementsTool createEdgeTool(AQLInterpreter interp
                         )
                 )
                 .dialogDescriptionId(this.diagramIdProvider.getId(edgeTool.getDialogDescription()))
+                .keyBindings(keyBindings)
```
sbegaudeau: Now that I've seen the code, I prefer not to see any impact on this kind of tools

sbegaudeau (changes requested):

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/DiagramRenderer.tsx
gdaniel: No you are right, I didn't notice a performance issue when I tested but this is obviously not good, I'll remove this hook, and use the `[0,0]` position you propose some comments below

sbegaudeau: I thought about it a bit more and if it was necessary, we could maybe use a hook like `useMousePosition` but it would have to keep track of the position in a ref instead of a state to prevent re-renders and with a function to only return the mouse position when asked:

```
const { getMousePosition } = useMousePosition()
```

I'm not sure that it would work but it could be a viable option without the performance impact.

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
gdaniel: The field has been moved to `SingleClickOnDiagramElementTool`

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/usePaletteContents.tsx
gdaniel: Done

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/palette/DiagramPalette.tsx
gdaniel: It wasn't necessary

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/representations/EntityNodeDescriptionProvider.java
gdaniel: I set `CTRL + s` for **s**tring.
The default browser behavior for `CTRL + s` can be overridden, while `CTRL + t` cannot.
So if there is an issue with the letter I selected, please let me know what you prefer for this key binding.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/DiagramRenderer.tsx
gdaniel: In the end I implemented the `ref` solution, which works fine and does not trigger unnecessary re-renders.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
gdaniel: With the new implementation of `useMousePosition` the performance issue related to re-rendering is not a problem anymore.

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/DropNodeToolBuilder.java
gdaniel: I can't, I updated the ecore file to have `keyBindings` in `NodeTool` and not in other tools.

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/DropToolBuilder.java
gdaniel: I updated the ecore file to have `keyBindings` in `NodeTool` and not in other tools.

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/EdgeReconnectionToolBuilder.java
gdaniel: I updated the ecore file to have `keyBindings` in `NodeTool` and not in other tools.

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/EdgeToolBuilder.java
gdaniel: I updated the ecore file to have `keyBindings` in `NodeTool` and not in other tools.

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/SourceEdgeEndReconnectionToolBuilder.java
gdaniel: I updated the ecore file to have `keyBindings` in `NodeTool` and not in other tools.

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/LabelEditToolBuilder.java
gdaniel: No, I updated the ecore file to have `keyBindings` in `NodeTool` and not in other tools.

packages/view/backend/sirius-components-view-builder/src/main/java/org/eclipse/sirius/components/view/builder/generated/diagram/TargetEdgeEndReconnectionToolBuilder.java
gdaniel: I updated the ecore file to have `keyBindings` in `NodeTool` and not in other tools.

packages/view/backend/sirius-components-view-diagram/src/main/resources/model/diagram.ecore
gdaniel: No, I updated the ecore file to have `keyBindings` in `NodeTool` and not in other tools.

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/tools/api/IKeyBindingFactory.java
gdaniel: I don't mind renaming it `KeyBindingConverter`, but in this case I don't understand what makes `NodeToolFactory` a factory and not a converter.

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/tools/ViewConnectorToolsProvider.java
gdaniel: I removed the impacts on these tools

packages/core/frontend/sirius-components-core/src/mouse/useMousePosition.ts
gdaniel: It has been changed to a solution using a ref instead of a state

packages/view/backend/sirius-components-view-diagram/src/main/java/org/eclipse/sirius/components/view/diagram/impl/NodeToolImpl.java
gdaniel: This is added during the generation.
It is also present in `NodeToolSectionImpl#eSet`. I suspect it is added when the generated code manipulates a collection.

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/java/org/eclipse/sirius/components/collaborative/diagrams/dto/SingleClickOnDiagramElementTool.java
gdaniel: done

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/java/org/eclipse/sirius/components/collaborative/diagrams/dto/SingleClickOnTwoDiagramElementsTool.java
gdaniel: I removed key bindings from this tool

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/java/org/eclipse/sirius/components/collaborative/diagrams/dto/SingleClickOnTwoDiagramElementsTool.java
gdaniel: I removed key bindings from this tool

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/java/org/eclipse/sirius/components/collaborative/diagrams/dto/SingleClickOnDiagramElementTool.java
gdaniel: Done

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
```diff
@@ -0,0 +1,155 @@
...
+          state.currentEvent.key;
+        const toolLabels = toolToExecute.map((tool) => tool.label).join(', ');
+        addErrorMessage(`Found multiple tools associated to ${keysPressed}: ${toolLabels}`);
+      }
+      // There is no tool associated to the pressed keys, don't do anything.
+      // In any case, reset the current event, we don't need it anymore and we don't want to reuse it.
+      setState((prevState) => ({
+        ...prevState,
+        currentEvent: null,
+      }));
```
gdaniel, line 123: I had to add this to prevent this hook from creating a new element on every click on the diagram once a valid key binding has been entered.
This feels weird but I didn't see another way to skip the get palette query.
To reproduce: comment this `setState`, open a studio diagram, click on the diagram, press `CTRL + e`, an entity is created, then click anywhere else on the diagram.

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/diagram/tools/ViewGroupPaletteProvider.java
```diff
@@ -186,24 +183,6 @@ private Palette getGroupDiagramPalette(org.eclipse.sirius.components.view.diagra
         return new Palette(paletteId, quickAccessTools, paletteEntries);
     }
 
-    private ITool createNodeTool(NodeTool viewNodeTool, VariableManager variableManager, AQLInterpreter interpreter) {
```
gdaniel, line 189: I reused `NodeToolFactory` here, which does the same thing as this method

gdaniel:
Since there are quite a lot of comments in this PR, here are the key points after I integrated your feedbacks:

- I didn't rename `KeyBindingFactory` to `KeyBindingConverter`, I followed the same pattern we use for `NodeToolFactory`
- I had to add a weird `setState` in `useDiagramKeyBindings` to make it work properly with the exposed `skip` parameter

sbegaudeau:

This PR successfully implements key bindings for diagram tools, allowing users to trigger actions via keyboard shortcuts (e.g., Ctrl+E for New Entity). It correctly spans the full stack from EMF model updates and GraphQL schema
  changes to React hook implementations and comprehensive E2E tests, effectively addressing issue #1626 (https://github.com/eclipse-sirius/sirius-web/issues/1626).


  <details>


   * React Performance: Missing Dependency Array in `useEffect`:
      In packages/core/frontend/sirius-components-core/src/mouse/useMousePosition.ts, the useEffect hook lacks a dependency array. This causes the mousemove event listener to be added and removed on every single render, which is a significant performance bottleneck.
       * Rule: react-frontend-reviewer rule 5 (Effect Dependencies).
       * Fix: Add an empty dependency array [].
   1     useEffect(() => {
   2       const updateMousePosition = (event: MouseEvent) => { /* ... */ };
   3       window.addEventListener('mousemove', updateMousePosition);
   4       return () => window.removeEventListener('mousemove', updateMousePosition);
   5     }, []); // Add this


   * React Stale Closures in `DiagramRenderer`:
      In packages/diagrams/frontend/sirius-components-diagrams/src/renderer/DiagramRenderer.tsx, the onKeyDown callback has an empty dependency array [] but it calls onKeyBinding. Since onKeyBinding is re-created (or at least comes from a hook in the same component), onKeyDown will capture a stale version of the function.
       * Rule: react-frontend-reviewer rule 5 (Effect Dependencies).
       * Fix: Include onKeyBinding, onDirectEdit, and onDelete in the dependency array.


   * React Stale Closures in `useDiagramKeyBinding`:
      The useEffect in useDiagramKeyBinding.ts is missing several critical dependencies in its array: store, getMousePosition, viewport, elementIds, editingContextId, diagramId, invokeSingleClickTool, addErrorMessage, and getTargetObjectId.
  This risks executing logic with stale state.
       * Rule: react-frontend-reviewer rule 5 (Effect Dependencies).
       * Fix: Ensure all variables used inside useEffect are listed in the dependency array.


   * Java Null Safety in DTO Builder:
      In packages/core/backend/sirius-components-collaborative/src/main/java/org/eclipse/sirius/components/collaborative/dto/KeyBinding.java, the Builder.key() method does not check for null.
       * Rule: java-backend-reviewer rule 6 (No Null Values In Builders).
       * Fix:


   1     public Builder key(String key) {
   2         this.key = Objects.requireNonNull(key); // Add check
   3         return this;
   4     }

  </details>

  <details>


   * Type Duplication (DRY Violation): The interface GQLSingleClickOnDiagramElementTool and GQLKeyBinding are defined identically in both packages/palette/.../Palette.types.ts and packages/diagrams/.../Palette.types.ts. You should centralize these in the palette package and import them in the diagrams package to avoid maintenance overhead and potential desynchronization.
   * UX: Over-aggressive `event.preventDefault()`: The onKeyBinding function calls event.preventDefault() for any key press that isn't a lone modifier or in a text field, regardless of whether a matching tool exists. This will break standard browser shortcuts (e.g., Ctrl+P, Ctrl+F). You should only call preventDefault() if a tool is actually matched in the palette.
   * Missing Modifier Enforcement: The CHANGELOG states that a key binding needs at least one modifier (Control, Alt, or Meta), but this is not enforced in the frontend logic or the Java DTOs. Consider adding a validation check to ensure tools aren't accidentally triggered by single key presses (like 'e').
   * Memoization:
       * In useMousePosition.ts, getMousePosition should be wrapped in useCallback.
       * In useDiagramKeyBinding.ts, elementIds should be wrapped in useMemo to prevent unnecessary re-runs of the usePaletteContents hook due to reference changes.
   * Java Naming: In KeyBinding.java, the isCtrl, isMeta, isAlt boolean fields in the record are fine, but ensure the builder setters follow the same naming convention consistently.

  </details>

  <details>


   * Excellent Test Coverage: Including both Playwright E2E tests and Spring Boot integration tests for the palette controller is great practice.
   * Clean EMF Integration: The addition of KeyBinding to the core view.ecore and its integration into the generated builders is architecturally sound.
   * GraphQL Standards: Proper use of the GQL prefix for frontend types and matching naming conventions for the schema.
   * Encouraging Tone: The use of tooltips to discover key bindings is a very user-friendly addition.


  </details>

packages/core/frontend/sirius-components-core/src/mouse/useMousePosition.ts
```diff
@@ -0,0 +1,37 @@
...
+    const updateMousePosition = (event: MouseEvent) => {
+      mousePositionRef.current.x = event.clientX;
+      mousePositionRef.current.y = event.clientY;
+    };
+    window.addEventListener('mousemove', updateMousePosition);
+
+    return () => {
+      window.removeEventListener('mousemove', updateMousePosition);
+    };
+  });
```
sbegaudeau: `useEffect` without a dependency array is quite dangerous, do we need to add and remove a listener anytime a component using this hook is re-rendered?

packages/core/backend/sirius-components-collaborative/src/main/java/org/eclipse/sirius/components/collaborative/dto/KeyBinding.java
```diff
@@ -0,0 +1,77 @@
...
+
+        public Builder isAlt(boolean isAlt) {
+            this.isAlt = isAlt;
+            return this;
+        }
+
+        public Builder key(String key) {
+            this.key = key;
+            return this;
+        }
```
sbegaudeau, line 71: Objects.requireNonNull

packages/palette/frontend/sirius-components-palette/src/Palette.ts
```diff
@@ -28,3 +28,24 @@ export const isPaletteDivider = (entry: GQLPaletteDivider): entry is GQLToolSect
...
+  let tooltip: string = tool.label;
+  if (isSingleClickOnDiagramElementTool(tool) && tool.keyBindings.length > 0) {
```
sbegaudeau: > `isSingleClickOnDiagramElementTool`

You can't add here more code that depends on diagram. I won't block the PR just for that but, keep in mind that future contributions that may want to provide a palette in a representation that is not a diagram will have to rework this piece of code.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
```diff
@@ -0,0 +1,155 @@
...
+      }
+
+      setState((prevState) => ({
+        ...prevState,
+        currentEvent: event,
+      }));
+
+      // We have to call preventDefault here and not in the useEffect (once we are sure there is a tool for the key binding):
+      // the useEffect happens after the event has been propagated, hence calling preventDefault inside it doesn't do anything.
+      event.preventDefault();
```
sbegaudeau: Are you sure that we want to stop EVERY SINGLE browser event triggered by key bindings? On my machine, it kills every single browser shortcut apart from a couple hardcode ones like new tab or new window. We may have to do this in one method instead of two because of this.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/representations/EntityNodeDescriptionProvider.java
sbegaudeau: Yes it is an issue, we won't start creating conflicts with basic browser features.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/representations/EntityNodeDescriptionProvider.java
```diff
@@ -232,9 +233,9 @@ private EdgeTool entitySupertypeEdgeTool(NodeDescription edgeTargetNodeDescripti
...
+        var booleanCreationTool = this.entityAttributeCreationTool("Boolean", "aql:'/icons/full/obj16/BooleanAttribute.svg'", "newBoolean", "aql:domain::DataType::BOOLEAN", "b");
+        var numberCreationTool = this.entityAttributeCreationTool("Number", "aql:'/icons/full/obj16/NumberAttribute.svg'", "newNumber", "aql:domain::DataType::NUMBER", "n");
```
sbegaudeau: I somehow missed that we have here Ctrl+n which opens an entire new browser window. This is unusable 😄

packages/palette/frontend/sirius-components-palette/src/tool-list-item/ToolListItem.tsx
```diff
@@ -57,8 +58,10 @@ export const ToolListItem = ({ tool, disabled, onToolClick }: ToolListItemProps)
     onToolClick(tool);
   };
 
+  const tooltip = getToolTooltip(tool);
```
sbegaudeau: I did not check the code in action previously but we don't to display a tooltip, we want the same thing as in any other application. We can merge this as a first version but it will have to be removed.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
```diff
@@ -0,0 +1,155 @@
...
+    x = (mouseX - (bounds?.left ?? 0) - viewportX) / viewportZoom;
+    y = (mouseY - (bounds?.top ?? 0) - viewportY) / viewportZoom;
+  }
+  return { x, y };
+};
+
+const toolHasKeyBinding = (tool: GQLSingleClickOnDiagramElementTool, event: React.KeyboardEvent<Element>) => {
+  return tool.keyBindings.some(
+    (keyBinding) =>
+      keyBinding.key === event.key &&
```
sbegaudeau, line 55: You can't make it work by just using the key on the event, otherwise pretty much nothing will work for complex events. I have tested various use cases to challenge this code and this fails to match the value typed in tons of situations. Remember the issue we had matching `Shift` where we ended up saying that people would just have to type the uppercase version of the letter? Well it's worse than that, any letter that can be transformed by a key binding is transformed and can thus not be used.

Without [`navigator.keyboard`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/keyboard), I'm not even sure that we can make it work in all use cases. Funnily enough, this is an experimental API which of course is not available everywhere. (edit: even with this API it wouldn't work in various use cases)

In fact, matching a predefined keybinding with an event seems to be opening a sizable can of worms. There doesn't seem to be any API that can reliably tell you that someone has typed "Ctrl+Alt+a" / "Command+Alt+a" in every browser and in every keyboard layout (qwerty vs azerty).

To sum up, it works, as in "the code is connected",  but it will fail for tons of users in various browsers for some key bindings. Between existing browser / OS features and this, I'm pretty sure for example that it would fail in more than 50% of the cases on macOS (I can't speak for other operating systems) because we are not matching events properly and I'm not even sure we could do so properly.

In my tests, I ended up trying to match event keys like `ß`, `æ`, etc...

sbegaudeau (dismissed):
I've fixed some basic issues but we will need to improve both:

- The display of the key binding to match what is being done in every application
- The matching of the key binding because right now it doesn't work for tons of use cases

I'll rebase and merge it as-is for a first version

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/key-binding/useDiagramKeyBinding.ts
sbegaudeau: And the API we would need is not experimental as in "it works in one browser but will soon be available in other browsers", it's experimental as in "it works in one browser and other browsers are raised objections to the shape of this API for privacy reasons and refuse to implement it as is".

Fun fact, even web app developers at Adobe have the same kind of problems we do, because you can't match arbitrary key bindings against any OS, browser and keyboard layout. So it will be merged, and will fail spectacularly when real users will try to use it for anything more than some very specific keybindings for a specific subset of their end users...

sbegaudeau: Even MUI is facing the [same issue](https://github.com/mui/mui-x/pull/14220#discussion_r1720441510). If anyone wants to have fun, you can start by reading [this](https://github.com/w3c/uievents/issues/377) to understand the scope of the issue.

sbegaudeau:
I'll probably wait for a couple of days to think about this a bit more before considering merging it. We may have to limit key bindings to Ctrl/Meta + ASCII characters, I'm more and more convinced that Alt cannot be supported (for the same reasons as Shift) and we couldn't support anything more than basic ASCII characters.

sbegaudeau (changes requested):
@gdaniel After some consideration, here is what we will do. Please add some documentation (in a new file `doc/specifier/representation-diagram.adoc`) to indicate that specifiers will have to carefully think about the key bindings that they will provide and that they will have to consider how both shift and alt transform the input typed by the end user. You can add the links that I've mentioned in the comments of the PR to help explain the issue.

There's also one last thing that I'll need you to modify in the PR, the display of the key bindings in the palette. Please make sure that the first available key binding is visible just like in any other application (on the right of the tool, in light grey like in the screenshot below)

](https://docs-assets.developer.apple.com/published/1196662c916a44013329c4c6a1ba03d4/components-the-menu-bar-intro%402x.png)

gdaniel:
I have added some documentation in `doc/specifier/representation-diagram.adoc` as suggested.
I also reworked the way we display the key bindings in the palette, it is now more aligned with what you want (note that the meta icon is used for the meta modifier to look like on your screenshot, but plain text is used for other modifiers).

Note that I also removed the key binding from the tooltip on quick tools (as a result the key bindings for quick tools aren't visible anywhere anymore in the UI).

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/representations/EntityNodeDescriptionProvider.java
gdaniel: I changed it to `Ctrl+i` (integer).
This is not a good solution, but to be honest I don't know what to do with that, because we also learned that `alt` is a dangerous option.
At least the key binding does what it is supposed to do.

packages/palette/frontend/sirius-components-palette/src/tool-list-item/ToolListItem.tsx
gdaniel: Fixed, the first key binding is now displayed in the palette

packages/core/backend/sirius-components-collaborative/src/main/java/org/eclipse/sirius/components/collaborative/dto/KeyBinding.java
gdaniel: done

packages/palette/frontend/sirius-components-palette/src/tool-list-item/ToolListItem.tsx
```diff
@@ -48,7 +50,33 @@ const useStyle = makeStyles()((theme) => ({
...
+      <div className={classes.keyBinding} data-testid={`key-binding-${tool.label}`}>
+        {keyBinding.isCtrl ? 'Ctrl+' : ''}
+        {keyBinding.isMeta ? <KeyboardCommandKeyIcon /> : null}
+        {keyBinding.isAlt ? 'Alt+' : ''}
+        {keyBinding.key}
+      </div>
+    );
+  }
+  return null;
+};
```
sbegaudeau: A function which takes some data and return some JSX is a React component. I'll update it

sbegaudeau (approved):
I'll merge it after the build

PR #5986, opened 2025-12-18
[5982] Add the ability to trigger context menu entries with key bindings

### LINKED ISSUE #5982: Add the ability to trigger context menu entries using key bindings

Similarly to #1626, it should be possible for a specifier to define a KeyBinding (key along with `ctrl`, `alt`, `meta`) for a context menu entry. 
As for diagrams, we want to:

- Display the key binding in the contextual menu
- Support multiple key bindings for an entry

Executing an entry via its key binding should be able to trigger impact analysis.

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/5982
Fixes #5982 

Based on #5968 (this PR adds `KeyBinding` in the view DSL, which is reused here)

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/DefaultMenuItem.tsx
```diff
@@ -10,166 +10,42 @@
...
-              fetchTreeItemContextMenuEntryData(treeItemId: $treeItemId, menuEntryId: $menuEntryId) {
-                urlToFetch
-                fetchKind
-              }
-            }
-          }
-        }
-      }
-    }
+import { useInvokeContextMenuEntry } from './useInvokeContextMenuEntry';
```
gdaniel, line 20: The code to invoke a context menu entry is now in a hook to make it reusable here and in `useTreeKeyBinding`

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/TreeItemContextMenu.tsx
```diff
@@ -64,7 +65,18 @@ export const TreeItemContextMenu = ({
...
+
+  useEffect(() => {
+    getContextMenuEntries({
+      variables: {
+        editingContextId,
+        representationId: treeId,
+        treeItemId: item.id,
+      },
+    });
+  }, []);
```
gdaniel: As for diagrams, the query to get the context menu entries now has to be lazy to be able to trigger it only if the user entered a valid key combination.

integration-tests-playwright/playwright/e2e/key-bindings.spec.ts
```diff
@@ -53,3 +53,31 @@ test.describe('diagram - key bindings', () => {
     await expect(page.getByTestId('IconLabel - newBoolean')).toBeAttached();
   });
 });
+
+test.describe('explorer - key bindings', () => {
```
gdaniel: I don't know if this test should be here (because it targets the same feature - key bindings), or if I should split the tests of each view as it is proposed in [this PR](https://github.com/eclipse-sirius/sirius-web/pull/5751/files#diff-9dc232adda9b8842b0d48fd1afbf78a3a93b87da339ed02b85ce0c136d515806)

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/studio/services/representations/DomainViewTreeDescriptionProvider.java
```diff
@@ -291,12 +292,29 @@ private List<TreeItemContextMenuEntry> getContextMenuEntries() {
...
                 .kind(FetchTreeItemContextMenuEntryKind.OPEN)
+                .keyBindings(new KeyBindingBuilder()
+                                .ctrl(true)
+                                .key("h")
+                                .build(),
+                        new KeyBindingBuilder()
+                                .meta(true)
+                                .key("h")
+                                .build()
+                )
```
gdaniel, line 304: As for diagrams, I added key bindings on existing entries

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/useInvokeContextMenuEntry.ts
```diff
@@ -0,0 +1,214 @@
...
+  const { addErrorMessage } = useMultiToast();
+
+  const { data: treeItemContextMenuOverrideContributions } = useData<TreeItemContextMenuOverrideContribution[]>(
+    treeItemContextMenuEntryOverrideExtensionPoint
+  );
+
+  const [state, setState] = useState<UseInvokeContextMenuEntryState>({
+    currentEntry: null,
+    onEntryExecution: () => {},
+    onClick: () => {},
```
gdaniel: See my general comment on the PR for more details on this

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/useInvokeContextMenuEntry.ts
```diff
@@ -0,0 +1,214 @@
...
+  ) => {
+    const menuEntryIsOverridden = treeItemContextMenuOverrideContributions.some((contribution) =>
+      contribution.canHandle(menuEntry)
+    );
+
+    if (menuEntryIsOverridden) {
+      // Do not attempt to invoke an overridden menu entry with the regular entry invocation mutations:
+      // overridden entries define their own behavior and we cannot assume they rely on these mutations.
+      return;
+    }
```
gdaniel, line 58: See my general comment on the PR for more details on this

gdaniel (commented):
I had to implement some dubious things in `useInvokeContextMenuEntry` which indicate to me that some work is needed on the lifecycle of menu entries and how we define menu overrides.

I had to store the `onClick` function in the state because fetch entries need to execute another query to get the URL to fetch, and navigate to said URL once it is received. This cannot work if the context menu is closed.
To me this indicates that the execution of the entry is too coupled to the menu, this became particularly visible with key bindings, where I had to [provide a dummy function](https://github.com/eclipse-sirius/sirius-web/pull/5986/changes#diff-6f640823eb7c1697c64fac514306e32f8ad112c5f7bfd775ca70a6803d0babeaR53) for the `onClick`. 
I think we should have something like we do in diagrams: a context that is responsible of entry execution, regardless of whether the menu is open or not.

I also had to check if the entry to invoke is overridden, and if so, not invoke it. I didn't find a way to invoke overridden entries in a generic way, because these entries are provided as components that embed their execution logic. There is no menu when we activate a key binding, so these component don't even exist, and we can't fake their invocation with a click on the component for example.
Anyways, to me this points out an issue in the way we define menu entry overrides: I don't think we should provide components, but a function that is executed when the menu entry is clicked. This way we can display these entries as `DefaultMenuItem`, and bind the `onClick` on the function. In the context of key bindings, we could simply invoke this function without worrying about the component being here or not.

I didn't want to work on these issues in this PR, because:

1. I want to validate the solutions I propose
2. I want to keep this PR relatively small, and I think each solution deserves its own PR

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/trees/TreeItemContextMenuControllerTests.java
```diff
@@ -161,20 +169,15 @@ public void givenPapayaProjectWhenTheContextMenuActionsAreRequestedOnAnObjectThe
     @DisplayName("Given a studio, when the context menu actions are requested on an Entity, then the correct actions are returned")
     public void givenAStudioWhenTheContextMenuActionsAreRequestedOnAnEntityThenTheCorrectActionsAreReturned() {
         // 1- retrieve the tree description id of the DSL Domain explorer example
-        var explorerDescriptionId = new AtomicReference<String>();
-
-        Map<String, Object> explorerVariables = Map.of(
-                "editingContextId", StudioIdentifiers.SAMPLE_STUDIO_EDITING_CONTEXT_ID.toString()
-        );
+        Map<String, Object> explorerVariables = Map.of("editingContextId", StudioIdentifiers.SAMPLE_STUDIO_EDITING_CONTEXT_ID);
```
gdaniel, line 172: I reworked a bit this test to make checkstyle happy.
If needed I can rework it more to split it into smaller pieces

packages/trees/frontend/sirius-components-trees/src/trees/useTreeKeyBinding.ts
```diff
@@ -0,0 +1,105 @@
...
+          (state.currentEvent.ctrlKey ? 'CTRL + ' : '') +
+          (state.currentEvent.metaKey ? 'META + ' : '') +
+          (state.currentEvent.altKey ? 'ALT + ' : '') +
+          state.currentEvent.key;
+        const entryLabels = entryToExecute.map((entry) => entry.label).join(', ');
+        addErrorMessage(`Found multiple entries associated to ${keysPressed}: ${entryLabels}`);
+      }
+      // There is no entry associated to the pressed keys, don't do anything
+      // In any case, reset the current event, we don't need it anymore and we don't want to reuse it.
+      setState((prevState) => ({
```
gdaniel, line 65: I had to do the same thing I did for [diagrams](https://github.com/eclipse-sirius/sirius-web/pull/5968#discussion_r2697912879) here to ensure the entry is not triggered when it shouldn't.
In this case, the scenario to reproduce is to comment the `setState`, trigger a tool via a key binding, then expand another item -> the tool is triggered another time.

sbegaudeau (commented):
I've rebased the PR to make it reviewable

sbegaudeau:

  Pull Request Status: This PR is ready for review. It contains comprehensive changes covering the GraphQL schema, backend providers, and frontend interaction logic, supported by both end-to-end and integration tests.


  Summary: This PR successfully addresses issue #5982 by introducing key binding support for tree context menu entries. It aligns tree functionality with the existing capabilities in diagrams, allowing specifiers to define shortcuts that
  end-users can trigger directly from the keyboard when a tree item is focused.

  <details>


   - Keyboard Event Swallowing (Major UX Issue): In useTreeKeyBinding.ts, the onKeyBinding hook calls event.preventDefault() on all key events whenever a tree item is focused. This will break standard browser and system shortcuts (e.g., F5 for reload, Ctrl+C/Ctrl+V for copy-paste, Cmd+R on Mac) that are not explicitly defined as Sirius key bindings.
     - Fix: Only call event.preventDefault() if a matching key binding is found in the contextMenuEntries.
     - Rule enforced: General UX and accessibility standards.


   - Deprecated ErrorPayload.message (Rule Violation): In useInvokeContextMenuEntry.ts and its associated types, the code relies on the message property of ErrorPayload. This property is explicitly marked as @deprecated in the GraphQL schema (referenced in issue #6194).
     - Actionable Fix: Update GQLErrorPayload to include messages: GQLMessage[] and use addMessages(payload.messages) from useMultiToast.


   1   // useInvokeContextMenuEntry.ts
   2   if (isErrorPayload(invokeSingleClickTreeItemContextMenuEntry)) {
   3     addMessages(invokeSingleClickTreeItemContextMenuEntry.messages);
   4   }
     - Rule enforced: react-frontend-reviewer Rule 5 ("Error Payload Message: We MUST not add new code depending on the property message from ErrorPayload. We MUST instead rely on messages").

  </details>

  <details>


   - Internationalization (i18n): The getEntryTooltip function in DefaultMenuItem.tsx uses hardcoded strings for "CTRL +", "META +", and "ALT +". These should be retrieved from an internationalization service or utility to support multiple locales.
   - UI Discoverability: Keybindings are currently only shown in a tooltip on hover. In the palette, they are immediately visible. Consider displaying the shortcut next to the label in the MenuItem (e.g., using a secondary Typography component) for better discoverability.
   - Component Reuse: The logic and styling for rendering keybindings (including the use of icons for the Meta/Command key) are duplicated from packages/palette. Consider moving the KeyBinding component to packages/core to ensure a consistent visual language across the platform.
   - Performance - Lazy Loading: Triggering a GraphQL query (useContextMenuEntries) upon every key press (when focused) might introduce a perceived delay if the cache is empty. While acceptable for trees due to their dynamic nature, ensure that the useQuery fetch policy is optimized for responsiveness.

  </details>

  <details>


   - Comprehensive Testing Strategy: The addition of a Playwright end-to-end test (key-bindings.spec.ts) and a backend integration test (TreeItemContextMenuControllerTests.java) provides excellent coverage and verification of the new feature.
   - Clean Hook Refactoring: Encapsulating the context menu invocation logic into useInvokeContextMenuEntry is a significant architectural improvement over the previous implementation in DefaultMenuItem, making the code much more readable and maintainable.
   - Type Safety: The PR maintains high standards for TypeScript typing across all new hooks and components, strictly following the GQL prefix convention and separating types into dedicated files.


  </details>

gdaniel:
I reworked the PR based on the discussions on the related PR for diagrams.
Updates:

- The key binding is now visible in the context menu, not in a tooltip
- The `KeyBinding` component has been moved to `sirius-components-core` as suggested by Gemini
- I also used `messages` instead of `message` in `useInvokeContextMenuEntry`
- `event.preventDefault()` has been removed from `useTreeKeyBinding`, as a result, I had to change the key bindings of the tree items to something that actually works with my browser (and is absolutely not intuitive to the end user 😑)

I also noted that key bindings do not trigger the corresponding entry if the contextual menu is open (while it does on diagrams when the palette is open). This is due to the way we retrieve the selected tree item:

```typescript
const treeItemId = (document.activeElement as HTMLElement).dataset.treeitemid || null;
``` 

Which returns null as soon as the contextual menu is open.

packages/core/frontend/sirius-components-core/src/key-binding/KeyBinding.tsx
```diff
@@ -18,6 +18,7 @@ import { KeyBindingProps } from './KeyBinding.types';
 const useKeyBindingStyle = makeStyles()((theme) => ({
   keyBinding: {
     flex: '0 0 auto',
+    paddingLeft: theme.spacing(1),
```
sbegaudeau: Why? You can't make any hypothesis on where this will be used. If a gap is needed, the parent should add it.

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/DefaultMenuItem.tsx
sbegaudeau: It makes sense but I suspect that this hook has more responsibilities than what we may need

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/useInvokeContextMenuEntry.ts
sbegaudeau: My main issue here is that this hook is doing the job of at least three, four or event five hooks and is thus tied to a lot of responsibilities.

- There should be a dedicated hook to execute the query `getFetchTreeItemContextMenuEntryDataQuery` without any special behavior in the useEffect
- There should be a dedicated hook to execute the mutation `invokeSingleClickTreeItemContextMenuEntryMutation` without any special behavior in the useEffect
- There should be at least one but maybe two places (dedicated hooks or just in their respective components) where these first two hooks would be reused.

I'm not sure that extracting 100% of the behavior of `DefaultMenuItem` into a hook and then try to make it reusable with a hack like this is a viable option.

packages/trees/frontend/sirius-components-trees/src/trees/Tree.tsx
```diff
@@ -42,79 +43,74 @@ export const Tree = ({
...
-      return () => {
-        element.removeEventListener('keydown', downHandler);
-      };
-    }
-    return () => {};
-  }, [treeElement, onExpandedElementChange]);
+  const handleKeyDown = (event: React.KeyboardEvent<Element>) => {
+    onTreeNavigation(event);
+    onKeyBinding(event);
+  };
```
sbegaudeau, line 108: Doesn't this change the lifecycle of the tree navigation from being executed after the selection is updated to being executed before? This may not be an issue, I've not tested it but I wonder why would that change be necessary?

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/tree/ViewTreeItemContextMenuEntryProvider.java
```diff
@@ -33,6 +34,7 @@
 import org.eclipse.sirius.components.view.emf.IViewRepresentationDescriptionSearchService;
 import org.eclipse.sirius.components.view.emf.ViewRepresentationDescriptionPredicate;
 import org.eclipse.sirius.components.view.emf.api.IViewAQLInterpreterFactory;
+import org.eclipse.sirius.components.view.emf.diagram.tools.api.IKeyBindingConverter;
```
sbegaudeau: Don't depend on diagram specific code in the tree part, move it to a more generic package

sbegaudeau (changes requested):
I have not tested manually for now.

sbegaudeau:


  Pull Request Status: The pull request #5986 is currently OPEN. All CI checks have PASSED, including Continuous Integration and Metadata Review.


  Summary: This PR introduces the ability to trigger tree context menu entries via key bindings, mirroring functionality already available for diagram tools. It involves changes across the backend (EMF models, DTOs, GraphQL schema, and
  providers) and the frontend (new hooks for key binding handling and refactored menu item logic). While the core functionality is implemented and tested, there are several architectural and compliance issues that need to be addressed.

  <details>


   * Deprecated ErrorPayload Usage (React Rule 5): In useInvokeContextMenuEntry.ts, the code incorrectly relies on the deprecated message property of ErrorPayload. You MUST use the messages property instead.


   1     // useInvokeContextMenuEntry.ts
   2     if (isErrorPayload(invokeSingleClickTreeItemContextMenuEntry)) {
   3       // VIOLATION: Relying on deprecated 'message'
   4       // addErrorMessage(invokeSingleClickTreeItemContextMenuEntry.message);
   5
   6       // FIX: Use 'messages' or a helper that handles the list of messages
   7       addErrorMessage(invokeSingleClickTreeItemContextMenuEntry.messages[0]?.body);
   8     }
   * Aggressive event.preventDefault() (UX/Bug): In useTreeKeyBinding.ts, event.preventDefault() is called for every key press as long as a tree item is focused. This breaks standard keyboard navigation (e.g., Delete for deletion, F2 for rename, or alpha-numeric searching) when no modifier is used.


   1     // useTreeKeyBinding.ts
   2     const hasModifier = event.altKey || event.ctrlKey || event.metaKey;
   3     if (treeItemId && hasModifier) { // Fix: Only prevent default if it's potentially a key binding
   4       event.preventDefault();
   5     }
   * Inline Styles Violation (React Rule 4): DefaultMenuItem.tsx contains hardcoded inline styles. Strictly forbid the use of standard React style objects.


   1     // DefaultMenuItem.tsx
   2     // VIOLATION:
   3     // <div style={{ marginRight: '16px' }} />
   4
   5     // FIX: Use the 'sx' prop
   6     <Box sx={{ marginRight: 2 }} />
   * Stale Closure Risk in Hooks: In useInvokeContextMenuEntry.ts, the useEffect that handles the lazy query result depends on data and error but uses state.onClick. Since state is not in the dependency array, it may close over a stale
     version of the onClick callback if invokeContextMenuEntry is called multiple times.
   * Component Lifecycle Coupling: As noted in your comments, useInvokeContextMenuEntry is currently tied to the DefaultMenuItem lifecycle. For Fetch entries, if the menu closes before the query returns, the logic will not execute. While you noted this for future PRs, it remains a critical architectural weakness in the current implementation.

  </details>


  <details>


   * Missing Type Prefixes (React Rule 2): In useInvokeContextMenuEntry.types.ts, the interface  FetchTreeItemContextMenuEntry is missing the mandatory GQL prefix. It should be renamed to  GQLFetchTreeItemContextMenuEntry.
   * Java Null Safety (Java Rule 2): While ViewTreeItemContextMenuEntryProvider correctly uses Objects.requireNonNull(), ensure that all new records and DTOs also enforce non-nullability where appropriate, especially since these are projected into the GraphQL layer.
   * Redundant .toString() removal: In TreeItemContextMenuControllerTests.java, you've removed several .toString() calls on UUID identifiers. While cleaner if the runners now support UUID, ensure this doesn't lead to ambiguous method calls if multiple overloads exist.

  </details>


  <details>


   * Comprehensive Testing: The addition of a Playwright end-to-end test (key-bindings.spec.ts) is excellent and effectively validates the feature across different platforms (Ctrl/Meta).
   * Architecture Alignment: The reuse of the KeyBinding concept from the diagram module in the tree module shows a great commitment to system-wide consistency and architectural integrity.
   * Clean Refactoring: Moving the execution logic out of DefaultMenuItem into a dedicated useInvokeContextMenuEntry hook significantly improves the separation of concerns, even with the acknowledged lifecycle challenges.


  </details>

packages/trees/frontend/sirius-components-trees/src/trees/Tree.tsx
gdaniel: I don't see how this changes the lifecycle of the tree navigation, from what I understand it may change when the event listener is attached to the tree, but it doesn't change what happens when the handler is executed.
I am not sure I understand what you mean by "after the selection".

I may be missing something though, if you can give me more information I'll take a look. In the meantime I completely removed the `ref` since it is not used anymore in the current implementation.

For the record I made this change to avoid mixing dom-level event handlers (for the tree navigation) and react event listeners (for the key bindings), which looks like an [error-prone thing to do](https://fortes.com/2018/mixing-react-and-dom-events/).

packages/view/backend/sirius-components-view-emf/src/main/java/org/eclipse/sirius/components/view/emf/tree/ViewTreeItemContextMenuEntryProvider.java
gdaniel: Done, I move `IKeyBindingConverter` and `KeyBindingConverter` in `org.eclipse.sirius.components.view.emf`

packages/core/frontend/sirius-components-core/src/key-binding/KeyBinding.tsx
gdaniel: Fixed

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/DefaultMenuItem.tsx
```diff
@@ -10,177 +10,42 @@
...
+      onClick={() => invokeContextMenuEntry(editingContextId, treeId, item.id, entry, onClick)}
       data-testid={`context-menu-entry-${entry.label}`}
       disabled={readOnly}
       aria-disabled>
       <ListItemIcon>
         {entry.iconURL.length > 0 ? (
           <IconOverlay iconURLs={entry.iconURL} alt={entry.label} title={entry.label} />
         ) : (
-          <div style={{ marginRight: '16px' }} />
+          <Box sx={{ marginRight: '16px' }} />
```
gdaniel: I also integrated the Gemini suggestion on using a `Box` with `sx` instead of a styled `Div`

gdaniel:
I did most of the refactoring you suggested.

I split `useInvokeContextMenuEntry` into the following hooks:

- `useInvokeSingleClickContextMenuEntry`: executes a single click entry (with or without impact analysis)
- `useInvokeFetchContextMenuEntry`: executes a fetch entry (relies on `useFetchContextMenuEntryData`)
- `useFetchContextMenuEntryData`: retrieves the data for to execute a fetch entry

Note that both `useInvokeXXX` hooks require an `onClick` function used to close the contextual menu. This is needed because the menu needs to stay open for the entire execution of the tool, which involves retrieving the fetch data (for fetch entries), and retrieve the impact analysis report (for single click entries with impact analysis).

I also kept `useInvokeContextMenuEntry`, which is now just a wrapper that calls `useInvokeFetchContextMenuEntry` or `useInvokeSingleClickContextMenuEntry` based on the type of the entry. I could remove it, but I didn't see the point: I'd have to basically duplicate the code in `DefaultMenuItem` and `useTreeKeyBinding`.

So, in the end, the refactoring didn't get rid of the `onClick` hack, and the only thing I see to get rid of it is to invoke the context menu entries in a dedicated context instead of the menu. I don't think it should be done in this MR, because it is quite far from key bindings, as it addresses an issue that was already there when executing context menu entries (it was just less visible since everything was in the same file). Anyways, let me know what you think/what I should do on that.

packages/trees/frontend/sirius-components-trees/src/treeitems/context-menu/DefaultMenuItem.tsx
sbegaudeau: You should also use the theme to retrieve the gap size instead of hardcoding 16px :)

sbegaudeau (approved):
I'll merge it after the build

PR #4463, opened 2025-01-23
[4462] Add support for library visualization

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/4462

This PR contains the following features (the other features related to library publication will be contributed in separated PRs to try to help with the reviews)
- New database table to store libraries and their dependencies
- New page `/libraries` to visualize the libraries stored in the database
- New GraphQL query to list libraries
- New mutation to publish libraries from a project
- First version of studio publication

# Pull request template

## General purpose
What is the main goal of this pull request?
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management
- [ ] Has the pull request been added to the relevant project and milestone? (Only if you know that your work is part of a specific iteration such as the current one)
- [ ] Have the `priority:` and `pr:` labels been added to the pull request? (In case of doubt, start with the labels `priority: low` and `pr: to review later`)
- [ ] Have the relevant issues been added to the pull request?
- [ ] Have the relevant labels been added to the issues? (`area:`, `difficulty:`, `type:`)
- [ ] Have the relevant issues been added to the same project and milestone as the pull request?
- [ ] Has the `CHANGELOG.adoc` been updated to reference the relevant issues?
- [ ] Have the relevant API breaks been described in the `CHANGELOG.adoc`? (Including changes in the GraphQL API)
- [ ] In case of a change with a visual impact, are there any screenshots in the `CHANGELOG.adoc`? For example in `doc/screenshots/2022.5.0-my-new-feature.png`

## Architectural decision records (ADR)
- [ ] Does the title of the commit contributing the ADR start with `[doc]`?
- [ ] Are the ADRs mentioned in the relevant section of the  `CHANGELOG.adoc`?

## Dependencies
- [ ] Are the new / upgraded dependencies mentioned in the relevant section of the `CHANGELOG.adoc`?
- [ ] Are the new dependencies justified in the `CHANGELOG.adoc`?

## Frontend

This section is not relevant if your contribution does not come with changes to the frontend.

### General purpose
- [ ] Is the code properly tested? (Plain old JavaScript tests for business code and tests based on React Testing Library for the components)

### Typing
We need to improve the typing of our code, as such, we require every contribution to come with proper TypeScript typing for both changes contributing new files and those modifying existing files.
Please ensure that the following statements are true for each file created or modified (this may require you to improve code outside of your contribution).

- [ ] Variables have a proper type
[...]

PR #6007, opened 2025-12-29
[5585] Replace use of foreign keys between different bounded contexts with proper events

ntinsalhi-obeo:
# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/representationdata/services/RepresentationMetadataDeletionService.java
+++ b/packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/representationdata/services/RepresentationMetadataDeletionService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2024 Obeo.
+ * Copyright (c) 2024, 2026 Obeo.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
@@ -56,5 +56,14 @@ public class RepresentationMetadataDeletionService implements IRepresentationMet
             result = new Failure<>(this.messageService.notFound());
         }
         return result;
+    }
+
+    @Override
+    public IResult<Void> deleteRepresentationMetadata(ICause cause, UUID semanticDataId) {
+        var allRepresentationMetadata = this.representationMetadataRepository.findAllRepresentationMetadataBySemanticDataId(semanticDataId);
+        allRepresentationMetadata.forEach(representationMetadata -> representationMetadata.dispose(cause));
+        this.representationMetadataRepository.deleteAll(allRepresentationMetadata);
+
+        return new Success<>(null);
     }
 }
--- a/packages/sirius-web/backend/sirius-web-tests-data/src/main/resources/sirius-web-scripts/cleanup.sql
+++ b/packages/sirius-web/backend/sirius-web-tests-data/src/main/resources/sirius-web-scripts/cleanup.sql
@@ -8,4 +8,5 @@ DELETE FROM representation_content;
 DELETE FROM representation_metadata;
 DELETE FROM image;
 DELETE FROM nature;
+DELETE FROM project_image;
 DELETE FROM project;
--- a/packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/representations/RepresentationMetadataIntegrationTests.java
+++ b/packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/representations/RepresentationMetadataIntegrationTests.java
@@ -1,0 +1,87 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.web.application.controllers.representations;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Optional;
+
+import org.eclipse.sirius.web.AbstractIntegrationTests;
+import org.eclipse.sirius.web.data.PapayaIdentifiers;
+import org.eclipse.sirius.web.domain.boundedcontexts.representationdata.events.RepresentationMetadataDeletedEvent;
+import org.eclipse.sirius.web.domain.boundedcontexts.representationdata.services.api.IRepresentationMetadataSearchService;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.SemanticData;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.events.SemanticDataDeletedEvent;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.services.api.ISemanticDataDeletionService;
+import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.services.api.ISemanticDataSearchService;
+import org.eclipse.sirius.web.services.api.IDomainEventCollector;
+import org.eclipse.sirius.web.tests.data.GivenSiriusWebServer;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.data.jdbc.core.mapping.AggregateReference;
+import org.springframework.test.context.transaction.TestTransaction;
+import org.springframework.transaction.annotation.Transactional;
+
+/**
+ * Integration tests of the representation metadata controllers.
+ *
+ * @author ntinsalhi
+ */
+@Transactional
+@SuppressWarnings("checkstyle:MultipleStringLiterals")
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+public class RepresentationMetadataIntegrationTests extends AbstractIntegrationTests {
+
+    @Autowired
+    private IRepresentationMetadataSearchService representationMetadataSearchService;
+
+    @Autowired
+    private ISemanticDataSearchService semanticDataSearchService;
+
+    @Autowired
+    private ISemanticDataDeletionService semanticDataDeletionService;
+
+    @Autowired
+    private IDomainEventCollector domainEventCollector;
+
+    @BeforeEach
+    public void beforeEach() {
+        this.domainEventCollector.clear();
+    }
+
+    @Test
+    @GivenSiriusWebServer
+    @DisplayName("Given representations linked to semantic data, when a SemanticDataDeletedEvent is published, then the representation metadata is deleted")
+    public void givenRepresentationsLinkedToSemanticDataWhenASemanticDataDeletedEventIsPublishedThenTheRepresentationMetadataIsDeleted() {
+        Optional<SemanticData> optionalSemanticData = this.semanticDataSearchService.findById(PapayaIdentifiers.PAPAYA_EDITING_CONTEXT_ID);
+        assertThat(optionalSemanticData).isPresent();
+
+        var semanticDataRepresentationsMetadata = this.representationMetadataSearchService.findAllRepresentationMetadataBySemanticData(AggregateReference.to(PapayaIdentifiers.PAPAYA_EDITING_CONTEXT_ID));
+        assertThat(semanticDataRepresentationsMetadata).isNotEmpty();
+
+        this.semanticDataDeletionService.deleteSemanticData(null, optionalSemanticData.get().getId());
+
+        TestTransaction.flagForCommit();
+        TestTransaction.end();
+        TestTransaction.start();
+
+        assertThat(this.domainEventCollector.getDomainEvents()).anyMatch(SemanticDataDeletedEvent.class::isInstance);
+        assertThat(this.domainEventCollector.getDomainEvents()).anyMatch(RepresentationMetadataDeletedEvent.class::isInstance);
+
+        semanticDataRepresentationsMetadata = this.representationMetadataSearchService.findAllRepresentationMetadataBySemanticData(AggregateReference.to(PapayaIdentifiers.PAPAYA_EDITING_CONTEXT_ID));
+        assertThat(semanticDataRepresentationsMetadata).isEmpty();
+    }
+}
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -110,6 +110,9 @@ Downstream projects relying on this should switch to a better initialization or 
 The optional editing context id argument has been removed since it can be computed from the variable manager.
 - https://github.com/eclipse-sirius/sirius-web/issues/4860[#4860] [core] The `workbenchViewContributionExtensionPoint` frontend extension point has slightly changed since the `side` attribute has been removed and it is now managed directly by the workbench configuration.
 - https://github.com/eclipse-sirius/sirius-web/issues/5815[#5815] [diagram] `isLastNodeSelected` should be set when converting custom nodes.
+- https://github.com/eclipse-sirius/sirius-web/issues/5585[#5585] [sirius-web] Additional foreign keys set up between bounded contexts have been deleted.
+These should not have axisted in the first place and have now been replaced with transaction event listeners.
+As a consequence, additional domain events can now be fired while using Sirius Web giving downstream projects additional insights into the consequences of an action.
 
 
 === Dependency update
@@ -172,6 +175,7 @@ With this architecture, downstream projects can customize this behavior by provi
 With the workbench configuration, downstream projects can give the representation to open when the user open a project and in which side panel views will be available.
 In conjonction of the `workbenchViewContributionExtensionPoint` frontend extension point, downstream projects can add custom views with their workbench configuration.
 - https://github.com/eclipse-sirius/sirius-web/issues/5815[#5815] [diagram] Highlight the last selected elements with custom resize handles, this can be useful when executing a layout group tool.
+- https://github.com/eclipse-sirius/sirius-web/issues/5585[#5585] [sirius-web] Replace use of foreign keys between different bounded contexts with proper events
 
 
 === Improvements
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/images/listeners/ProjectImageCleaner.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/images/listeners/ProjectImageCleaner.java
@@ -10,35 +10,35 @@
  * Contributors:
  *     Obeo - initial API and implementation
  *******************************************************************************/
-package org.eclipse.sirius.web.application.semanticdata.listeners;
+package org.eclipse.sirius.web.application.images.listeners;
 
 import java.util.Objects;
 
-import org.eclipse.sirius.web.domain.boundedcontexts.projectsemanticdata.events.ProjectSemanticDataDeletedEvent;
-import org.eclipse.sirius.web.domain.boundedcontexts.semanticdata.services.api.ISemanticDataDeletionService;
+import org.eclipse.sirius.web.domain.boundedcontexts.project.events.ProjectDeletedEvent;
+import org.eclipse.sirius.web.domain.boundedcontexts.projectimage.services.api.IProjectImageDeletionService;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.transaction.event.TransactionalEventListener;
 
 /**
- * Service used to delete the semantic data of a project.
+ * Used to delete a project image.
  *
  * @author ntinsalhi
  */
 @Service
-public class SemanticDataCleaner {
+public class ProjectImageCleaner {
 
-    private final ISemanticDataDeletionService semanticDataDeletionService;
+    private final IProjectImageDeletionService projectImageDeletionService;
 
-    public SemanticDataCleaner(ISemanticDataDeletionService semanticDataDeletionService) {
-        this.semanticDataDeletionService = Objects.requireNonNull(semanticDataDeletionService);
+    public ProjectImageCleaner(IProjectImageDeletionService projectImageDeletionService) {
+        this.projectImageDeletionService = Objects.requireNonNull(projectImageDeletionService);
     }
 
     @Transactional(propagation = Propagation.REQUIRES_NEW)
     @TransactionalEventListener
-    public void onProjectSemanticDataDeletedEvent(ProjectSemanticDataDeletedEvent event) {
-        var semanticDataId = event.projectSemanticData().getSemanticData().getId();
-        this.semanticDataDeletionService.deleteSemanticData(event, semanticDataId);
+    public void onProjectDeletedEvent(ProjectDeletedEvent event) {
+        var projectId = event.project().getId();
+        this.projectImageDeletionService.deleteAllProjectImages(event, projectId);
     }
 }
--- a/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/representation/listeners/RepresentationMetadataCleaner.java
+++ b/packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/representation/listeners/RepresentationMetadataCleaner.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2025, 2026 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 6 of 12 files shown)


packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/library/events/LibraryDeletedEvent.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.time.Instant;
+import java.util.UUID;
+
+/**
+ * Event fired when a library is deleted.
+ *
+ * @author ntinsalhi
+ */
+public record LibraryDeletedEvent(
```
gdaniel: Something may have changed since the original specification of libraries, but at that time, libraries were explicitly immutable, and couldn't be deleted.

You can find some information in [this PR](https://github.com/eclipse-sirius/sirius-web/pull/4463), especially [this comment](https://github.com/eclipse-sirius/sirius-web/pull/4463#discussion_r1950958162) on the `LibraryDeletedEvent` and [this comment](https://github.com/eclipse-sirius/sirius-web/pull/4463#discussion_r1950934406) on library deletion.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/listeners/LibraryCleaner.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final ILibraryDeletionService libraryDeletionService;
+
+    public LibraryCleaner(ILibraryDeletionService libraryDeletionService) {
+        this.libraryDeletionService = Objects.requireNonNull(libraryDeletionService);
+    }
+
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    @TransactionalEventListener
+    public void onSemanticDataDeletedEvent(SemanticDataDeletedEvent event) {
```
gdaniel: (I am assuming we now want to support library deletion, or we want to implement the behavior of the foreign key, even if library deletion is not supported)

I think this should be the other way around: the foreign key allowed to cascade deletion on semantic data, not the opposite. Here you would delete a library if _any_ of its semantic data is deleted.

sbegaudeau: We will not support library deletion but we want to code what's necessary for us to keep the consistency of the system without the foreign key. We will have the same behavior as before this contribution. It doesn't change anything on our side.

> the foreign key allowed to cascade deletion on semantic data, not the opposite

Humm? How could that be possible, semantic data has no link whatsoever to libraries. It was always the library table that had a foreign key to the semantic data table. If the semantic data was deleted, then the library would be deleted too. Here we are just performing the same thing in code.

> Here you would delete a library if any of its semantic data is deleted

Hummmmmm? What are you talking about? There is only one "semantic data" per library:

```
@Table("library")
public class Library extends AbstractValidatingAggregateRoot<Library> implements Persistable<UUID> {
   @Column("semantic_data_id")
    private AggregateReference<SemanticData, UUID> semanticData;
}
```

If a row in the SemanticData table is deleted and if it matches the SemanticData of a library, then the library cannot exist anymore. The `SemanticDataDeletedEvent` is not about deleting some data from the editing context, it's about deleting the entire semantic data (i.e. nuking the whole editing context from the database). If we receive this event, there's nothing left.

So either there's something that I'm really not understanding about your comment or there's something you really did misunderstood from this contribution because:

- There is no way in Sirius Web to reach this part of the code using our user interface or HTTP/WebSocket API
- This is the same behavior as before
- Once this event is received, everything of value has already been deleted, we are just clearing the remaining metadata (namespace, name, version)

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/projectsemanticdata/listeners/ProjectSemanticDataCleaner.java
```diff
@@ -12,15 +12,15 @@
...
-import java.util.Objects;
-
 import org.eclipse.sirius.web.domain.boundedcontexts.project.events.ProjectDeletedEvent;
 import org.eclipse.sirius.web.domain.boundedcontexts.projectsemanticdata.services.api.IProjectSemanticDataDeletionService;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.transaction.event.TransactionalEventListener;
 
+import java.util.Objects;
```
sbegaudeau: import order issue in the configuration of your IDE?

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/listeners/LibraryCleaner.java
sbegaudeau: Technically, it's a bit more subtle, that foreign key was a bit different since it was:

```
<addForeignKeyConstraint
    baseTableName="library"
    baseColumnNames="semantic_data_id"
    constraintName="fk_library_semantic_data_id"
    referencedTableName="semantic_data"
    referencedColumnNames="id" />
```

Contrary to other foreign keys, it didn't have the `onDelete="CASCADE"` option so it would not have deleted the library if one executed something like `DELETE FROM semantic_data;`, instead it would have prevented the deletion entirely. People would have to run `DELETE FROM library;` first.

Using foreign keys between bounded contexts has always been a shortcut that we would have to properly remove one day and today is that day. Considering bounded contexts as independent micro-services that could in theory be executed we a dedicated database, we can not:

- Create a foreign key between library and semantic data
- Prevent anyone from deleting the semantic data underneath a library

I encouraged taking this shortcut for those two bounded contexts knowing full well that we would remove it one day. Even before, anyone with access to the database could just execute a SQL query to remove a library even if we did not supported the feature.

Yet it brings one question, removing the semantic data of a library can be done technically but it's not a state that is reachable in our code, so should we prepare some other pieces of code (`LibraryCleaner`, `ILibraryDeletionService`, `LibraryDeletedEvent`, `Library#dispose`, etc...) to handle this state that should never be reachable.

I was torn on this issue but I knew it would not involve a lot of code so I've asked Naila to just write this along with a test to **see how it feels like**. She just did what I asked for 😄 

If we don't have this code, what would happen is if someone wrote some code to reach this unreachable state in the application, they would have to perform the cleanup themselves. This is not complicated to do, that's very manageable. On the other hand, if we did have this code, we would be a tiny bit more secured in case someone try to do something fishy in Sirius Web by deleting semantic data that should not and cannot be deleted by the regular flow of the application.

I'm really not a fan of dead code and this one should not be reachable at all but one thing that I find **VERY INTERESTING** about you commenting on this pull request is that the simple existence of this code made you believe that we could be changing our mind on the deletion of libraries. Which is why I'll remove this part from the PR.

I've not executed it yet I'm sure it works nicely but because it's just a bunch of unreachable code that can make people think that we are doing things we really don't want to do, it should not be there.

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/library/events/LibraryDeletedEvent.java
sbegaudeau: We still don't want to support the deletion of libraries, it's just that if we had properly respected the boundaries of the various bounded contexts, we could never have prevented someone from deleting the semantic data in a library. Technically, someone could have done this and it would have worked:

```
DELETE FROM library WHERE id = '...';
```

This is exactly the same thing but in code, I asked Naila to remove the foreign key that should never have been added (just like tons of other foreign keys)

packages/sirius-web/backend/sirius-web-domain/src/main/java/org/eclipse/sirius/web/domain/boundedcontexts/representationdata/services/RepresentationMetadataDeletionService.java
```diff
@@ -57,4 +59,20 @@ public IResult<Void> delete(ICause cause, UUID representationMetadataId) {
         }
         return result;
     }
+
+    @Override
+    public IResult<Void> deleteRepresentationMetadata(ICause cause, UUID semanticDataId) {
+        IResult<Void> result = new Failure<>(this.messageService.notFound());
+
+        List<RepresentationMetadata> representationMetadataList = this.representationMetadataRepository.findAllRepresentationMetadataBySemanticDataId(semanticDataId);
+        if (!representationMetadataList.isEmpty()) {
```
sbegaudeau: I don't think that removing nothing should be a failure, it's just a no-op I consider that as a success

packages/sirius-web/backend/sirius-web-infrastructure/src/main/resources/db/changelog/2026.1/01-drop-fks-between-bounded-contexts.xml
```diff
@@ -16,13 +16,22 @@
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.9.xsd">
 
-    <changeSet id="01-drop-fks-project-semantic-data-add-unique-constraints" author="ntinsalhi">
+    <changeSet id="01-drop-fks-between-bounded-contexts" author="ntinsalhi">
```
sbegaudeau: Nope, once a changeset is merged it CANNOT BE CHANGED ever after, it has been deployed "in production", it's identity and checksum is stored in some database in production, it's done. You need a new one.

packages/sirius-web/backend/sirius-web-infrastructure/src/main/resources/db/changelog/2026.1/2026.1.0.xml
```diff
@@ -15,5 +15,5 @@
         xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.9.xsd">
-    <include file="db/changelog/2026.1/01-drop-fks-project-semantic-data.xml" />
+    <include file="db/changelog/2026.1/01-drop-fks-between-bounded-contexts.xml" />
```
sbegaudeau: a new changeset is compulsory

packages/sirius-web/backend/sirius-web-tests-data/src/main/resources/sirius-web-scripts/cleanup.sql
```diff
@@ -9,3 +9,4 @@ DELETE FROM representation_metadata;
 DELETE FROM image;
 DELETE FROM nature;
 DELETE FROM project;
+DELETE FROM project_image;
```
sbegaudeau, line 11: This should be before `DELETE FROM project;` I think

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/libraries/LibraryControllerIntegrationTests.java
```diff
@@ -449,6 +461,32 @@ public void givenProjectWithDependenciesWhenDependencyIsUpdatedThenEditingContex
 
     }
 
+    @Test
+    @GivenSiriusWebServer
+    @DisplayName("Given an existing library linked to semantic data, when a SemanticDataDeletedEvent is published, then the library is deleted")
+    void givenExistingLibraryLinkedToSemanticDataWhenSemanticDataDeletedEventIsPublishedThenLibraryIsDeleted() {
```
sbegaudeau: - ALWAYS use an explicit visibility, here `public` would be nice
- I'll remove this test since it will not be relevant anymore

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/representations/RepresentationMetadataIntegrationTests.java
```diff
@@ -0,0 +1,92 @@
...
+        this.domainEventCollector.clear();
+    }
+
+    @Test
+    @Sql(scripts = {"/sirius-web-scripts/representations-order.sql"},
+            executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD, config = @SqlConfig(transactionMode = SqlConfig.TransactionMode.ISOLATED))
+    @Sql(scripts = {"/sirius-web-scripts/cleanup.sql"},
+            executionPhase = Sql.ExecutionPhase.AFTER_TEST_METHOD, config = @SqlConfig(transactionMode = SqlConfig.TransactionMode.ISOLATED))
+    @DisplayName("Given representations linked to semantic data, when a SemanticDataDeletedEvent is published, then the representation metadata is deleted")
+    public void givenRepresentationsLinkedToSemanticDataWhenASemanticDataDeletedEventIsPublishedThenTheRepresentationMetadataIsDeleted() {
```
sbegaudeau, line 68: Why not use `@GivenSiriusWebServer`? It would be more representative of a real test

sbegaudeau (changes requested):
I'll update the PR and ensure that it gets merged

sbegaudeau (approved):

sbegaudeau:
ARRRGH now it's 2026 and we need to update the copyright of everything we touch -.-

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/listeners/LibraryCleaner.java
gdaniel: > Hummmmmm? What are you talking about?

My bad, I got stuff mixed up, you are right, the foreign key obviously works the way you describe. And there is only one semantic data per library, I wrongly thought we did the same for libraries and projects, I should have double checked.

There is still something I don't understand though: 
- When we delete a project, we fire a `ProjectDeletedEvent` which is used to clean up the associated semantic data (which is fine)
- When we delete the semantic data of a library, we delete the library (which is also fine)

But to be fully consistent, shouldn't we also delete the semantic data of a library when the library is deleted ? 
And shouldn't we delete a project if all its semantic data has been deleted ?

What bugs me here is that we expect stuff to be deleted in a given order, which seems to be based only on the constraints we had regarding the database (using foreign keys to ensure consistency). In the previous version of this PR this lead to a `LibraryDeletionService` that deletes the semantic data of a library instead of the library itself, which feels odd to me.

Anyways, this is probably not the place to discuss it, and it is maybe planned in a future PR, but something was hitching me with the way we implement consistency at the application level.

sbegaudeau: > In the previous version of this PR this lead to a LibraryDeletionService that deletes the semantic data of a library instead of the library itself, which feels odd to me.

The `LibraryDeletionService` was deleting a library after the deletion of its semantic data by who knows who. It was just a theory since that code path should not be reachable.

> Anyways, this is probably not the place to discuss it, and it is maybe planned in a future PR, but something was hitching me with the way we implement consistency at the application level.

I don't mind discussing it here. Technically in a regular database, we have foreign keys which indicate a constraint and a direction for this constraint. For example, let's consider something like this (which is not exactly what we had but it's not an issue for this discussion): 

```
<addForeignKeyConstraint
    baseTableName="library"
    baseColumnNames="semantic_data_id"
    constraintName="fk_library_semantic_data_id"
    referencedTableName="semantic_data"
    referencedColumnNames="id"
    onDelete="CASCADE" />
```

Such a foreign key both indicate that there is a link between `Library#semanticDataId` and `SemanticData#id` and that the link is from `Library` to `SemanticData`. A deletion can only work by deleting the `Library` and after that its `SemanticData` which would just be dangling otherwise. For projects, the relation is a bit different given the existence of `ProjectSemanticData`. The foreign key is in the other direction, from `ProjectSemanticData` to `Project` and to `SemanticData`.

We could represent this coupling using various approaches and thus direction, but we need to pick one at the end of the day. In order to remove those couple problematic foreign keys (we will not remove ALL foreign keys), we need to add the equivalent code in the application layer. We still need to pick a side for this relationship but now, the database does not give us any constraint preventing us from (in theory, because we don't want it) deleting a `Library` and after that its `SemanticData` or deleting the `SemanticData` and then removing the dangling `Library`.

In practice, there is an API to create and delete `SemanticData`, it's there, we are using it, it won't go anywhere. So nothing prevents a downstream application from asking for the deletion of the `SemanticData` of a `Library`. It would not make sense given our expectation but the code path exists for that today. Which is why we tried to see if adding something to react to a `SemanticDataDeletedEvent` in the case of library made sense... because half of the path is already there. One could say that we could add something like this too:

```
@Service
public class LibrarySemanticDataCleaner {
  @TransactionalEventLister
  public void onLibraryDeletedEvent(LibraryDeletedEvent event) {
    // Clean up the semantic data of a library after the deletion of the library
  }
}
```

I would argue against such an addition because this would explicitly support the deletion of a library and cleaning up after the use of this feature. Something like this, which is what we consider in this PR, is a tiny bit more different in my mind:

```
@Service
public class LibraryCleaner {
  @TransactionalEventLister
  public void onSemanticDataDeletedEvent(SemanticDataDeletedEvent event) {
    // Clean up the library after the deletion of its semantic data
  }
}
```

Here, we would be cleaning up a library after the use of a perfectly valid feature, the deletion of some semantic data, which already exist and is actively used. I would 100% agree that the difference between the two approaches is very subtle but I think it's meaningful. There are some workflow in the application that we want to see supported and not others.

We could also add a listener to delete a project if all its `SemanticData` have been deleted but I don't think that it's a code path that we would want to support. I believe that we would instead prevent anyone from deleting the last `SemanticData` of a project and thus force the end user to go to the path of deleting the project. I don't see us adding something like this in a future PR.

So to conclude, the direction of those flows matter and I don't think we want to support all possible directions for all our relations but we also need to think carefully about those chains of events and their consequences. For 2026, I consider that it will be critical to have a map of those different flows because they are still quite a few of them in Sirius Web right now but if you start considering extensions that exist in the wild, it's starting to become quite a complex map to keep in mind. Every single `@TransactionalEventLister` in Sirius Web has (or should have) a meaningful reason to exist the way it is.

Right now, it's manageable (at least in my own mind) but I'm also 100% sure that some downstream projects would find it quite difficult to understand what kind of event they should listen to in order to synchronize their workflows with ours. I know some people had issues with that and they were experts in Sirius Web related stuff.

gdaniel: > The LibraryDeletionService was deleting a library after the deletion of its semantic data by who knows who. It was just a theory since that code path should not be reachable.

I can't retrieve this version of the code, but to me this was the role of `LibraryCleaner`, not `LibraryDeletionService`. Anyways, I don't think the question is on what was in these services anyways, since you removed them.

To me there is no conceptual difference between a project containing semantic data and a library containing semantic data. When we implement them as a database schema, I can understand we have to deal with technical constraints , and we have to make some choices.

Now, we want to remove these database constraints, and implement a similar behavior at the application level. Since we aren't constrained by the database anymore, we can re-think how we want to implement these constraints, and make them more aligned with the conceptual schema.

In practice, we could have something like this:
- If `A` contains `B`, and if there is a service to delete `A`, then there is a listener that deletes `B` when `A` is deleted
- If `A` contains `B`, and if there is a service to delete `B`, then there is either 
  * A listener to delete `A` (if `A` cannot exist without `B`, e.g. a library cannot exist without its semantic data)
  * Something to prevent the deletion of B (this is the example you gave for the deletion of the last semantic data of a project)
  * Nothing (if `A` can exist without `B`, e.g. a project can exist without project images)

My point is, if we want to support the deletion of a library (and I know we don't, but let's assume we do), we should provide a service for it, which does exactly that: delete the library, and send an event indicating the library has been deleted (which will be used to delete its semantic data). The fact that there is another path that also leads to the deletion of a library shouldn't imply this path is the default path.

In fact, this other path could be treated the same way you'd handle the deletion of the last semantic data for a project, and forbidden. This would force the user to go through the path of deleting the library (or, since we don't want to support such deletion, it would prevent it from happening and ensure libraries always keep their semantic data, since neither should be deletable).

sbegaudeau: > I can't retrieve this version of the code, but to me this was the role of LibraryCleaner, not LibraryDeletionService. Anyways, I don't think the question is on what was in these services anyways, since you removed them.

`LibraryCleaner` was the application service and its only job was to call `LibraryDeletionService`.

> To me there is no conceptual difference between a project containing semantic data and a library containing semantic data

Technically, there is one, the 1..n versus 1..1 link with the SemanticData.

> When we implement them as a database schema, I can understand we have to deal with technical constraints , and we have to make some choices. Now, we want to remove these database constraints, and implement a similar behavior at the application level. Since we aren't constrained by the database anymore, we can re-think how we want to implement these constraints, and make them more aligned with the conceptual schema.

I'm not sure that the database was creating any issue in the way our conceptual schema was implemented in the database. I was quite happy with our foreign keys and various constraints. It's just that it does not match what we want in terms of domain driven design approach. We are preventing our own code to know about the consequences of its own actions to maximize the extensibility of the platform.

> If `A` contains `B`, and if there is a service to delete `A`, then there is a listener that deletes `B` when `A` is deleted

I agree but I would rephrase it a bit as "in order to delete `A` and `B`, we will expose a service to delete `A` and it will fire an event that will be listened by another service which will delete `B`". This flow, that we are trying to do every time leads to the creation of a service to delete `B` which gives the ability to someone to just start deleting `B` without trying to delete `A` first.

> If `A` contains `B`, and if there is a service to delete `B`, then there is either, a listener to delete `A` (if `A` cannot exist without `B`, e.g. a library cannot exist without its semantic data)

This was what we tried to test in this PR to see how it would behave, this was the first time we tried it.

> Something to prevent the deletion of `B` (this is the example you gave for the deletion of the last semantic data of a project)

I'm not entirely sure that it can be done in all cases. I don't think that we can make the hypothesis that we can prevent such a deletion. In some cases, `B` is completely independent from `A` and can thus be manipulated independently but no-one should try to do something like deleting `B` directly like that. If someone tries to do this, they would have to deal with the consequences. Nothing can prevent someone from deleting the semantic data of a library.

> Nothing (if A can exist without B, e.g. a project can exist without project images)

I would disagree here, in a distributed architecture "shit happens" and we need to be able to live with an inconsistent state. After thinking about it, if `A` in a micro-service references `B` in another micro-service and `B` no longer exists because "who knows why", then `A` has to be able to live with it. Since our approach requires us to deal with eventual consistency, we need to be able to deal with inconsistencies.

Let's consider the creation for example of a project and its semantic data. There is a time during which the project exists but not its semantic data. It's an inconsistent state with regards to our business concerns but we deal with it anyway and in a micro-service architecture, we would have to deal with even more of those state since the request to create the semantic data of a newly created project may be in transit to the semantic data micro-service.

So if we go back to the basics, the unit of a consistent state is the bounded context. It is the unit of a transaction, you can't expect the state to be perfectly synchronized and consistent outside of the bounded context. If you entity cannot be used with another entity in another bounded context, you have to rethink your bounded contexts to merge them or split them in another way or you just have to provide what's necessary as a fallback.

Honestly if someone use a raw SQL query to delete the semantic data of a library what will happen? Those semantic data will not be found anymore when we collect recursive dependencies and we will have a couple of broken references in the model. I'm not even sure that the application would even suffer any crash. Trying to display the library using `/libraries/:namespace/:name/:version` may result in an empty workbench and that may be it. I think this would be pretty normal.

> My point is, if we want to support the deletion of a library (and I know we don't, but let's assume we do), we should provide a service for it, which does exactly that: delete the library, and send an event indicating the library has been deleted (which will be used to delete its semantic data). The fact that there is another path that also leads to the deletion of a library shouldn't imply this path is the default path.

Yes it should not be the default path, it's wasn't our goal to treat it as the default path but I was wondering if we should try to force some additional consistency. I don't think so now that we have tested it :smile:

> In fact, this other path could be treated the same way you'd handle the deletion of the last semantic data for a project, and forbidden. This would force the user to go through the path of deleting the library (or, since we don't want to support such deletion, it would prevent it from happening and ensure libraries always keep their semantic data, since neither should be deletable).

While `ProjectSemanticData` can have a rule in their application services to check that there is at least one `ProjectSemanticData` (and no duplicate for example too), given that there isn't anything specific to `SemanticData`, we will always have to live with the fact that someone can do something fishy and live with the ability to handle, and maybe recover from, inconsistent states.

Of course, this entire discussion is HIGHLY theoretical given that it would imply that someone would extend Sirius Web with a piece of code to create "forbidden" workflows (i.e. deleting the semantic data directly) without handling the consequences of such actions 😛

sbegaudeau: > I agree but I would rephrase it a bit as "in order to delete A and B, we will expose a service to delete A and it will fire an event that will be listened by another service which will delete B". This flow, that we are trying to do every time, leads to the creation of a service to delete B which gives the ability to someone to just start deleting B without trying to delete A first.

I could refine this by adding that technically, we will create an application service to delete `A` (along with a domain service to delete `A`) and only a domain service to delete `B`. As such, we are explicitly not providing support for the direct deletion. This is what makes one path clear.

Issue #5810, opened 2025-11-17
Migration participants are not applied on representations during project upload


frouene:
During upload, diagrams are created via `RepresentationPersistenceService#save`. Since the representation does not yet exist, its `RepresentationMigrationData` is initialized with the latest available `migrationParticipant`. As a result, no migrations are applied to the uploaded diagrams.
This causes an issue: if new `migrationParticipants` were added between the export and the upload, those migrations are not executed on these diagrams.

PR #6021, opened 2026-01-07
[doc] Add ADR on representation migration tracking

theogiraudet:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/5810

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/adrs/221_add_migration_version_to_exported_representation.adoc
+++ b/doc/adrs/221_add_migration_version_to_exported_representation.adoc
@@ -1,0 +1,53 @@
+= ADR-221 - Add migration version to exported representation
+
+== Context
+
+When importing a representation, we assign it the version of the newest migration participant for the current migration.
+Therefore, we cannot migrate a representation when new participants have been added to Sirius Web after exporting it.
+
+== Decision
+
+=== Backend
+
+We must keep track of the latest migration version applied on the representation.
+Such a metadata can be directly stored inside the exported `manifest.json`:
+
+[source,json]
+----
+{
+  "representations": {
+    "<representationId>": {
+      "lastMigrationPerformed": "<migrationName>",
+      "migrationVersion": "<version>"
+    }
+  }
+}
+----
+
+=== Import
+
+The migration data will be obtained by `RepresentationProjectContentImportParticipant#getRepresentationManifest` in `RepresentationProjectContentImportParticipant#createRepresentations`.
+If the data does not exist, `0` and `none` will be used as default value to apply all migration participants over the representation.
+`IRepresentationImporterUpdateService#update` will be updated to take the migration data as parameters.
+
+As all representations will now have migration data `IRepresentationContentUpdateService#updateContentByRepresentationId` will be removed and `IRepresentationContentUpdateService#updateContentByRepresentationIdWithMigrationData` will be renamed to `update`.
+
+=== Export
+
+`ProjectRepresentationDataExportParticipant#exportData` will be updated to put the migration data into the representation manifest.
+
+== Consequences
+
+- Representation metadata in the exported JSON manifest now have two new keys `migrationVersion` and `lastMigrationPerformed`.
+We will keep support for JSON without such fields to avoid breaking changes at this level.
+- `IRepresentationImporterUpdateService#update` has two new parameters: `migrationVersion` and `lastMigrationPerformed`.
+- `IRepresentationContentUpdateService#updateContentByRepresentationId` is removed.
+- `IRepresentationContentUpdateService#updateContentByRepresentationIdWithMigrationData` is renamed to `update`.
+
+== Cutting backs
+
+/
+
+== Status
+
+Proposed.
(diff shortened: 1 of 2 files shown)


doc/adrs/221_add_migration_version_to_exported_representation.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+When importing a representation, we assign it for current migration the version of the newest migration participant.
+Therefore, we cannot migrate a representation when new participants have been added to Sirius Web after exporting it.
+
+== Decision
+
+=== Backend
+
+We must add migration metadata to the exported representation, just like it is done for ecore-based documents.
+This metadata will take the form of a new key placed at the root of the JSON document:
```
sbegaudeau: I'm not sure that it's the easiest way to do that. It may be more complicated than adding it somewhere else in the archive produced like in the manifest somewhere. You would have to transform the JSON exported to add it and transform the JSON imported to remove it.

doc/adrs/221_add_migration_version_to_exported_representation.adoc
```diff
@@ -0,0 +1,66 @@
...
+{
+  "migration": {
+    "lastMigrationPerformed": "...",
+    "migrationVersion": "..."
+  }
+}
+----
+
+They will be added to the JSON by the `ProjectRepresentationDataExportParticipant#exportData` method,
+which currently serialize the representation.
```
sbegaudeau: Always use one sentence per line, don't create a new line mid-sentence.

doc/adrs/221_add_migration_version_to_exported_representation.adoc
```diff
@@ -0,0 +1,66 @@
...
+The migration metadata will be obtained using `IRepresentationContentMigrationService#getLatestMigrationData`,
+which returns a MigrationData of the newest migration participant for the given representation kind.
+This method will be added to the migration service interface to avoid coupling with the persistence layer.
```
sbegaudeau: Why would the last migration participant be applied on the representation being exported? That's not how Sirius Web works. Just ask the persistence layer for this specific piece of information. We must already be loading the `RepresentationContent` from the database, we are thus already loading this information from the database.

doc/adrs/221_add_migration_version_to_exported_representation.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+=== Import
+
+`RepresentationProjectContentImportParticipant#getRepresentationImportData` will extract migration data from the JSON and place it into `RepresentationImportData`.
+Those data are nullable, enabling import of projects created with Sirius Web versions older than the one implementing the migration for representation.
```
sbegaudeau: I don't believe that any of those field is really nullable, they are not DTO and we may not have added any NonNullable check or metadata but don't make this hypothesis. We will have to provide some value for previously exported projects.

doc/adrs/221_add_migration_version_to_exported_representation.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+In `DiagramImporterUpdateService:171`, `this.representationContentUpdateService.updateContentByRepresentationIdWithMigrationData` will be used when the migration data are present in the `RepresentationImportData` given in parameters.
```
sbegaudeau: - Don't use that level of details in the ADR, as an example, this line does not match anything relevant for your ADR because the service has already been changed.
- The method `updateContentByRepresentationId` should probably instead be removed and `updateContentByRepresentationIdWithMigrationData` be renamed to `update`

doc/adrs/221_add_migration_version_to_exported_representation.adoc
```diff
@@ -0,0 +1,66 @@
...
+When migration data are null, three scenarios are possible:
+
+==== Option 1: No Version Tracking (Current Behavior)
+Call `updateContentByRepresentationId()` without migration data.
+- Behavior depends on how `RepresentationContentMigrationService` handles NULL
```
sbegaudeau: Why `NULL`? I think that the column is not nullable, what is the current value in the database when a project is uploaded? I think we could go with a full migration if the default value is valid.

sbegaudeau (changes requested):

doc/adrs/221_add_migration_version_to_exported_representation.adoc
sbegaudeau: The current value seems to be `none` so there are not three use cases `null` vs some hardcoded value vs `none`. We should just make sure that `none` triggers everything.

sbegaudeau (approved):

PR #6026, opened 2026-01-08
[5974] Tweak the node actions UI to reduce the risk of invoking them by mistake

### LINKED ISSUE #5974: Reduce the risk of misclick on node actions
When a node provides quick actions, for example "Manage visibility", the corresponding icon appears right to the node's label as soon as the user hovers on the node.
Because the icon was invisible right before and is relatively large, it makes it easy when trying to select a node to trigger the action by mistake.

We will:

1. reduce (a little) the size of the icon
2. move it further into the corner
3. add a small delay before it appears

For example on SysON, now (left) and what we want (right):

Image

pcdavid:
This also reduces the number of requests sent to the backend when the user moves the cusrors on a diagram, as we now only fetch the actions after a node has been hovered for 0.5s instead of immediately.

Before:

https://github.com/user-attachments/assets/455275a4-e9af-4693-b504-7088a3a38a68

After:

https://github.com/user-attachments/assets/51eddc90-3b1a-476a-8f8f-ffa265c30586

Bug: https://github.com/eclipse-sirius/sirius-web/issues/5974

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/Action.tsx
+++ b/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/Action.tsx
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2025 Obeo.
+ * Copyright (c) 2025, 2026 Obeo.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
@@ -20,11 +20,14 @@ import { useInvokeAction } from './useInvokeAction';
 
 const useStyles = makeStyles()((theme) => ({
   actionIcon: {
+    alignSelf: 'start',
     '&:hover': {
       backgroundColor: theme.palette.action.selected,
     },
   },
 }));
+
+export const ACTION_ICON_SIZE = 10;
 
 export const Action = ({ action, diagramElementId }: ActionProps) => {
   const { classes } = useStyles();
@@ -45,7 +48,13 @@ export const Action = ({ action, diagramElementId }: ActionProps) => {
   } else {
     return (
       <IconButton className={classes.actionIcon} size="small" onClick={() => invokeAction(action)}>
-        <IconOverlay iconURLs={action.iconURLs} title={action.tooltip} alt={action.tooltip} />
+        <IconOverlay
+          iconURLs={action.iconURLs}
+          customIconWidth={ACTION_ICON_SIZE}
+          customIconHeight={ACTION_ICON_SIZE}
+          title={action.tooltip}
+          alt={action.tooltip}
+        />
       </IconButton>
     );
   }
--- a/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/ActionsContainer.tsx
+++ b/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/ActionsContainer.tsx
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2025 Obeo.
+ * Copyright (c) 2025, 2026 Obeo.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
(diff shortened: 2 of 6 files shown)


packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/useActions.tsx
```diff
@@ -49,26 +49,30 @@ const isDiagramDescription = (
...
+  const [fetchActions, { data, error }] = useLazyQuery<GQLGetActionsData, GQLGetActionsVariables>(getActionsQuery);
 
   useEffect(() => {
     if (error) {
-      addErrorMessage('An unexpected error has occurred while retrieving actions, please refresh the page');
+      addErrorMessage(error.message);
     }
   }, [error]);
 
+  const getActions = (diagramElementId: string) => {
```
AxelRICHARD: Shouldn't it be named `loadActions` or something like that instead of `getActions`? Because this method does not return the actions.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/ActionsContainer.tsx
AxelRICHARD, line 64: If I'm not mistaken, calling `actions` before calling `getActions` will return an empty list, right?
If so, could you write, somewhere that we need to call `getActions` before calling actions` please?

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/Action.tsx
```diff
@@ -20,12 +20,15 @@ import { useInvokeAction } from './useInvokeAction';
 
 const useStyles = makeStyles()((theme) => ({
   actionIcon: {
+    alignSelf: 'start',
```
AxelRICHARD, line 23: The mechanism has been designed to handle one or several actions.
With this change, how are several actions are displayed?
Could you test that please?

pcdavid: I've tested on a Papaya class diagram, it works as expected with multiple actions:

image

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/ActionsContainer.tsx
sbegaudeau: I'm not in favor of this, we have countless of hooks with this pattern:

```
const { value, doSometing} = useStuff();
```

Everyone should know that the `value` is updated by calling `doSometing`. I don't want it to be a requirement, that's the basic behavior of several hooks in React.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/useActions.tsx
sbegaudeau: This pattern is visible in countless hooks in our codebase. But here, the change could be even smaller, we could simply ask for the property `skip` like this: 

```
export const useActions = (diagramElementId: string, skip: boolean): UseActionsValue => {
```

And use it below. There's no need to make this a lazy query. It would just be:

```
const { actions } = useActions(diagramElementId, shouldDisplay);
```

It wouldn't need one more rendering tick from the useEffect in `ActionsContainer`

pcdavid: Done.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/ActionsContainer.tsx
pcdavid: Done. I've actually also changed the semantics a little, so that `actions` is `null` (and nothing is displayed) until we've actually loaded the available actions from the backend (in which case it will be non-null, but might be empty).

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/useActions.tsx
pcdavid: Actually, not done. I had not seen @sbegaudeau's remark. I'll rework the PR.

pcdavid: Done.

AxelRICHARD (approved):
All remarks have been taken into account, thank you.

PR #6078, opened 2026-01-19
[doc] Add ADR on layout on diagram creation

frouene:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -8,7 +8,7 @@
 
 === Architectural decision records
 
-
+- [ADR-222] Add auto layout until manual change on diagram
 
 === Deprecation warning
 
--- a/doc/adrs/222_add_auto_layout_until_manual_change.adoc
+++ b/doc/adrs/222_add_auto_layout_until_manual_change.adoc
@@ -1,0 +1,37 @@
+= ADR-222 - Add auto layout until manual change on diagram
+
+== Context
+
+Some diagrams need to be laid out using an external layout algorithm like ELK.
+It is currently possible to either trigger this layout manually (via a button in the UI) or configure the diagram description to apply it automatically.
+In the latter case, the user cannot manually change the position of a node, as the algorithm automatically runs after every action.
+A third scenario proposed in this ADR would be to trigger this layout until any user wants to manually customize the layout.
+
+== Decision
+
+=== View DSL
+
+Transform the boolean `autoLayout` feature into a `layoutOption` enum as follows:
+
+[source,ecore]
+----
+<eClassifiers xsi:type="ecore:EEnum" name="DiagramLayoutOption">
+<eLiterals name="None"/>
+<eLiterals name="Auto" value="1"/>
+<eLiterals name="AutoUntilManual" value="2"/>
+</eClassifiers>
+----
+
+=== Backend
+
+Add a migration participant to migrate existing diagram descriptions.
+
+=== Frontend
+
+Add a state in the diagram to know if auto-layout is activated or not.
+If a user triggers one action that creates some manual layout (like moving a node), then this state is deactivated.
+Add a button in the DiagramPanel to activate or deactivate this auto-layout.
+
+== Status
+
+Proposed.


doc/adrs/XXX_add_layout_on_diagram_creation.adoc
```diff
@@ -0,0 +1,37 @@
+= ADR-XXX - Add Layout on Diagram Creation
+
+== Context
+
+Some diagrams need to be laid out using an external layout algorithm like ELK.
+It is currently possible to either trigger this layout manually (via a button in the UI) or configure the diagram description to apply it automatically.
+In the latter case, the user cannot manually change the position of a node, as the algorithm automatically runs after every action.
+A third scenario proposed in this ADR would be to trigger this layout only at creation, allowing the user to manually modify the layout afterward.
```
sbegaudeau: In my mind, a key point of context is missing. Tons of diagrams are using asynchronous node descriptions, auto-layouting things on diagram creation would not do anything for those use cases since the issue would only be visible after that.

doc/adrs/XXX_add_layout_on_diagram_creation.adoc
```diff
@@ -0,0 +1,37 @@
...
+----
+
+=== Backend
+
+Add a migration participant to migrate existing diagram descriptions.
+
+=== Frontend
+
+It is currently not possible to easily determine if a diagram is being opened for the first time.
+However, it is possible to check if the diagram has not yet been modified (in terms of its layout) by verifying that there is no data in the `DiagramLayoutData`.
```
sbegaudeau: How would that distinguish between a diagram with no content and this use case:

- I create a diagram
- I put something inside
- I remove everything
- I use a tool creating some diagram elements

Unless I'm mistaken, here I would receive a diagram without any layout data and I would be far from having just created the diagram.

We shouldn't think about this in terms of the workflow of Sirius Web. Creating a diagram is a 100% backend operation. Everything after that is just subscribing to a diagram. It can be empty or not but I don't think you can distinguish an empty diagram from a diagram that would have been created and never opened.

We could also consider use cases where someone would create a diagram, open it, close it and then re-open it and start playing with it.

sbegaudeau (changes requested):
Given the issues in terms of lifecycle and use cases (unsynchronized diagrams), I'm not even sure that what we want (as a generic platform feature) is really to "add the ability to trigger the layout when a diagram is created".

For example, another approach could be to have the auto-layout state be synchronized with the diagram state and updated. For example, start with a diagram with the auto-layout active and when some operation "disable it" when it switches to a manual layout. It would work with unsynchronized diagram which would stay in auto-layout until some end user would start changing things for example and it would not try to guess what is a newly created diagram.

frouene:
Okay, so we shouldn't talk about “at creation,” because that's not a state that can be captured, but rather “as long as there are no changes,” which can be captured.
Knowing that this responds to the use case.
What do you think to change enum value `LAYOUT_ON_CREATION` by `AUTO_UNTIL_MANUAL` ?

sbegaudeau:
> What do you think to change enum value `LAYOUT_ON_CREATION` by `AUTO_UNTIL_MANUAL` ?

Something along those lines yes. We should also identify what would deactivate the auto-layout. I suspect that the auto layout could just become a diagram state (i.e. something like `Diagram#isAutoLaidOut`) that can be activated or deactivated in the `DiagramPanel` like the helper line. I'm not really sure that it would make sense for one person to have the diagram in auto-layout while not for others.

In this case, one could switch it on or off by just clicking the button. It would be fairly easy to identify if it is on or off become its state would be visible. One could also re-activate it easily.

I also suspect that any manual layout change such as:

- move
- resize
- create / edit bendpoints

Would switch it off automatically. It would thus probably require a dedicated mutation for the button in the `DiagramPanel`.

frouene:
I modified the ADR to no longer refer to creation but to “until a manual change”
And add the notion of state to know if the auto-layout is activated or not

sbegaudeau (approved):

Issue #6086, opened 2026-01-19
Edges "to self" are not created where the user asked


pcdavid:
When creating an edge between two nodes, the actual location of the anchor points are determined by the relative location of the nodes instead of where the user clicked, which makes sense in this case.

When the edge is created from a node to itself, the side chosen "between the node and itself" is always on the left, regardless of where the user chose and what of feedback showed:

![Image](https://github.com/user-attachments/assets/90b51bd0-acdf-4af4-8060-9a8405ccfdb7)

PR #6177, opened 2026-02-11
[6067] Add support for relative position bending points for edges

### LINKED ISSUE #6067: Manually placed bendpoints should move when the source/target of the edge is moved
- Create a container C
- Create two nodes inside A and B
- Create an edge between A and B with specific bendpoints moved manually
- Move C, the bendpoints should be moved too and they are not

### LINKED ISSUE #6085: Moving a node with a manually routed "edge to self" has unexpected behavior
When an edge between two nodes has manually routed segments, moving one of the nodes keeps all but the segment connected to the moving node in place. This is nice.

However in the case of an "edge to self", this has an unexpected behavior. As an end-users, I would expect that the whole edge moves with the node, and it is only connected to it. However, the first and last segments move, and intermediate ones are kept "static" on the diagram.

Image

frouene:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/6067

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


mcharfadi (changes requested):
I have an issue reproducing the case https://github.com/eclipse-sirius/sirius-web/issues/6067
with several child nodes.
Also note the weird lines at the left of the node

https://github.com/user-attachments/assets/99a1a5b0-2942-47fc-9bd5-b0952f26c76a

I'm not sure the issue https://github.com/eclipse-sirius/sirius-web/issues/6086 should be linked, when creating an edge on self, the handles are still placed where I did not click.

https://github.com/user-attachments/assets/1a6c8fbb-2fb7-4898-9529-6c8e3ffe450f

frouene:
Actually, it wasn't the right bug, it was #6085.

mcharfadi (commented):
https://github.com/user-attachments/assets/db273b8f-82fc-4bf8-b51f-38fe6fc4e56b
If I move the node quickly I can change the position of the bending points.
I suspect that this behavior would be easier to reproduce with a diagram with more elements.

packages/diagrams/frontend/sirius-components-diagrams/src/converter/ListNodeConverter.ts
```diff
@@ -118,6 +118,7 @@ const toListNode = (
     minComputedWidth: gqlNodeLayoutData?.minComputedSize?.width ?? null,
     minComputedHeight: gqlNodeLayoutData?.minComputedSize?.height ?? null,
     isLastNodeSelected: false,
+    moving: false,
```
mcharfadi, line 122: What's the use of this `moving `property ?
Is it only used in `useDropNodes `?

How would this work if we're moving a node an we receive another payload from another user ?

I feel like this property should be local only and not change when converting the diagram.

If the current moving property does not reflect the nodes that are actually moving we might want to better update this property when receiving a `NodeChange `from ReactFlow.

frouene: Yes it's only used in useDropNodes, I need this information to know if the node move even it's not an actual move, like for child node.
I used it like the other datas needed for useDropNodes: isDraggedNode, isDropNodeTarge, isDragNodeSource

mcharfadi: But should we not set the value to the previous value instead of false.

What would happen if a node has `moving true` but `another user refresh the diagram`, that would set `moving to false` and create an incorrect state, no ?

We might want to do like the selection and retrieve the current one after converting.

As a side note we should find a naming convention or something to differentiate between `local ` attributes of a node and synchronized ones.

frouene: In this case, we only want capture the move if it's made by the current user, it's the same thing that for the other data like `isDraggedNode`, if a refresh from another user is received during DnD the data is reset to false.

PR #6182, opened 2026-02-12
[doc] Add an ADR to support download/upload of multiple semantic data

gdaniel:
# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [x] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/doc/adrs/222_add_support_for_download_upload_of_projects_with_multiple_semantic_data.adoc
+++ b/doc/adrs/222_add_support_for_download_upload_of_projects_with_multiple_semantic_data.adoc
@@ -1,0 +1,59 @@
+= ADR-222 - Add support for download/upload of projects with multiple semantic data
+
+== Context
+
+Downloading a project produces a zip file that only contains data from one `ProjectSemanticData` (_main_ by default, but it can be set to a specific one with the appropriate parameters).
+
+As a result, some information is lost when downloading a project containing multiple `ProjectSemanticData`.
+This is typically not an issue in Sirius Web, because we don't provide user-level ways to create multiple `ProjectSemanticData` for a given project, but it can happen in downstream applications.
+
+We should ensure all the `ProjectSemanticData` are downloaded for a given project, and that they are all correctly re-created when the zip file is uploaded in Sirius Web.
+
+== Decision
+
+We will change the structure of the produced zip file as follows:
+
+====
+/
+  <project_name>
+
+    <project_semantic_data_name1>
+      documents
+        (list of documents in project semantic data 1)
+      representations
+        (list of representations on objects from project semantic data 1)
+      manifest.json (manifest of project semantic data 1, i.e. metamodels, dependencies, representations)
+
+    <project_semantic_data_name2>
+      documents
+        (list of documents in project semantic data 2)
+      representations
+        (list of representations on objects from project semantic data 2)
+      manifest.json
+
+    manifest.json (manifest of the project itself, e.g. to store the nature)
+====
+
+=== Export
+
+We will update `ProjectExportService` to retrieve all the `ProjectSemanticData` associated to the provided project.
+We will update the `IProjectExportParticipant` to make them aware of the `ProjectSemanticData` being processed (so they can use its name to store their data).
+
+=== Import
+
+We will update `SemanticDataCreator` to create all the `SemanticData` of an uploaded project.
+We will update `ProjectSemanticDataCreator` to create all the `ProjectSemanticData` of an uploaded project.
+Since `ProjectSemanticDataCreator` reacts to the creation of `SemanticData`, we need to have a way to indicate the name of the `ProjectSemanticData` to create.
+We will update `IProjectContentImportParticipant` implementations to take into account the new structure of the zip file.
+
+== Consequences
+
+Project duplication will also support the duplication of all the `ProjectSemanticData` of a project, since it relies on the same service as the download/upload feature.
+
+== Cutting backs
+
+
+
+== Status
+
+Proposed.
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -8,6 +8,7 @@
 
 === Architectural decision records
 
+- [ADR-222] Add support for download/upload of projects with multiple semantic data
 
 
 === Deprecation warning


doc/adrs/222_add_support_for_download_upload_of_projects_with_multiple_semantic_data.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+    <project_semantic_data_name2>
+      documents
+        (list of documents in project semantic data 2)
+      representations
+        (list of representations on objects from project semantic data 2)
+      manifest.json
+
+    manifest.json (manifest of the project itself, e.g. to store the nature)
+====
```
gdaniel, line 35: I don't know what's best between 1 manifest containing all the information and 1 manifest / project semantic data folder.

This looks a tiny bit easier to implement given the current code, but it's not a big deal

doc/adrs/222_add_support_for_download_upload_of_projects_with_multiple_semantic_data.adoc
```diff
@@ -0,0 +1,59 @@
...
+=== Export
+
+We will update `ProjectExportService` to retrieve all the `ProjectSemanticData` associated to the provided project.
+We will update the `IProjectExportParticipant` to make them aware of the `ProjectSemanticData` being processed (so they can use its name to store their data).
+
+=== Import
+
+We will update `SemanticDataCreator` to create all the `SemanticData` of an uploaded project.
+We will update `ProjectSemanticDataCreator` to create all the `ProjectSemanticData` of an uploaded project.
+Since `ProjectSemanticDataCreator` reacts to the creation of `SemanticData`, we need to have a way to indicate the name of the `ProjectSemanticData` to create.
```
gdaniel, line 46: I don't know how to do this.
`SemanticDataCreator` can create all the semantic data for a given project, but it needs to pass the name of the `ProjectSemanticData` to `ProjectSemanticDataCreator`.
One way to do it is to create an intermediate event where we put this information, and use it has the cause of the semantic data creation. This event could have as its own cause the `ProjectCreatedEvent` so we don't lose information.

But this solution feels weird: I don't think we do this kind of thing anywhere else in Sirius Web.

doc/adrs/222_add_support_for_download_upload_of_projects_with_multiple_semantic_data.adoc
sbegaudeau: I don't mind but this change needs to stay compatible with previous zip. I would suggest adding a manifest.json at the root under the project name with a field "version": "2.0" to easily detect which strategy to adopt.

Honestly, I wouldn't be surprised if we ended up with a set of services specifically tailored per versions because you will not have the same data structure and thus the same behavior to parse all that. For example, in `ProjectUploadApplicationService#uploadProject`, we could end up with something like this:

```
public IPayload uploadProject(UploadProjectInput input) {
    var version = this.detectVersion(input.file());
    var payload = new ErrorPayload("version not supported...");
    switch (version) {
        case "1.0":
           payload = this.uploadProjectv1(input);
           break;
        case "2.0":
           payload = this.uploadProjectv2(input);
           break;
    }
    return payload;
}
```

I wouldn't be surprised if, as a consequence we ended up with:

- A split of the common behavior between upload and duplicate
- Additional paths for the creation of projects that downstream applications would have to consider

doc/adrs/222_add_support_for_download_upload_of_projects_with_multiple_semantic_data.adoc
```diff
@@ -0,0 +1,59 @@
...
+        (list of representations on objects from project semantic data 2)
+      manifest.json
+
+    manifest.json (manifest of the project itself, e.g. to store the nature)
+====
+
+=== Export
+
+We will update `ProjectExportService` to retrieve all the `ProjectSemanticData` associated to the provided project.
+We will update the `IProjectExportParticipant` to make them aware of the `ProjectSemanticData` being processed (so they can use its name to store their data).
```
sbegaudeau, line 40: Can we? While supporting both variations of the zip files? I suspect that what you are describing would look like this `Service -> Service (updated)`, I'm not sure that it would be easier than:

```
                Service v1            Service v2               
                   ││                    ││                    
                   ││                    ││                    
                   ││                    ││                    
                   ││                    ││                    
                   ││                    ││                    
                   ││                    ││                    
 CommonServiceA◄───┘└──►CommonServiceB◄──┘└──►CommonServiceC   
```

where we would try to create a v2 version of the top level service (probably the entire content of `ProjectUploadApplicationService#uploadProject`) and identify common subservices that should be shared. I am not even sure that the key data structure used for the import (and project duplication) `ProjectZipContent` can even be used between such a v1 and v2.

This would introduce a split between the project import and the project duplication which would have to exist anyway for this feature to exist. We could copy all editing context in the project duplication, the split would still exist with the import of v1 zip files.

Compared to most pieces of code that we can just change, the support for the v1 zip files needs to continue to exists in parallel of the v2, and v3 one day. And we need to be able to improve the support of v1 zip files and fix potential bugs in their support.

The management of those various versions of the code, with potentially their dedicated data structure and events and cause and their impact on the rest of the code and downstream applications is the KEY problem to address here.

doc/adrs/222_add_support_for_download_upload_of_projects_with_multiple_semantic_data.adoc
sbegaudeau: I'm not sure that it's the goal of the `SemanticDataCreator`

doc/adrs/222_add_support_for_download_upload_of_projects_with_multiple_semantic_data.adoc
```diff
@@ -0,0 +1,59 @@
...
+=== Import
+
+We will update `SemanticDataCreator` to create all the `SemanticData` of an uploaded project.
+We will update `ProjectSemanticDataCreator` to create all the `ProjectSemanticData` of an uploaded project.
+Since `ProjectSemanticDataCreator` reacts to the creation of `SemanticData`, we need to have a way to indicate the name of the `ProjectSemanticData` to create.
+We will update `IProjectContentImportParticipant` implementations to take into account the new structure of the zip file.
+
+== Consequences
+
+Project duplication will also support the duplication of all the `ProjectSemanticData` of a project, since it relies on the same service as the download/upload feature.
```
sbegaudeau, line 51: I am not convinced that:

- everyone want this all the time (but I'm not strongly opposed to this, it's not a big deal we could make it configurable later)
- adding this constraint on top of everything that needs to be done first is the fastest way to proceed

I would rather break as little as possible of the existing code (extract in reusable services the minimum required for the new services). We have countless of existing zip in the wild and keeping the ability to import them without issues is absolutely critical. There shouldn't be any doubt regarding our ability to import v1 zip and update the code supporting v1 zip to fix issues. And it should not impact the support for v2 zip. But copying everything may also create potential issues in terms of the number of caused and possible events fired.

sbegaudeau (changes requested):
I think first, we will need a map of the event chains involved in this to understand the consequences of this topic.

gdaniel:
I see your point on having to support v1 and v2, I'll update the ADR to take this into account.
Looking at the event map, here is the current chain of events with some relevant classes for a project upload:

Raw chain of events:
```
UploadProjectInput 
  -> InitializeProjectInput 
    -> ProjectCreatedEvent 
      -> SemanticDataCreatedEvent 
        -> ProjectSemanticDataCreatedEvent
```

Relevant classes:

- `SemanticDataCreator` reacts to the `ProjectCreatedEvent` to create the semantic data of the project. At the moment this listener creates a single semantic data.
- `ProjectSemanticDataCreator` reacts to the `SemanticDataCreatedEvent` to produce the project semantic data. Since there is only one semantic data produced in the process, there is only one project semantic data created.
- `ProjectContentImporter` reacts to the `ProjectSemanticDataCreatedEvent` to initialize the content of the semantic data, and to `SemanticDataUpdatedEvent` to create the representations associated to the semantic data

There are other listeners that react to `SemanticDataCreatedEvent`, but I don't think they are related to this feature, I put them below in case I am missing something:
- StandardLibraryInitializer
- ProjectLibrariesImporter
- ProjectSemanticDataCreator
- StudioPublicationListener
- StudioLibraryRepresentationPublicationListener
- IndexLifecycleManager
- ForkedStudioSemanticDataCreationService

In a similar way, `ProjectSemanticDataInitializer` reacts to `ProjectSemanticDataCreatedEvent`, but it not related to this feature since it is used to initialize projects from templates.

Looking at the change I put in the ADR, my point was to find a way to create multiple semantic data from a single initial `UploadProjectInput`.
The solution I proposed was to update the `SemanticDataCreator` with something like this:

```
void onProjectCreatedEvent(ProjectCreatedEvent event) {
  if (event.causedBy() instanceof InitializeProjectInput input) {
    for (String name : this.findProjectSemanticDataNames(input.getProjectContent()) {
      this.semanticDataCreationService.create(event, List.of(), List.of(), List.of());
    }
  } else {
    // Regular project creation, create just one semantic data
    this.semanticDataCreationService.create(event, List.of(), List.of(), List.of());
  }
}
```

But it doesn't feel right, because:
- This makes `SemanticDataCreator` aware of the name of the `ProjectSemanticData` (not sure if this is an issue)
- `SemanticDataCreator` needs to pass this `name` somehow down the event chain to let `ProjectSemanticDataCreator` create the appropriate `ProjectSemanticData` (with the correct name). This could be done with a custom event (something like `ProjectCreatedEvent -> CustomEventWithTheNameOfTheProjectSemanticData -> SemanticDataCreatedEvent` but we don't do that usually in Sirius Web, and it feels like breaking bounded contexts.

sbegaudeau:
> The solution I proposed was to update the SemanticDataCreator with something like this:

I'm not fan of this since it goes against the responsibilities of `SemanticDataCreator` which is not there to handle project semantic data concerns. This should be the responsibility of the `ProjectSemanticDataCreator`. There's something in this which is indeed a bit odd.

gdaniel:
Changed this PR to a draft since we didn't agree on a solution and this work has been stopped

PR #6188, opened 2026-02-13
[4163] Allow external resource loaders to return a report on upload

### LINKED ISSUE #4163: IUploadDocumentReportProvider do not cover some use cases
In the current implementation (2024.9.0) the _IUploadDocumentReportProvider.createReport_ method is called on the after the reloading the Resource from the database.

But is most cases the error/warning we want to report to the user are found during _IUploadDocumentReportProvider.getResource(InputStream, URI, ResourceSet, boolean)_. I have not found a nice way to add elements in the report problems detected during the phase. I have tries to add some adapters or use _Resource.getErrors()_ but since the Resource is reloaded from the database everything is lost.

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/4163

Fixes #4163

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management
- [x] Have the relevant API breaks been described in the `CHANGELOG.adoc`? (Including changes in the GraphQL API)

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/services/TestUploadDocumentReportProvider.java
gdaniel, line 1: We don't need this class anymore, we have a `DefaultUploadDocumentReportProvider` that can handle `LoadingReport`

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/document/services/UploadDocumentEventHandler.java
```diff
@@ -123,7 +122,7 @@ public void handle(Sinks.One<IPayload> payloadSink, Sinks.Many<ChangeDescription
                     var id = optionalId.get();
                     var name = optionalName.get();
 
-                    String report = this.getReport(newResource);
+                    String report = this.getReport(success.data());
```
gdaniel, line 125: This allows report providers to work with both the final resource and the report if needed.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/document/services/api/IUploadDocumentReportProvider.java
```diff
@@ -12,17 +12,15 @@
...
 package org.eclipse.sirius.web.application.document.services.api;
 
-import org.eclipse.emf.ecore.resource.Resource;
-
 /**
  * Use to provide report when uploading a document into Sirius Web.
  *
  * @author arichard
  */
 public interface IUploadDocumentReportProvider {
```
gdaniel, line 20: To be fair I think this class is a bit overkill: we could produce a report as a list of strings in the `IExternalResourceLoaderService` implementations and simply send it in the payload.

But since this interface was already there I didn't want to add a breaking change like that, especially if some downstream applications use the `resource` to produce a report, which is not exactly the same here than in `IExternalResourceLoaderService` implementations (it is the resource before being sanitized).

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/documents/DocumentControllerIntegrationTests.java
```diff
@@ -176,6 +187,36 @@ public void givenStudioWhenTheUploadOfReadOnlyDomainXMIDocumentIsPerformedThenTh
...
 
+    @Test
+    @GivenSiriusWebServer
+    @DisplayName("Given a studio, when the upload of a new domain document produces a report, then the report is returned")
+    public void givenStudioWhenUploadOfDomainDocumentProducesAReportThenTheReportIsReturned() {
+        String reportMessage = "Test error";
+        LoadingReport testReport = new LoadingReport(List.of(reportMessage));
+        // Stub getResource to check if returning a LoadingReport is correctly propagated to the returned payload.
+        doReturn(new ExternalResourceLoadingResult(Optional.of(new JSONResourceFactory().createResourceFromPath("empty")), testReport))
+                .when(this.jsonExternalResourceLoaderService).getResource(any(), any(), any(), anyBoolean());
```
gdaniel: The existing `IExternalResourceLoaderService` implementations never returns a report with actual data.
I stubbed the method to test it correctly gets propagated instead of implementing an `IExternalResourceLoaderService` for the tests.
Let me know if you prefer a test-level implementation.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/document/services/api/IExternalResourceLoaderService.java
```diff
@@ -13,21 +13,18 @@
...
  *
  * @author arichard
  */
 public interface IExternalResourceLoaderService {
 
     boolean canHandle(InputStream inputStream, URI resourceURI, ResourceSet resourceSet);
 
-    Optional<Resource> getResource(InputStream inputStream, URI resourceURI, ResourceSet resourceSet, boolean applyMigrationParticipants);
-
+    ExternalResourceLoadingResult getResource(InputStream inputStream, URI resourceURI, ResourceSet resourceSet, boolean applyMigrationParticipants);
```
sbegaudeau: I thought this was supposed to return an `Optional<ExternalResourceLoadingResult>` instead of having an `Optional<Resource>` inside (by the way, we pretty much never used optional in fields and data structure).

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/document/services/DocumentSanitizedJsonContentProvider.java
```diff
@@ -76,9 +77,10 @@ public Optional<SanitizedResult> getContent(ResourceSet resourceSet, String name
         Optional<SanitizedResult> optionalResult = Optional.empty();
 
         URI resourceURI = new JSONResourceFactory().createResourceURI(name);
-        Optional<Resource> optionalInputResource = this.getResource(resourceSet, resourceURI, inputStream, applyMigrationParticipants);
-        if (optionalInputResource.isPresent()) {
-            Resource inputResource = optionalInputResource.get();
+        Optional<ExternalResourceLoadingResult> optionalLoadingResult = this.getResource(resourceSet, resourceURI, inputStream, applyMigrationParticipants);
+        if (optionalLoadingResult.isPresent() && optionalLoadingResult.get().resource().isPresent()) {
```
sbegaudeau: To continue on my comment, does it make sense to have a non empty result but with an empty resource? From what I see here, since you are checking both at the same time it doesn't.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/document/services/LoadingReport.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.List;
+
+import org.eclipse.sirius.web.application.document.services.api.ILoadingReport;
+
+/**
+ * A loading report.
+ *
+ * @author gdaniel
+ */
+public record LoadingReport(List<String> report) implements ILoadingReport {
```
sbegaudeau: A report containing a report... I'll probably rename the record component

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/documents/DocumentControllerIntegrationTests.java
sbegaudeau: I'm not really a fan of mockito and everything that feels like magic because it's fine to get started but when you need to debug something, it gets messy. I'll have to look into it.

sbegaudeau (changes requested):

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/document/services/DocumentSanitizedJsonContentProvider.java
gdaniel: My rationale for that was:

- Yes, it does make sense to fail to import a resource and still provide a report to the end user, instead of the generic error message we display right now.
- I didn't know if / how you wanted to display this error, so I just created the data structure for it in case we want it in the future. If we don't do that, then yes, I agree we can have `IExternalResourceLoaderService#getResource` return an `Optional<ExternalResourceLoadingResult>`.

Now that I look back at it, I am not a fan of this half-cooked feature, I should have committed fully on one or the other solution. So I can either remove it and do what you suggest in your previous comment, or make sure a report without a resource is correctly handled and the report is displayed to the end user (we would basically always return a success payload in `UploadDocumentEventHandler`).

Let me know which solution you prefer.

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/documents/DocumentControllerIntegrationTests.java
gdaniel: Before using Mockito I tried to add errors/warnings from the loaded resource in the report, but I couldn't find a JSON/XMI content that produces  errors/warnings without throwing an `IOException`.

I also considered creating a specific implementation of `IExternalResourceLoaderService` for testing purposes, it felt a bit overkill but I guess it's not that complicated to do.

I can do that if it works better for you.

sbegaudeau (approved):

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/documents/DocumentControllerIntegrationTests.java
sbegaudeau: I've added a custom `IUploadDocumentReportProvider` and it does the trick with a narrow check for `canHandle`

PR #6205, opened 2026-02-18
[6202] Prevent compatible nodes from being highlighted when dragging …

### LINKED ISSUE #6202: Although the move is not actually possible, potential targets are highlighted when dragging a borderNode
https://github.com/user-attachments/assets/e954369d-b1f9-43ed-8a7e-dc7dc81459da

As drag and drop of a borderNode is not implemented, targets should not be highlighted.

frouene:
…a border node

Bug: https://github.com/eclipse-sirius/sirius-web/issues/6202

# Pull request template

## General purpose
- [x] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


sbegaudeau:

This PR successfully addresses the issue where compatible nodes were being highlighted as drop targets when dragging a border node. It introduces a filtering mechanism in the frontend and expands the E2E test suite with a new
  scenario to verify this behavior.

  <details>


   * Frontend: Stale Closures in `useDropNodes.ts`: In the onNodesDragStart callback, several dependencies are used but missing from the dependency array. According to the react-frontend-reviewer rule on Effect Dependencies, all dependencies must be correctly listed to prevent stale closures.


   1     // packages/diagrams/frontend/sirius-components-diagrams/src/renderer/dropNode/useDropNodes.ts
   2     const onNodesDragStart: OnNodeDrag<Node<NodeData>> = useCallback(
   3       (_event, _node, nodes) => {
   4         // ... uses diagramDescription, initializeDrop, setNodes, storeApi ...
   5       },
   6       [getNodes] // 🚨 Missing diagramDescription, initializeDrop, setNodes, storeApi
   7     );
       * Action: Add diagramDescription, initializeDrop, setNodes, and storeApi to the dependency array.

  </details>


  <details>


   * Backend: Naming Convention Violation: In DiagramDnDDomainProvider.java, the variable attr is used. According to the java-backend-reviewer rule 9. Naming Conventions, variable names should be meaningful and not shortened to a few characters.


   1     // packages/sirius-web/backend/sirius-web-e2e-tests/src/main/java/org/eclipse/sirius/web/e2e/tests/diagramdnd/DiagramDnDDomainProvider.java
   2     private void addAttribute(Entity entity, String name, DataType type) {
   3         Attribute attr = DomainFactory.eINSTANCE.createAttribute(); // ⚠️ Should be 'attribute'
   4         attr.setName(name);
   5         // ...
   6     }
   * Frontend: Variable Shadowing: In useDropNodes.ts, a local initialPositions is declared within onNodesDragStart, which shadows the initialPositions obtained from the DropNodeContext. This can lead to maintenance confusion. Consider
     renaming the local variable to draggedInitialPositions.


  </details>

  <details>


   * Excellent Test Coverage: The contributor not only fixed the bug but also took the time to rename the spec file for consistency (bordernode.spec.ts -> border-node.spec.ts) and added a comprehensive Playwright test case using
     waitForFunction to verify the CSS box-shadow state.
   * Clean Refactoring: The refactoring of DiagramDnDDomainProvider.java to use private helper methods for entity and relation creation significantly improves the readability of the test data setup.
   * Correct Metadata: Copyright headers were properly updated to include 2026, and the CHANGELOG.adoc was correctly updated.


  </details>

pcdavid (approved):

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/dropNode/useDropNodes.ts
```diff
@@ -154,7 +154,7 @@ export const useDropNodes = (): UseDropNodesValue => {
       if (nodes.length === 0) {
         return;
       }
-      const draggedNodes = nodes.map((node) => getDraggableNode(node));
+      const draggedNodes = nodes.filter((node) => !node.data.isBorderNode).map((node) => getDraggableNode(node));
```
sbegaudeau, line 167: Given that we want to be able to support the drag and drop of border nodes in the future, why would we hardcode something like this? Why would we try to prevent even more the support for this kind of drag and drop while there seems to be an issue with compatible nodes for the drag and drop? Shouldn't the fix be there in the computation of the drag and drop?

sbegaudeau (changes requested):

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/dropNode/useDropNodes.ts
sbegaudeau: I would even go one step further, for me the main thing that needs fixing is the addition of the support of the drag and drop of border nodes.

pcdavid: That was my initial reaction, but given that the position of border nodes is fixed to be around the border of their parent, "support of the drag and drop of border nodes" is not a trivial task (not technically difficult, but how do we want the UI to work when moving a border node which at the same time should stay on it's parent's border and *also* be allowed to be dragged and dropped elsewhere?).

frouene: Yes, it has been discussed, and adding support for DnD of border nodes is too costly right now, which is why we have this quick fix.

sbegaudeau: > Yes, it has been discussed, and adding support for DnD of border nodes is too costly right now, which is why we have this quick fix.

It is 100% the opposite of what was discussed at the beginning of the iteration (where supporting it was exactly the goal). I can understand that we don't have the time to finish properly the feature but it doesn't explain why this kind of fix is needed in the first place. What makes us believe that the border node moved is to be considered for the drag and drop? Are you sure that it does not have a drop node tool? If we have a node (border node or not) without any drop node tool, we shouldn't even consider using it in `useDropNodes()`. What makes the code believe that this one is droppable?

Where is the issue in matching the node and the dropNodeCompatibility data? I'm not buying the "that's a border node so it's false", for me that's where the issue is.

frouene: ReactFlow does not make a distinction between a move and a drag. However, if a borderNode ends up carrying a dropTool (for example, in the context of reuse), the other nodes are updated to highlight compatible nodes, which can be confusing for the user, knowing that it is not possible for a border node to be moved outside its parent.
The idea behind this PR is to no longer update the list of compatible nodes if the dragged node is a border node, in order to avoid any confusion.

sbegaudeau: - So the border node should not have a drop tool in the first place, that's the first issue because with a "fix" like that, we will end up reversing it soon and downstream projects will have to remove their drop tools anyway. So they will have the same problem.
- Even without a drop tool, I can see that the drop node feedback can still be triggered if a border node is moved. So there seems to be a bug somewhere which makes the drop node feedback to be active without any reasons. [Why do I have this feedback?](https://www.youtube.com/watch?v=7u1RiwL9k4g)

If the issue where the drop node feedback can be active even if it's not relevant was fixed and if the drop node tool was removed, the issue would not occur without us having to add some hacks like that and when (not if) we will add the support for drag and drop of border nodes (which should have existed from the very beginning) then no one will have any additional work to do.

frouene: I made the changes so that the parent of a dragged node is no longer considered a target like the others and is no longer highlighted.
But to return to the initial problem, I don't think I can filter in the backend for potential candidates for dragging a borderNode, because in `DropNodeCompatibilityProvider`, we don't list the nodes but their descriptions, and I don't see a solution for knowing whether this node description will be used as a borderNode or not.

sbegaudeau:
I'll wait for the entire build to validate how I've handle the conflicts in the playwright tests

sbegaudeau (approved):

PR #6266, opened 2026-03-09
[6265] Apply layout on selected nodes in useArrangeAll

### LINKED ISSUE #6265: Apply layout on selected nodes in useArrangeAll

### Expected behavior

Allow useArrangeAll to target only a specific groups of nodes if they are selected.

### Cuttings backs

...

### Potential side effects

...

...

### Code cleaning

Removing the resolveNodeOverlap function result assigned to an unused constant (overlapFreeLaidOutNodes), and a .map() iteration that was performed on laidOutNodesWithElk without its result being used

OrlannC:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/6265

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- [x] Variables have a proper type
- [x] Functions’ arguments have a proper type
- [x] Functions’ return type are specified
- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?
The overall behavior hasn't changed much, so you can still use the auto-layout like you always do. However, you can now test the new feature by selecting specific nodes before triggering the layout.

### DIFF
--- a/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/layout/arrange-all/useArrangeAll.ts
+++ b/packages/diagrams/frontend/sirius-components-diagrams/src/renderer/layout/arrange-all/useArrangeAll.ts
@@ -20,6 +20,35 @@ import { useLayout } from '../useLayout';
 import { useSynchronizeLayoutData } from '../useSynchronizeLayoutData';
 import { UseArrangeAllValue } from './useArrangeAll.types';
 
+const getParentPosition = (node: Node<NodeData>, allNodes: Node<NodeData>[]): [number, number] => {
+  if (node.parentId) {
+    const parent = allNodes.find((n) => n.id === node.parentId);
+    if (parent) {
+      return getParentPosition(parent, allNodes);
+    }
+  }
+  return [node.position.x, node.position.y];
+};
+
+const computeSelectionOrigin = (nodes: Node<NodeData>[], allNodes: Node<NodeData>[]): [number, number] => {
+  let minX = Infinity;
+  let minY = Infinity;
+  let minXNode = 0;
+  let minYNode = 0;
+
+  nodes.forEach((node) => {
+    [minXNode, minYNode] = getParentPosition(node, allNodes);
+    if (minXNode < minX) {
+      minX = minXNode;
+    }
+    if (minYNode < minY) {
+      minY = minYNode;
+    }
+  });
+
+  return [minX, minY];
+};
+
 export const useArrangeAll = (): UseArrangeAllValue => {
   const { getNodes, getEdges, setNodes, setEdges } = useReactFlow<Node<NodeData>, Edge<EdgeData>>();
   const { layout } = useLayout();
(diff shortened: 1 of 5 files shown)


sbegaudeau (changes requested):
This seems like a pretty significant change in behavior. Unless I missed something by looking at the code (I haven't tested it yet), anytime someone would try to use "arrange all" with one element selected in the diagram (which happens quite frequently) nothing would happen anymore?

This kind of changes in behavior cannot just be spontaneously contributed like that, they require a discussion first using  a [pitch](https://basecamp.com/shapeup/1.5-chapter-06) to evaluate how this feature would be accessible, what kind of workflow would be impacted, etc. Instinctively I would imagine that interacting with a specific set of diagram elements would use the group palette.

You can't just change existing behavior like that.

OrlannC:
> This seems like a pretty significant change in behavior. Unless I missed something by looking at the code (I haven't tested it yet), anytime someone would try to use "arrange all" with one element selected in the diagram (which happens quite frequently) nothing would happen anymore?
> 
> This kind of changes in behavior cannot just be spontaneously contributed like that, they require a discussion first using a [pitch](https://basecamp.com/shapeup/1.5-chapter-06) to evaluate how this feature would be accessible, what kind of workflow would be impacted, etc. Instinctively I would imagine that interacting with a specific set of diagram elements would use the group palette.
> 
> You can't just change existing behavior like that.

I understand your point, I will put this PR in draft in the meantime

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/layout/arrange-all/useArrangeAll.ts
```diff
@@ -29,10 +44,21 @@ export const useArrangeAll = (): UseArrangeAllValue => {
   const { resolveNodeOverlap } = useOverlap();
   const { fitView } = useFitView();
   const { elkLayout } = useElkLayout();
+  const { isOpened } = useDiagramPalette();
 
   const arrangeAll = async (layoutOptions: LayoutOptions): Promise<void> => {
-    await elkLayout(getNodes(), getEdges(), layoutOptions).then(
+    const selectedNodes: Node<NodeData>[] = getNodes().filter((n) => n.selected);
+    const [minX, minY] = computeSelectionOrigin(selectedNodes);
+    const nodesToLayout: Node<NodeData>[] = !isOpened ? getNodes() : selectedNodes;
```
frouene: You cannot rely on whether the palette is open or closed to determine whether to apply arrangeAll only to the current selection. It is up to the caller of arrangeAll to specify the scope of application.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/layout/arrange-all/useArrangeAll.ts
```diff
@@ -21,6 +21,21 @@ import { RawDiagram } from '../layout.types';
 import { useLayout } from '../useLayout';
 import { useSynchronizeLayoutData } from '../useSynchronizeLayoutData';
 import { UseArrangeAllValue } from './useArrangeAll.types';
+import { useDiagramPalette } from '../../palette/useDiagramPalette';
+
+const computeSelectionOrigin = (nodes: Node<NodeData, string>[]): [number, number] => {
```
frouene: If I understand this part correctly, you apply a kind of offset to the positions so that the selected elements are not placed anywhere in the diagram.
The intention is good, but I don't understand what happens when it's used; my selection group moves. I wonder if this is related to the fact that certain positions are relative to the parent, whereas in the code they are all compared to each other without distinction of hierarchy.

frouene (changes requested):

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/layout/arrange-all/useArrangeAll.ts
OrlannC: The selected nodes were originally being reset to (0,0). To fix this, I captured their initial x and y coordinates so I could restore their positions before resolveOverlap is called

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/layout/arrange-all/useArrangeAll.ts
OrlannC: I anticipated that. I’ve already moved the selection logic into a useArrangeSelection hook used only in groups. Should I push it?

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/layout/arrange-all/useArrangeAll.types.ts
```diff
@@ -13,5 +13,5 @@
 import { LayoutOptions } from 'elkjs/lib/elk-api';
 
 export interface UseArrangeAllValue {
-  arrangeAll: (layoutOptions: LayoutOptions) => Promise<void>;
+  arrangeAll: (layoutOptions: LayoutOptions, selectedNodesIds?: string[]) => Promise<void>;
```
frouene: This new variable should not be called “selectedNodesIds”; the fact that the list of nodes to be arranged corresponds to the current selection is a use case specific to the group palette, but we could imagine a future feature where the selection would have no relation to the group of nodes to be arranged.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/layout/arrange-all/useArrangeAll.ts
frouene: This behavior is always problematic when you have two nodes on different references

frouene: https://github.com/user-attachments/assets/7061ae7b-bf13-4a45-82b4-1416f72edece

PR #6378, opened 2026-04-03
[6377] Add frontend support for rendering node decorators

### LINKED ISSUE #6377: Add frontend support for node decorators

Sirius Web's frontend should be able to render node decorators.

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/6377

Fixes #6377

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/integration-tests-cypress/cypress/e2e/project/impactanalysis/diagram-impact-analysis.cy.ts
+++ b/integration-tests-cypress/cypress/e2e/project/impactanalysis/diagram-impact-analysis.cy.ts
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2025 Obeo.
+ * Copyright (c) 2025, 2026 Obeo.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
@@ -44,7 +44,7 @@ describe('Impact analysis - diagram', () => {
         details.getTextField('Name').type('InitialName{enter}');
         new Explorer().createRepresentation('Root', diagramDescriptionName, diagramTitle);
 
-        diagram.getNodes('Impact analysis diagram', 'InitialName').findByTestId('Label - InitialName').rightclick();
+        diagram.getNodes('Impact analysis diagram', 'InitialName').rightclick();
         diagram.getPalette().should('exist');
         diagram.getPalette().findByTestId('tool-Tool with impact analysis').click();
         impactAnalysis.getImpactAnalysisDialog().should('exist');
--- a/packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
+++ b/packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
@@ -125,6 +125,25 @@ type Node {
   deletable: Boolean!
   customizedStyleProperties: [NodeStyleProperty!]!
   initialBorderNodePosition: BorderNodePosition!
+  decorators: [NodeDecorator!]!
+}
+
+type NodeDecorator {
+  label: String!
+  position: NodeDecoratorPosition!
+  iconURL: String!
+}
+
+enum NodeDecoratorPosition {
+  NORTH
+  NORTH_EAST
+  EAST
+  SOUTH_EAST
+  SOUTH
+  SOUTH_WEST
+  WEST
+  NORTH_WEST
+  CENTER
 }
 
 enum NodeStyleProperty {
--- a/packages/diagrams/backend/sirius-components-diagrams-graphql/src/main/java/org/eclipse/sirius/components/diagrams/graphql/datafetchers/diagram/NodeDecoratorsDataFetcher.java
+++ b/packages/diagrams/backend/sirius-components-diagrams-graphql/src/main/java/org/eclipse/sirius/components/diagrams/graphql/datafetchers/diagram/NodeDecoratorsDataFetcher.java
@@ -1,0 +1,87 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.components.diagrams.graphql.datafetchers.diagram;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.eclipse.sirius.components.annotations.spring.graphql.QueryDataFetcher;
+import org.eclipse.sirius.components.diagrams.Node;
+import org.eclipse.sirius.components.diagrams.NodeDecorator;
+import org.eclipse.sirius.components.diagrams.NodeDecoratorPosition;
+import org.eclipse.sirius.components.graphql.api.IDataFetcherWithFieldCoordinates;
+import org.springframework.beans.factory.annotation.Value;
+
+import graphql.execution.DataFetcherResult;
+import graphql.schema.DataFetchingEnvironment;
+
+/**
+ * Data fetcher returning decorators for a node (hardcoded for now).
+ *
+ * @author gdaniel
+ */
+@QueryDataFetcher(type = "Node", field = "decorators")
+public class NodeDecoratorsDataFetcher implements IDataFetcherWithFieldCoordinates<DataFetcherResult<List<NodeDecorator>>> {
+
+    private static final String PRIVATE = "/icons/papaya/full/ovr16/Visibility_PRIVATE.svg";
+
+    private static final String PROTECTED = "/icons/papaya/full/ovr16/Visibility_PROTECTED.svg";
+
+    private static final String PUBLIC = "/icons/papaya/full/ovr16/Visibility_PUBLIC.svg";
+
+    private static final String PACKAGE = "/icons/papaya/full/ovr16/Visibility_PACKAGE.svg";
+
+    private final boolean enabledDecorators;
+
+    public NodeDecoratorsDataFetcher(@Value("${sirius.components.diagram.enableDecorators:false}") boolean enableDecorators) {
+        this.enabledDecorators = enableDecorators;
+    }
+
+    @Override
+    public DataFetcherResult<List<NodeDecorator>> get(DataFetchingEnvironment environment) throws Exception {
+        List<NodeDecorator> decorators = new ArrayList<>();
+        Node node = environment.getSource();
+
+        Map<String, Object> parentLocalContext = environment.getLocalContext();
+        Map<String, Object> localContext = new HashMap<>();
+        localContext.put("diagram", parentLocalContext.get("diagram"));
+
+        if (this.enabledDecorators) {
+            if (node != null && node.getTargetObjectLabel() != null) {
+                if (node.getTargetObjectLabel().contains("Decorator")) {
+                    decorators.add(new NodeDecorator("North Decorator", NodeDecoratorPosition.NORTH, PUBLIC));
+                    decorators.add(new NodeDecorator("North Decorator 2", NodeDecoratorPosition.NORTH, PRIVATE));
+                    decorators.add(new NodeDecorator("North East Decorator", NodeDecoratorPosition.NORTH_EAST, PRIVATE));
+                    decorators.add(new NodeDecorator("East Decorator", NodeDecoratorPosition.EAST, PROTECTED));
+                    decorators.add(new NodeDecorator("South East Decorator", NodeDecoratorPosition.SOUTH_EAST, PACKAGE));
+                    decorators.add(new NodeDecorator("South Decorator", NodeDecoratorPosition.SOUTH, PUBLIC));
+                    decorators.add(new NodeDecorator("South West Decorator", NodeDecoratorPosition.SOUTH_WEST, PUBLIC));
+                    decorators.add(new NodeDecorator("West Decorator", NodeDecoratorPosition.WEST, PROTECTED));
+                    decorators.add(new NodeDecorator("North West Decorator", NodeDecoratorPosition.NORTH_WEST, PRIVATE));
+                    decorators.add(new NodeDecorator("Center Decorator", NodeDecoratorPosition.CENTER, PACKAGE));
+                } else if (node.getTargetObjectLabel().contains("Papaya")) {
+                    decorators.add(new NodeDecorator("Papaya", NodeDecoratorPosition.NORTH, "/project-templates/Cosmic-Papaya.jpeg"));
+                    decorators.add(new NodeDecorator("Papaya2", NodeDecoratorPosition.EAST, "/project-templates/Retro-Papaya.jpeg"));
+                    decorators.add(new NodeDecorator("Papaya3", NodeDecoratorPosition.WEST, PUBLIC));
+                    decorators.add(new NodeDecorator("Papaya4", NodeDecoratorPosition.SOUTH_WEST, PRIVATE));
+                }
+            }
+        }
+        return DataFetcherResult.<List<NodeDecorator>> newResult()
+                .data(decorators)
+                .localContext(localContext)
+                .build();
+    }
+}
--- a/packages/diagrams/backend/sirius-components-diagrams/src/main/java/org/eclipse/sirius/components/diagrams/NodeDecorator.java
+++ b/packages/diagrams/backend/sirius-components-diagrams/src/main/java/org/eclipse/sirius/components/diagrams/NodeDecorator.java
@@ -1,0 +1,29 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.components.diagrams;
+
+import java.util.Objects;
+
+/**
+ * A node decorator.
+ *
+ * @author gdaniel
+ */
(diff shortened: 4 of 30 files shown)


packages/diagrams/backend/sirius-components-diagrams-graphql/src/main/java/org/eclipse/sirius/components/diagrams/graphql/datafetchers/diagram/NodeDecoratorsDataFetcher.java
```diff
@@ -0,0 +1,138 @@
...
+import graphql.execution.DataFetcherResult;
+import graphql.schema.DataFetchingEnvironment;
+
+/**
+ * Data fetcher returning decorators for a node (hardcoded for now).
+ *
+ * @author gdaniel
+ */
+@QueryDataFetcher(type = "Node", field = "decorators")
+public class NodeDecoratorsDataFetcher implements IDataFetcherWithFieldCoordinates<DataFetcherResult<List<NodeDecorator>>> {
```
gdaniel, line 36: As discussed, this PR introduces this data fetcher as a temporary solution to test the frontend code.
It will be removed in a future PR.
For now, you can enable it with the `sirius.components.diagram.enableDecorators=true` property, and rename a node to `Decorator` or `Papaya` to add decorators.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/actions/ActionsContainer.tsx
```diff
@@ -23,6 +23,8 @@ const useStyles = makeStyles()((theme) => ({
   actionsContainer: {
     gridRow: 1,
     gridColumn: 1,
+    width: 'fit-content',
+    justifySelf: 'flex-end',
```
gdaniel, line 27: This ensures that `ActionContainer` doesn't take all the horizontal space in its parent.
Otherwise, it covers the entire label, and since it is on top of the decorator, it effectively de-activates all the top decorators

Image

With this change the action div only takes as much space as it needs, so it only covers the decorators that are actually covered by actions

Image

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/decorators/Decorator.tsx
```diff
@@ -0,0 +1,116 @@
...
+    alignSelf: 'start',
+  },
+  centerDecorator: {
+    gridArea: '2 / 2',
+    justifySelf: 'center',
+    alignSelf: 'center',
+  },
+}));
+
+export const Decorator = ({ position, decorators }: DecoratorProps) => {
```
gdaniel, line 68: This whole component feels a bit sketchy, but I didn't a find a nicer way to bind the CSS class to the position of the  decorator.

Note that at the moment only the first decorator in the list is rendered, in a future PR, we'll use the entire list to 
1. Render the first decorator
2. Render all the other in a tooltip

For now, only the label of the first decorator is rendered in the tooltip.

gdaniel:
I didn't add any safeguard to improve the rendering when decorators with different sizes are provided.

I experimented on what can be done, but ultimately decided to keep it as is. You can have a look by renaming a node to `Papaya` in any diagram, and check how the papaya decorators (large images) and the visibility decorators (small images) work together.

If we want some safeguards, we should probably do more specification on that, and decide what we want to achieve, for example:
- Always see as many decorators as possible
- Always expand images as much as possible
- Resize images to fit the available space
- etc

scripts/check-coverage.jsh
gdaniel, line 1: Lowered the expected code coverage on `sirius-components-diagrams` and `sirius-components-diagrams-graphql`. 
These should be reset once we have the entire feature implemented and the temporary data fetchers removed.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/DiagramRenderer.types.ts
```diff
@@ -88,6 +89,24 @@ export enum BorderNodePosition {
...
+  NORTH,
+  NORTH_EAST,
+  EAST,
+  SOUTH_EAST,
+  SOUTH,
+  SOUTH_WEST,
+  WEST,
+  NORTH_WEST,
+  CENTER,
+}
```
sbegaudeau: Don't use enum in TypeScript, they do not behave the way you think, use an union instead

packages/diagrams/backend/sirius-components-diagrams/src/main/java/org/eclipse/sirius/components/diagrams/Node.java
```diff
@@ -257,6 +263,8 @@ public static final class Builder {
 
         private Set<String> customizedStyleProperties;
 
+        private List<NodeDecorator> decorators = List.of();
```
sbegaudeau: Don't initialize anything for builders. I suspect that you don't need any changes in this file for the PR to work. This change should be done in the next step.

packages/diagrams/backend/sirius-components-diagrams/src/main/java/org/eclipse/sirius/components/diagrams/NodeDecorator.java
```diff
@@ -0,0 +1,89 @@
...
+        return this.label;
+    }
+
+    public NodeDecoratorPosition getPosition() {
+        return this.position;
+    }
+
+    public String getIconURL() {
+        return this.iconURL;
+    }
```
sbegaudeau: This part can be a record

packages/diagrams/frontend/sirius-components-diagrams/src/converter/convertDecorators.ts
```diff
@@ -0,0 +1,68 @@
...
+    case 'WEST':
+      decoratorPosition = NodeDecoratorPosition.WEST;
+      break;
+    case 'NORTH_WEST':
+      decoratorPosition = NodeDecoratorPosition.NORTH_WEST;
+      break;
+    case 'CENTER':
+      decoratorPosition = NodeDecoratorPosition.CENTER;
+      break;
+  }
```
sbegaudeau: If you don't use an enum but instead an union, you don't have anything to convert here

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
```diff
@@ -124,6 +124,25 @@ type Node {
   deletable: Boolean!
   customizedStyleProperties: [NodeStyleProperty!]!
   initialBorderNodePosition: BorderNodePosition!
+  decorators: [NodeDecorator!]!
+}
+
+type NodeDecorator {
+  label: String!
```
sbegaudeau, line 132: I suspect that you will need an identifier on the frontend to distinguish two decorators at the same position otherwise React will have an issue with a missing key property

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/DiagramRenderer.types.ts
```diff
@@ -88,6 +89,24 @@ export enum BorderNodePosition {
...
+
+export interface NodeDecorator {
+  label: string;
+  iconURL: string;
+  position: NodeDecoratorPosition;
+}
```
sbegaudeau: Why do you need a new type instead of the `GQLNodeDecorator`? Does it bring anything?

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/node/ListNode.tsx
```diff
@@ -41,6 +42,7 @@ const listNodeStyle = (
   const listNodeStyle: React.CSSProperties = {
     width: '100%',
     height: '100%',
+    position: 'relative',
```
sbegaudeau, line 45: Why this on this type of node and not the other ones?

packages/diagrams/frontend/sirius-components-diagrams/src/converter/convertDecorators.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+import { GQLNodeDecorator, GQLNodeDecoratorPosition } from '../graphql/subscription/nodeFragment.types';
+import { NodeDecorator, NodeDecoratorPosition } from '../renderer/DiagramRenderer.types';
+
+export const convertDecorators = (decorators: GQLNodeDecorator[]): Map<NodeDecoratorPosition, NodeDecorator[]> => {
```
sbegaudeau: If I go one step further than my comments, I don't know why this exists. The transformation from a list to a map can be done in DecoratorContainer and there's one thing less to do in every single node types.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/decorators/Decorator.tsx
```diff
@@ -0,0 +1,116 @@
...
+    case NodeDecoratorPosition.WEST:
+      cssClass = classes.westDecorator;
+      break;
+    case NodeDecoratorPosition.NORTH_WEST:
+      cssClass = classes.northWestDecorator;
+      break;
+    case NodeDecoratorPosition.CENTER:
+      cssClass = classes.centerDecorator;
+      break;
+  }
```
sbegaudeau: This entire function could become something like this: `classes[position.toLowerCase()]`

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/decorators/DecoratorContainer.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { DecoratorContainerProps } from './DecoratorContainer.types';
+
+const useStyles = makeStyles()((_) => ({
+  decoratorContainer: {
+    position: 'absolute',
+    top: 0,
+    left: 0,
+    display: 'grid',
+    gridTemplateRows: '1fr 1fr 1fr',
+    gridTemplateColumns: '1fr 1fr 1fr',
```
sbegaudeau, line 30: Could you use named grid areas, it would greatly improve the understanding of the positions of the decorators

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/decorators/DecoratorContainer.tsx
```diff
@@ -0,0 +1,48 @@
...
+    height: '100%',
+    width: '100%',
+    padding: '1px',
+    gap: '1px',
```
sbegaudeau: Why a gap?

sbegaudeau (changes requested):

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/decorators/DecoratorContainer.tsx
gdaniel: Removed, and padding also removed

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/decorators/DecoratorContainer.tsx
gdaniel: Done, the grid areas are named `N`, `S`, `SW`, etc

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/decorators/Decorator.tsx
gdaniel: Done, as suggested, I removed the `NodeDecorator` type and used the `GQLNodeDecoratorPosition` union as a key to find the appropriate class. Thanks for the tip

packages/diagrams/frontend/sirius-components-diagrams/src/converter/convertDecorators.ts
gdaniel: Done, I removed this file, `NodeDecorator`, and `NodeDecoratorPosition` types. Each node type receives a `GQLNodeDecorator[]` and passes it to the `DecoratorContainer`, which converts it into a map

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/node/ListNode.tsx
gdaniel: `position: relative` ensures the absolute positioning of the decorator grid takes into account the borders of the node.
As you suggested, there was a couple other places where this change should have been done, I added them.
BTW, I didn't update `IconLabelNode` since it doesn't support borders.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/DiagramRenderer.types.ts
gdaniel: I assumed we didn't want to have `GQLXXX` types passed the `DiagramRenderer`,  but based on your comment my assumption was wrong 😅
Since it didn't bring anything there was no reason to keep it around, I removed it.

packages/diagrams/backend/sirius-components-collaborative-diagrams/src/main/resources/schema/diagram.graphqls
gdaniel: 2 decorators at the same position should only render 1 thing (the first decorator), so the key could be the position itself.
In the next PR we'll add a tooltip to display all the decorators on a given position. We will most likely need an ID then, but I don't think we need it now.
Anyways, if you prefer, I can add it now as part of the `NodeDecorator` type and use it later.

packages/diagrams/frontend/sirius-components-diagrams/src/converter/convertDecorators.ts
gdaniel: Fixed

packages/diagrams/backend/sirius-components-diagrams/src/main/java/org/eclipse/sirius/components/diagrams/NodeDecorator.java
gdaniel: Done, I also removed the builder.
I don't know what's the rule with builders in these classes. I can see some classes with builders, some without builders. Since `NodeDecorator` is relatively simple I decided to remove the builder. Let me know if I should change this.

packages/diagrams/backend/sirius-components-diagrams/src/main/java/org/eclipse/sirius/components/diagrams/Node.java
gdaniel: There is no change in the `Node` class anymore.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/DiagramRenderer.types.ts
gdaniel: Done, this enum doesn't exist anymore

integration-tests-cypress/cypress/e2e/project/impactanalysis/diagram-impact-analysis.cy.ts
```diff
@@ -44,7 +44,7 @@ describe('Impact analysis - diagram', () => {
         details.getTextField('Name').type('InitialName{enter}');
         new Explorer().createRepresentation('Root', diagramDescriptionName, diagramTitle);
 
-        diagram.getNodes('Impact analysis diagram', 'InitialName').findByTestId('Label - InitialName').rightclick();
+        diagram.getNodes('Impact analysis diagram', 'InitialName').rightclick();
```
gdaniel, line 47: Fixed the cypress test: I suspect cypress wasn't able to rightclick the div because it is entirely covered by the decorator div.
Right clicking on the node itself (instead of the label) opens the palette the same way, so no need to check for the label

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/node/IconLabelNode.tsx
```diff
@@ -80,6 +81,7 @@ export const IconLabelNode: NodeComponentsMap['iconLabelNode'] = memo(
         onDragOver={onDragOver}
         onDrop={handleOnDrop}
         data-testid={`IconLabel - ${data?.insideLabel?.text}`}>
+        <DecoratorContainer decorators={data.decorators} />
```
frouene, line 84: In a previous comment you suggest that IconLabel does not support decorator, why add this container?

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/decorators/Decorator.tsx
```diff
@@ -0,0 +1,86 @@
...
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+import { ServerContext, ServerContextValue } from '@eclipse-sirius/sirius-components-core';
+import Tooltip from '@mui/material/Tooltip';
+import { useContext } from 'react';
+import { makeStyles } from 'tss-react/mui';
+import { GQLNodeDecorator } from '../../graphql/subscription/nodeFragment.types';
+import { DecoratorProps } from './Decorator.types';
+
+const useStyles = makeStyles()((_) => ({
```
frouene, line 20: Why add the `_`?

gdaniel: It's used in other places, I just copied it there. 
I don't know what's the guideline for that in Sirius Web, to me it looks ok to specify that there is an unused parameter.

packages/diagrams/frontend/sirius-components-diagrams/src/renderer/node/IconLabelNode.tsx
gdaniel: If you are referring to this comment:

> position: relative ensures the absolute positioning of the decorator grid takes into account the borders of the node.
> As you suggested, there was a couple other places where this change should have been done, I added them.
BTW, I didn't update IconLabelNode since it doesn't support borders.

`IconLabelNode` doesn't support borders, so there is no reason to add the `position: relative` I added in other nodes.
Even if it doesn't support borders, it supports decorators, see the example below

image

sbegaudeau (approved):

PR #6634, opened 2026-06-29
[6632] Increase the default width of the palette

### LINKED ISSUE #6632: Increase the default width of the palette
It is currently [hard-coded to 200px](https://github.com/eclipse-sirius/sirius-web/blob/master/packages/palette/frontend/sirius-components-palette/src/Palette.tsx#L50).

At least in the context of SysON, it would be nicer to have a wider palette. It would avoid a horizontal scroll on the quick actions toolbar (maybe some of these could be removed, but still), and ellipsis on many of the tools for common elements:

Image

Image

If we do not want to change the default in Sirius Web itself, maybe we could use a CSS variable instead of a hard-coded value, so that SysON (or other downstream apps) can override it.

pcdavid:
### Before (200px)

Even a default action like "Delete from model" can not be shown fully.

image

### After (250px)

Capture d’écran du 2026-06-29 16-32-05

Bug: https://github.com/eclipse-sirius/sirius-web/issues/6632

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/packages/palette/frontend/sirius-components-palette/src/Palette.tsx
+++ b/packages/palette/frontend/sirius-components-palette/src/Palette.tsx
@@ -47,7 +47,7 @@ export const isPaletteDivider = (entry: GQLPaletteDivider): entry is GQLToolSect
 
 export const isTool = (entry: GQLPaletteEntry): entry is GQLTool => !isPaletteDivider(entry) && !isToolSection(entry);
 
-const paletteWidth = 200;
+const paletteWidth = 250;
 
 /**
  * The palette component forward both the ref and props
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -84,7 +84,7 @@ This is to make it easier to type when the area overlaps another elements.
 - https://github.com/eclipse-sirius/sirius-web/issues/6576[#6576] [diagram] Add `paletteToolOverrideExtensionPoint` to override tools retrieved from the backend.
 - https://github.com/eclipse-sirius/sirius-web/issues/825[#825] [doc] Add a first version of the Sirius Web documentation.
 The documentation can be built using `./doc/generate-local.bash`
-
+- https://github.com/eclipse-sirius/sirius-web/issues/6632[#6632] [diagram] The width of the palette has been increased from 200px to 250px.
 
 === Improvements
 


packages/sirius-web/frontend/sirius-web/src/variables.css
```diff
@@ -0,0 +1,3 @@
+:root {
```
AxelRICHARD: copyright is missing (see other .css files in the same folder)

pcdavid: Done.

packages/palette/frontend/sirius-components-palette/src/Palette.tsx
```diff
@@ -100,7 +98,7 @@ export const Palette = React.forwardRef<HTMLDivElement, PaletteProps & React.HTM
           borderRadius: '10px',
           zIndex: 5,
           position: 'fixed',
-          width: paletteWidth,
+          width: 'var(--paletteWidth, 250)',
```
sbegaudeau: I don't mind hardcoding the increase directly, I don't mind providing a mechanism to configure that either but right now is there anyone who ask for this to be configurable? I would rather just have the new size by default because if someone ask for this to be configurable, the next minute they will want it to be configurable by representation and so on and the entire thing will have to be configurable on the backend as usual.

We already have the theme to configure the appearance, extension points on the frontend, tons of other mechanisms on the backend. I don't think we need yet another new way to configure something for a customization that no-one is requirering. If it becomes necessary, adding an expression on Palette on the View DSL would be the best approach (existing mechanism and way more powerful).

sbegaudeau (changes requested):
This will need to be configurable

sbegaudeau (changes requested):
Image


There is an issue with some fields of the appearance in the palette. I suspect that in `RectangularNodePart`, `ImageNodePart`, `LabelAppearancePart`, `EdgeAppearancePart` and other appearance parts, the ListItem should have a layout (e.g. flex) which should stretch its content horizontally. The `width: '100%'` is not a viable approach and should be removed.

sbegaudeau (approved):
We will fix the minor regression between RC1 and RC2

Issue #6756, opened 2026-07-31
Forms should have a service to be created programatically


pdeville2:
* [x] **I have checked that this feature has not yet been suggested by someone else.**

### Expected behavior
As there are Services to create Diagrams or Tables (`org.eclipse.sirius.components.collaborative.diagrams.DiagramCreationService` and `org.eclipse.sirius.components.collaborative.tables.services.TableCreationService`, I was expected a service to create Froms but I didn't find it.

### Cuttings backs
-

### Potential side effects
None

### Describe alternatives you've considered
I have created this API on my side but it could be in sirius-web itself, I will propose a patch.

PR #6757, opened 2026-07-31
[6756] Add a Service to create Forms

pdeville2:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/6756

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?
This service could be use in `org.eclipse.sirius.components.collaborative.forms.handlers.CreateFormEventHandler` but if I had a new service in the constructor I get the following checkstyle error: "La méthode ou le constructeur a plus de 7 paramètres (trouvé 8). [ParameterNumber]" so I unfortunately could not do it. I could add it as @autowired field but it seems not recommended neither, what do you prefer ?

### DIFF
--- a/packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/api/IFormCreationService.java
+++ b/packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/api/IFormCreationService.java
@@ -1,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) 2026 CEA LIST.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     CEA LIST - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.sirius.components.collaborative.forms.api;
+
+import org.eclipse.sirius.components.forms.Form;
+import org.eclipse.sirius.components.forms.description.FormDescription;
+
+/**
+ * Service used to create forms from scratch.
+ *
+ * @author pdeville
+ */
+public interface IFormCreationService {
+
+    /**
+     * Creates a new form using the given parameters.
+     *
+     * @param targetObject
+     *            The object used as the target
+     * @param formDescription
+     *            The description of the form
+     * @return A new form
+     */
+    Form create(Object targetObject, FormDescription formDescription);
+
+}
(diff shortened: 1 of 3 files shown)


packages/forms/backend/sirius-components-collaborative-forms/src/main/java/org/eclipse/sirius/components/collaborative/forms/api/IFormCreationService.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.eclipse.sirius.components.forms.Form;
+import org.eclipse.sirius.components.forms.description.FormDescription;
+
+/**
+ * Service used to create forms from scratch.
+ *
+ * @author pdeville
+ */
+public interface IFormCreationService {
```
sbegaudeau, line 23: Given our recent work on the improvement of the rendering lifecycle, we will end up needing a service like this soon but it has to be used somewhere. We can't merge dead code.

Right now, there are two parts where this code may seem to be relevant `CreateFormEventHandler#handle` and `FormEventProcessor#refreshForm`.

#### Form Creation

In `CreateFormEventHandler`, such as service would not be very relevant given that the form we are creating there is useless and should be removed in the future. We just need the form metadata in the database, we don't really need the form content at all (which is why it is initialized as an empty and thus useless form).

It will be the same for all fully synchronized representations without any graphical data to persist (i.e. in some use cases transient representations would behave the same way).


#### Form Refresh

In `FormEventProcessor#refreshForm`, we will end up using a service like this one in the near future. You may see that `FormEventProcessor#refresh` is now deprecated given that we want to remove the responsibility of the refresh of the form outside of `FormEventProcessor`, which will require a new service to handle this responsibility.

For that, we will need a `FormRefresher` similar to `HierarchyRefresher` which will leverage a `IFormCreationService` just like `HierarchyRefresher` uses `IHierarchyCreationService`.

As such, the signature of a usable `IFormCreationService` must be:

```
public interface IFormCreationService {
    Form create(IEditingContext editingContext, FormDescription formDescription, Object targetObject, FormContext formContext);
}
```

It's implementation should thus leverage the `FormRenderer` to compute the entire form.

This work will be applied to every single representation event processor in the near future. It will help us separate various concerns that are (too) tightly mixed together in the various representation event processors and it will allow us to have the same logic during the creation of a persisted representation and its refresh (e.g. `CreateHierarchyEventHandler` and `HierarchyRefresher` are both leveraging `IHierarchyCreationService`).

We started this sizable refactoring with the hierarchy part given that it is not that frequently used so any issue would have a smaller impact but this new approach will be used for all representations. You can compare the complexity (and thus coupling) of `HierarchyEventProcessor` and `DiagramEventProcessor` to see the knot that we are trying to untie.

If you want to move forward with this work, you will need to update `FormEventProcessor` to introduce a `FormRefresher` and thus help untying the knot of responsibilities in the form rendering.

sbegaudeau (changes requested):

sbegaudeau:
I will close this PR since it has not been updated and that in the meantime the improvements of the lifecycle of form representations are [in progress](https://github.com/eclipse-sirius/sirius-web/pull/6927) and they will include the `FormCreationService` along with other improvements that I've detailed in my comment above.

pdeville2:
Ok thank you and sorry to not have take time to continue this according to you comments.

PR #6762, opened 2026-08-05
[6761] Extract the explorer subscription in a dedicated component

### LINKED ISSUE #6761: Extract the explorer subscription inside a dedicated component

mcharfadi:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/6761

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -94,6 +94,7 @@ This will allow us to reuse the change description to propagate more information
 - https://github.com/eclipse-sirius/sirius-web/issues/6688[#6688] [diagram] Use outside label height for child node layout in freeform diagrams
 - https://github.com/eclipse-sirius/sirius-web/issues/6753[#6753] [sirius-web] Allow project semantic data to be deleted by project identifier and name.
 - https://github.com/eclipse-sirius/sirius-web/issues/6763[#6763] [sirius-web] Allow project semantic data to be renamed.
+- https://github.com/eclipse-sirius/sirius-web/issues/6761[#6761] [tree] Extract the explorer subscription in a dedicated component in order to reduce the number of rerender.
 
 
 
--- a/packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerSubscriptionProvider.tsx
+++ b/packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerSubscriptionProvider.tsx
@@ -1,0 +1,380 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Obeo.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Obeo - initial API and implementation
+ *******************************************************************************/
+import {
+  RepresentationLoadingIndicator,
+  Selection,
+  SelectionEntry,
+  useSelection,
+} from '@eclipse-sirius/sirius-components-core';
+import {
+  FilterBar,
+  GQLGetTreePathVariables,
+  GQLTree,
+  GQLTreeItem,
+  TreeFilter,
+  TreeToolBar,
+  TreeToolBarContext,
+  TreeToolBarContextValue,
+  TreeView,
+  useTreeFilters,
+  useTreePath,
+  useTreeSelection,
+} from '@eclipse-sirius/sirius-components-trees';
+import Box from '@mui/material/Box';
+import { Theme } from '@mui/material/styles';
+import { useCallback, useContext, useEffect, useRef, useState } from 'react';
+import { makeStyles } from 'tss-react/mui';
+import { DuplicateObjectKeyboardShortcut } from './context-menu-contributions/duplicate-object/DuplicateObjectKeyboardShortcut';
+import {
+  ExplorerSubscriptionProviderProps,
+  ExplorerSubscriptionProviderState,
+} from './ExplorerSubscriptionProvider.types';
+import { TreeDescriptionsMenu } from './TreeDescriptionsMenu';
+import { useExplorerSubscription } from './useExplorerSubscription';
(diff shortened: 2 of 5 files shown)


CHANGELOG.adoc
```diff
@@ -93,6 +93,7 @@ Downstream applications can implement `IOmniboxCommandOrdererDelegate` to define
 This will allow us to reuse the change description to propagate more information.
 - https://github.com/eclipse-sirius/sirius-web/issues/6688[#6688] [diagram] Use outside label height for child node layout in freeform diagrams
 - https://github.com/eclipse-sirius/sirius-web/issues/6753[#6753] [sirius-web] Allow project semantic data to be deleted by project identifier and name.
+- https://github.com/eclipse-sirius/sirius-web/issues/6761[#6761] [trees] Extract the explorer subscription in a dedicated component in order to reduce the number of rerender.
```
frouene: There are other uses for “trees,” but usually we tend to use the singular form, as with “diagram,” for example.

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerView.tsx
```diff
@@ -347,67 +94,19 @@ export const ExplorerView = forwardRef<WorkbenchViewHandle, WorkbenchViewCompone
...
-                    data-testid="explorer://"
-                    useTreePalette={state.tree.capabilities.useTreePalette}
-                  />
-                </div>
-              </DuplicateObjectKeyboardShortcut>
-            </>
-          )}
-        </Box>
+        {treeFiltersLoading || descriptionLoading ? (
+          <RepresentationLoadingIndicator />
```
frouene, line 72: `RepresentationLoadingIndicator` is present in this file and in the `ExplorerSubscriptionProvider`

frouene (changes requested):

frouene:
I've found a regression, when switching the explorer the filters are not reload.
This can be seen with the custom explorer in domain

mcharfadi:
> I've found a regression, when switching the explorer the filters are not reload. This can be seen with the custom explorer in domain

Yes I'll push a fix for that, thanks

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerView.tsx
mcharfadi: Yes it's on purpose one for waiting for the queries and another for the subscription
It should behave like before since we're waiting for the subscription at the end.

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerSubscriptionProvider.tsx
```diff
@@ -0,0 +1,380 @@
...
+    id: gqlFilter.id,
+    label: gqlFilter.label,
+    state: gqlFilter.defaultState,
+  };
+};
+
+const isTreeRefreshedEventPayload = (payload: GQLTreeEventPayload): payload is GQLTreeRefreshedEventPayload =>
+  payload && payload.__typename === 'TreeRefreshedEventPayload';
+
+export const ExplorerSubscriptionProvider = ({
```
sbegaudeau, line 74: I find it quite odd that this `ExplorerSubscriptionProvider` (by the way, it provides the subscription to what? why not `ExplorerSubscriptionContainer`?) contains that much code for so many unrelated features such as:

- The internal organization of the explorer with `TreeToolBar`, `DuplicateObjectKeyboardShortcut`, `TreeView`
- Some CSS to layout things
- The internal lifecycle of the tree filters
- The filtering of the explorer
- The management of the selection

I don't understand why all these unrelated concerns are mixed in a component which should just be responsible of the subscription. This component should be as small as possible and look like this:

```
const ExplorerSubscriptionContainer = ({ ..., children }) => {
  const { payload } = useExplorerSubscription(...);

  return (
    <ExplorerContextProvider>
      {children}
    </ExplorerContextProvider>
  );
}
```

It has no reason whatsoever to care about the internal organization of the explorer. With so many responsibilities intertwined together, it's not a surprise that the PR is too big. I think that to move forward with this task, you need to:

- Extract some low level concerns into dedicated hooks or in their relevant components. For example, the state of the filter + the call to `useTreeFilters` + the relevant use effects should be one dedicated hook like `useTreeFiltering` which would encapsulate the entire feature. Another one could be the management of the selection with its state etc. All these useEffects from `ExplorerView` should be moved out of the way. This should be easy to extract, review and merge.
- Then you can contribute an `ExplorerSubscriptionContainer` which should only care about the subscription

I don't see how you can contribute a PR such as #6766 given that this current PR cannot be merged as is. It is just more expensive and complex to contribute new PRs while the previous ones have not been merged. Any refactoring in this PR will create a cost of rebase for downstream PRs. And adding additional responsibilities (i.e. the synchronization with the local storage) to a component dedicated to the management of the subscription is not the way to proceed.

One of our most critical requirement in Sirius Web is to lower the responsibilities of our concepts, not add new ones.

sbegaudeau (changes requested):

packages/sirius-web/frontend/sirius-web-application/src/views/edit-project/workbench-views/explorer/ExplorerSubscriptionProvider.tsx
mcharfadi: I agree that the naming is not perfect right now.

The main issue I have with the component `ExplorerView ` is that it starts rendering when we don't have a valid `treeDescriptionId`.

As such this PR allows us to work with a coherent `ExplorerView ` that can effectively starts a subscription at the "right" time.

The next step I wanted to implement was indeed to also extract the TreeFilters (since we also want them for others tree representation) but this seemed harder to do in the order you suggested. 

I guess I can extract the features you mentioned in dedicated hooks and then dispatch these hooks in dedicated container to fix the lifecycle issue on the subscription.

PR #6771, opened 2026-08-10
[doc] Add ADR on support for backend filtering in tree based represen…

mcharfadi:
…tations

# Pull request template

## General purpose
- [ ] Bug fixes
- [ ] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?


doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
```diff
@@ -0,0 +1,88 @@
...
+
+- `boolean isTreeItemCandidate(TreeItem treeItem, String searchedValue)`
+
+- `StyledString getHighlightedLabel(TreeItem treeItem, String searchedValue)`
+
+We will need to render the whole tree at first in order to match against the computed labels of the TreeItem.
+
+We can't decide to cut a parent leaf if we did not rendered all its children first in order to be sure they are not matching the filter.
+
+As such we will have another `TreeFilteredRender` utility class that will have this public method `public Tree render(Tree tree)`
```
sbegaudeau: Why the need for such class?

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
```diff
@@ -0,0 +1,88 @@
...
+=== Second Option
+
+We will add a query updateTreeFromSearchedValue that will have for effect to compute a new Tree.
+
+This would behave as with diagram when we send a layout query that has for effect to refresh the payload of the subscription.
+
+
+== Decision Outcome
+
+The second solution is not what is currently done for the `TreeFilters` and as such would mix 2 ways of rendering the Tree representation.
```
sbegaudeau: Respect the pattern for this part of the ADR

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
```diff
@@ -0,0 +1,88 @@
...
+
+=== Advantages
+
+- Less development time.
+- Offer the same features as before.
+
+
+=== Drawbacks
+
+- Rendering the Tree two times might be expensive but we can improve on it later if needed.
```
sbegaudeau: Two? Way more than that. Anytime the tree is refreshed, we will need to re-renderer the entire tree at least once per user to compute the impact of the filter. So if we have 5 users filtering the same tree in a different way, one refresh of the tree will trigger 5 full render of the tree with potentially an eAllContents of hundreds of thousands of model elements (think SysMLv2 libraries in SysON).

It's not "might be expensive but we can improve later", it will be borderline "putting the server on its knees and will be very complex to improve".

I'm not sure that we can do better given the feature expected and the state of the server today.

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
```diff
@@ -0,0 +1,88 @@
...
+The second solution is not what is currently done for the `TreeFilters` and as such would mix 2 ways of rendering the Tree representation.
+
+We will opt for the first solution as it would allow us to implement the backend filter without bringing massive architectural transformation and breaking changes.
+
+
+== Consequences
+
+=== Advantages
+
+- Less development time.
```
sbegaudeau: really?

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
```diff
@@ -0,0 +1,88 @@
...
+- Offer the same features as before.
```
sbegaudeau: Wouldn't the second approach other the same feature too?

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
```diff
@@ -0,0 +1,88 @@
...
+- Compatibility: A default filter should be implemented with the same behavior as what is currently available.
+- Extensibility: The default filter can be changed.
+- Performance: This will have an impact on performance since we'll have to check every element to know if it's candidate or not, the performance overhead should be kept as low as possible
+
+
+== Considered Options
+
+=== First Option
+
+We will add a variable `searchedValue` on the `tree subscriptions`.
```
sbegaudeau: Expand this a bit, the filter will become part of the subscription, anytime a filter is changed we will subscribe to a new versions of the tree.

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
```diff
@@ -0,0 +1,88 @@
...
+
+We will add theses interfaces `IDefaultSearchTreeProvider` `IComposedSearchTreeProvider` `ISearchTreeProviderDelegate`.
```
sbegaudeau: Why? You are listing just technical details, who will use this? Why?

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
```diff
@@ -0,0 +1,88 @@
...
+We will need to render the whole tree at first in order to match against the computed labels of the TreeItem.
+
+We can't decide to cut a parent leaf if we did not rendered all its children first in order to be sure they are not matching the filter.
+
+As such we will have another `TreeFilteredRender` utility class that will have this public method `public Tree render(Tree tree)`
+
+
+=== Second Option
+
+We will add a query updateTreeFromSearchedValue that will have for effect to compute a new Tree.
```
sbegaudeau: How would that really work? The frontend sends a query to compute a new tree and then?

sbegaudeau (changes requested):

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
mcharfadi: I have removed this part as it seemed to specific for an ADR.

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
mcharfadi: I was not sure about this comment, I tried to be more specific about the `Decision Outcome`

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
mcharfadi: I made it more clear that the solution will not perform well with large models

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
mcharfadi: I don't think there is a clear technical advantage between the 2 solutions, but I tried to elaborate more.

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
mcharfadi: I removed advantages that concern both solution.

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
mcharfadi: I've added more details.

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
mcharfadi: I've moved this part to the implementation section and removed some details.

doc/adrs/235_add_backend_filtering_in_tree_based_representation.adoc
mcharfadi: I've added more details so the difference between the 2 solutions is more clear

Issue #6743, opened 2026-07-28
Library's identifier should be derived from its namespace, name, and version


gdaniel:
* [x] **I have checked that this feature has not yet been suggested by someone else.**

The identifier of a Library should be derived from its namespace, name, and version.
Since the database doesn't allow duplicates of (namespace, name, version), it wouldn't be an issue to derive the identifier from these three icons.

It could be done in a similar way we do it for `RepresentationContent` and `RepresentationMetadata`, where the identifier is a combination of the representation ID and the semantic data ID.

The identifiers of existing libraries should be migrated.

gdaniel:
The original need for stable identifiers comes from e2e tests relying on libraries created on startup.
Some other tickets could benefit from stable identifiers, like https://github.com/eclipse-sirius/sirius-web/issues/6453

PR #6859, opened 2026-08-25
[6740] Add support for required libraries in templates

### LINKED ISSUE #6740: Add a way to automatically select libraries when choosing a template

When we create a project, we can now select the template, then select the librairies.
At the moment there is no link between the two steps.
It would be interesting to provide a way to:

- Automatically select some libraries when a given template is selected (because they are usually used with the project)
- Prevent users from un-selecting some libraries when a given template is selected (because the project can't work without them)

gdaniel:
Bug: https://github.com/eclipse-sirius/sirius-web/issues/6740
Fixes #6740

# Pull request template

## General purpose
- [ ] Bug fixes
- [x] New features
- [ ] Documentation
- [ ] Cleanup
- [ ] Tests
- [ ] Build / releng

## Project management

## Dependencies

## Frontend

### General purpose

### Typing

- Hooks are properly typed:

## Backend

### General purpose

### Architecture

### How to test this PR?

Select the papaya - blank template: the java 0.0.3 library should be checked and disabled.
You can select other libraries.
The created project contains all the libraries (disabled or not)


packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
```diff
@@ -55,10 +64,27 @@ public void onSemanticDataCreatedEvent(SemanticDataCreatedEvent event) {
         if (event.causedBy() instanceof ProjectCreatedEvent projectCreatedEvent && projectCreatedEvent.causedBy() instanceof ICreateProjectInput createProjectInput) {
             var projectSemanticData = event.semanticData();
 
-            List<Library> libraries = this.librarySearchService.findAllById(createProjectInput.libraryIds().stream()
+            Set<Library> libraries = new TreeSet<>(Comparator.comparing(Library::getId));
```
gdaniel, line 67: I am not sure we want to override `equals` in Library, so I used a set with a comparator instead. Let me know if I should do it the other way instead.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
```diff
@@ -55,10 +64,27 @@ public void onSemanticDataCreatedEvent(SemanticDataCreatedEvent event) {
...
+            // Always add the libraries required by the template: we cannot trust the input to always contain them.
+            this.projectTemplateProviders.stream()
+                    .map(IProjectTemplateProvider::getProjectTemplates)
+                    .flatMap(Collection::stream)
+                    .filter(projectTemplate -> projectTemplate.id().equals(createProjectInput.templateId()))
+                    .findFirst()
+                    .map(ProjectTemplate::libraries)
+                    .orElse(List.of())
```
gdaniel: The actual project template to use is known in `ProjectCreationApplicationService`, but lost.
We could use an event decorator to keep it around, but I am not sure this is really useful compared to this piece of code. Let me know.

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/projects/ProjectTemplateControllerIntegrationTests.java
```diff
@@ -173,4 +189,80 @@ public void givenPapayaProjectTemplateWhenMutationIsPerformedThenTheProjectIsCre
...
         assertThat(emfEditingContext.getDomain().getResourceSet().getResources()).hasSizeGreaterThan(10);
     }
+
+    @Test
+    @GivenSiriusWebServer
+    @DisplayName("Given a project template with required and optional libraries, when the mutation is performed, then the project is created with a dependency to the libraries")
+    public void givenProjectTemplateWithRequiredAndOptionalLibrariesWhenTheMutationIsPerformedThenTheProjectIsCreatedWithDependencyToTheLibraries(CapturedOutput capturedOutput) throws Exception {
+        TestTransaction.end();
+        this.papayaLibraryInitializer.run();
+        TestTransaction.start();
```
gdaniel: This is required because:

- Libraries are deleted after each test, but the command line runner is run only once when the spring context is created.
- The `end()` before and `start()` after allows to run the initializer and all the cascading events before opening the next transaction, so we are sure everything is around when we start the test.

packages/sirius-web/frontend/sirius-web-application/src/views/new-project/NewProjectView.types.ts
```diff
@@ -11,7 +11,7 @@
  *     Obeo - initial API and implementation
  *******************************************************************************/
 
-import { GQLProjectTemplate } from '../project-browser/create-projects-area/useProjectTemplates.types';
+import { GQLProjectTemplate } from './useAllProjectTemplates.types';
```
gdaniel, line 14: This is the hook actually used by NewProjectView to get the templates.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/project/services/api/ProjectTemplate.java
```diff
@@ -26,11 +26,13 @@ public record ProjectTemplate(
         @NotNull String id,
         @NotNull String label,
         @NotNull String imageURL,
-        @NotNull List<ProjectTemplateNature> natures) {
+        @NotNull List<ProjectTemplateNature> natures,
+        @NotNull List<ProjectTemplateRequiredLibrary> requiredLibraries) {
```
frouene, line 30: I guess this new mandatory field should be documented as a breaking change for downstream application

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/project/services/ProjectTemplateApplicationService.java
```diff
@@ -115,14 +126,25 @@ private Page<ProjectTemplateDTO> handleProjectTemplateModal(Pageable pageable) {
...
+                .map(this::toDTO)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .toList();
+
+        return new ProjectTemplateDTO(projectTemplate.id(), projectTemplate.label(), projectTemplate.imageURL(), libraryDTOs);
+    }
+
+    private Optional<LibraryDTO> toDTO(ProjectTemplateRequiredLibrary library) {
+        return this.librarySearchService.findByNamespaceAndNameAndVersion(library.namespace(), library.name(), library.version())
```
frouene, line 158: If the library does not exist, there is no error and the template is available without it. I don't know if we should block this situation

packages/sirius-web/frontend/sirius-web-application/src/libraryImport/LibrariesImportTable.tsx
```diff
@@ -40,7 +40,25 @@ export const LibrariesImportTable = ({ onSelectedLibrariesChange }: LibrariesImp
   const rows: GQLLibrary[] = state.data?.viewer.libraries.edges.map((edge) => edge.node) || [];
   const count: number = state.data?.viewer.libraries.pageInfo.count ?? 0;
 
-  const [rowSelection, setRowSelection] = useState<MRT_RowSelectionState>({});
+  const [rowSelection, setRowSelection] = useState<MRT_RowSelectionState>(() =>
+    Object.fromEntries(requiredLibraryIds.map((libraryId) => [libraryId, true]))
+  );
+
+  useEffect(() => {
+    setRowSelection((previousSelection) => {
```
frouene: When I switch from a template with a required library to another template the library is still selected. I don't know if it is something we want to avoid

frouene (changes requested):

packages/sirius-web/frontend/sirius-web-application/src/libraryImport/LibrariesImportTable.tsx
gdaniel: This is the existing behavior: if you select template A, check library X, then select template B, library X will be selected.
We can change it, but for the record this is how it works at the moment.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/project/services/api/ProjectTemplate.java
gdaniel: Right, I added a line in the changelog about it

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
sbegaudeau: Honestly I look at the entire method and it feels a bit too smart or complex. Given that on one side you have libraryIds and on the other you have "namespace + name + version", it feels odd and it's entirely under our control. Something doesn't feel right.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
sbegaudeau: This entire sequence of method call is way too long. It should at least be split in two

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
```diff
@@ -55,10 +64,27 @@ public void onSemanticDataCreatedEvent(SemanticDataCreatedEvent event) {
...
+            // Always add the libraries required by the template: we cannot trust the input to always contain them.
+            this.projectTemplateProviders.stream()
+                    .map(IProjectTemplateProvider::getProjectTemplates)
+                    .flatMap(Collection::stream)
+                    .filter(projectTemplate -> projectTemplate.id().equals(createProjectInput.templateId()))
+                    .findFirst()
+                    .map(ProjectTemplate::requiredLibraries)
+                    .orElse(List.of())
+                    .stream()
+                    .map(projectTemplateLibrary -> this.librarySearchService.findByNamespaceAndNameAndVersion(projectTemplateLibrary.namespace(), projectTemplateLibrary.name(),
```
sbegaudeau, line 77: The fact that we have to load them one by one is a red flag (findAllByIds exists for a reason, yet we can't use it).

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/project/services/ProjectTemplateApplicationService.java
sbegaudeau: Yes if we can't resolve the required library, then this should be logged and the template should not be available

sbegaudeau: Funnily enough, we have the same issue here. For each library here we also have to perform a single request for load it instead of batching all this with a `findAllByIds`.

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/projects/ProjectTemplateControllerIntegrationTests.java
sbegaudeau: In that case, the library must be in the sql files just like with other tables with similar issues.

packages/sirius-web/frontend/sirius-web-application/src/libraryImport/LibrariesImportTable.tsx
sbegaudeau: I would rather reset everything cleanly. Since the choice of the library is (and will be) derived from the choice of the template, it would be better not to keep what has been selected.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
sbegaudeau: I don't think that you can really change things properly with regards to this matching of library by id on one side and by namespace + name + version on the other side if we want to let specifiers provide a required library using the namespace, name and version instead of the id.

This just makes it more important to switch the way libraries are identified in the future from an UUID to a string concatenation of namespace, name and version to easily perform a findAllByIds from the namespace, name and version. Right now, it feels odd but I'm not sure that you can really do better. Could you at least indicate it in a comment in this method just to keep that in mind.

sbegaudeau (dismissed):

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
gdaniel: Sure, I'll reference https://github.com/eclipse-sirius/sirius-web/issues/6743 in the method

packages/sirius-web/backend/sirius-web/src/test/java/org/eclipse/sirius/web/application/controllers/projects/ProjectTemplateControllerIntegrationTests.java
gdaniel: @sbegaudeau I tried, but didn't manage to make it work because:

- The `CommandLineRunner` that creates the java libraries runs when the application context starts (and populates the database)
- The SQL script then runs and an unique constraint gets violated (on namespace + name + version)

To add it in the SQL script I should de-activate the command line runner in the tests.
This would mean that the library required by the Papaya Blank template is provided by the `CommandLineRunner` when running the application, and by the SQL script when testing the application,  which feels weird to me.
Let me know if I should go this way.

packages/sirius-web/frontend/sirius-web-application/src/libraryImport/LibrariesImportTable.tsx
gdaniel: Done, the table is now reset with only the required libraries when the template changes.

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/project/services/ProjectTemplateApplicationService.java
gdaniel: Fixed: we now log a message when the library cannot be found, and the `toDTO` method for the template returns an `Optional<ProjectTemplateDTO>`, so we can filter it out of the result when empty.

I've also added a comment with a reference to #6743 in this method to keep track of this limitation

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
gdaniel: Fixed

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
gdaniel: Done

packages/sirius-web/backend/sirius-web-application/src/main/java/org/eclipse/sirius/web/application/library/services/ProjectLibrariesImporter.java
```diff
@@ -54,11 +63,29 @@ public ProjectLibrariesImporter(ISemanticDataUpdateService semanticDataUpdateSer
...
+            // Retrieving the libraries looks more complex than it needs to be because we work with (namespace, name, version) as well as UUID identifiers. This should be unified once #6743 is fixed.
+            Set<Library> libraries = new TreeSet<>(Comparator.comparing(Library::getId));
+            // Always add the libraries required by the template: we cannot trust the input to always contain them.
+            Optional<ProjectTemplate> optionalProjectTemplate = this.projectTemplateProviders.stream()
+                    .map(IProjectTemplateProvider::getProjectTemplates)
+                    .flatMap(Collection::stream)
+                    .filter(projectTemplate -> projectTemplate.id().equals(createProjectInput.templateId()))
+                    .findFirst();
+
+            if (optionalProjectTemplate.isPresent()) {
```
frouene: maybe use `ifPresent` is clearer here

packages/sirius-web/frontend/sirius-web-application/src/libraryImport/LibrariesImportTable.tsx
```diff
@@ -40,7 +40,14 @@ export const LibrariesImportTable = ({ onSelectedLibrariesChange }: LibrariesImp
...
 
-  const [rowSelection, setRowSelection] = useState<MRT_RowSelectionState>({});
+  const [rowSelection, setRowSelection] = useState<MRT_RowSelectionState>(() =>
+    Object.fromEntries(requiredLibraryIds.map((libraryId) => [libraryId, true]))
+  );
+
+  useEffect(() => {
+    // Reset the selected rows and select the new required libraries when the required libraries change.
+    setRowSelection(Object.fromEntries(requiredLibraryIds.map((id) => [id, true])));
+  }, [requiredLibraryIds]);
```
frouene: The array is recreated at each time even if the values do not change, this makes the selection reset to the required one all the time.
We have to use its content to trigger this `useEffect`

CHANGELOG.adoc
```diff
@@ -45,6 +45,8 @@ A new GraphQL schema `palette.graphqls` has been added.
 - https://github.com/eclipse-sirius/sirius-web/issues/6815[#6815] [tree] In case a tree item palette provider is available for a representation, the `ITreeItemContextMenuEntryProvider` contributions will be ignored.
 A tree item palette provider is now provided for the explorer which will thus now ignore `ITreeItemContextMenuEntryProvider`.
 - https://github.com/eclipse-sirius/sirius-web/issues/6835[#6835] [tree] `withImpactAnalysis` attribute has been removed from `GQLFetchTreeItemContextMenuEntry` type, it's available on `GQLSingleClickTreeItemContextMenuEntry`.
+- https://github.com/eclipse-sirius/sirius-web/issues/6740[#6740] [sirius-web] Add support for required libraries in templates.
```
frouene: I'll move it to the cooldown branch and move these comments in the 2026.11.0 section

packages/sirius-web/backend/sirius-web-tests-data/src/main/resources/sirius-web-scripts/papaya-library.sql
gdaniel, line 1: @frouene I added the actual java@0.0.3 library, but it is not really used in the test, we just need a library, it could be empty.
I thought it was a good idea to have the same library has the one created by the application with the command line runner, but if you think it is too much we can just have an empty document in the library.

gdaniel: Note that this is the reason the build fails: the PR is now too large, because it contains ~600 lines of SQL

frouene (approved):

